PR #4, opened 2017-10-11
Backend changes for whites space character bug in target space.

KristianShishoev:


com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/cf/v1_0/CloudModelBuilderTest.java
```diff
@@ -332,8 +332,7 @@ public void setUp() throws Exception {
         DescriptorHandler handler = getDescriptorHandler();
 
         String targetName = extensionDescriptor.getDeployTargets().get(0);
-        Target implicitTarget = getDeployTargetFactory().create(targetName, platforms.get(0).getName());
-        Target target = handler.findTarget(targets, targetName, implicitTarget);
+        Target target = handler.findTarget(targets, targetName, null);
```
monochronique: Why is this changed to null? The DetectTargetStep still computes an implicit target during its execution.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/validators/MtaArchiveValidatorTest.java
```diff
@@ -35,7 +36,7 @@
 
     private static final long MAX_MTA_DESCRIPTOR_SIZE = 1024 * 1024l; // 1 MB
 
-    private static final String XS2_PLATFORM_NAME = "initial initial";
+    private static final String XS2_PLATFORM_NAME = "XS2-INITIAL";
```
monochronique: Why is this changed?

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/PublishConfigurationEntriesStepTest.java
```diff
@@ -81,7 +82,8 @@ public void setUp() throws Exception {
     public void prepareDao() throws Exception {
         for (ConfigurationEntry entry : exisitingConfigurationEntries) {
             Mockito.when(configurationEntryDaoMock.find(Mockito.matches(entry.getProviderNid()), Mockito.matches(entry.getProviderId()),
-                Mockito.matches(entry.getProviderVersion().toString()), Mockito.matches(entry.getTargetSpace()), Mockito.any(),
+                Mockito.matches(entry.getProviderVersion().toString()),
+                Mockito.any(CloudTarget.class), Mockito.any(),
```
monochronique: In this way the test no longer validates correct filtration based on the target. Can this be enhanced to take into consideration the ConfigurationEntry's target?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/resources/ConfigurationEntriesResource.java
```diff
@@ -165,9 +165,14 @@ public Response getConfigurationEntry(@PathParam(ID) long id) throws SLException
     @POST
     @Consumes(MediaType.APPLICATION_XML)
     public Response createConfigurationEntry(String xml) throws SLException {
+        
         ConfigurationEntryDto dto = parseDto(xml, CREATE_CONFIGURATION_ENTRY_SCHEMA_LOCATION);
-
         ConfigurationEntry result = entryDao.add(dto.toConfigurationEntry());
+
```
monochronique: The exception here is thrown after inserting the result in the DAO. Could this potentially lead to a corrupted database content?

com.sap.cloud.lm.sl.cf.web/src/test/java/com/sap/cloud/lm/sl/cf/web/resources/ConfigurationEntriesResourceTest.java
```diff
@@ -53,52 +54,65 @@ public ConfigurationEntriesResourceTest(TestCase<TestInput> test) {
...
                 new SearchRequestTest(new SearchRequestTestInput(Arrays.asList("{\"foo\":\"bar\",\"baz\":\"qux\"}"), "R:parsed-properties-01.json"), "R:configuration-entries-resource-test-output-08.json"),
             },
-            // (12)
+            // (13)
             {
                 new SearchRequestTest(new SearchRequestTestInput(Arrays.asList("a"), "R:parsed-properties-01.json"), "E:Could not parse content query parameter as JSON or list"),
             },
+            // (14)
+            {
+                new PostRequestTest(new PostRequestTestInput("configuration-entry-06.xml"), "R:configuration-entries-resource-test-output-01.json"),
```
monochronique: Should there be a test case where the user specifies both "target-space" and "provider-target"? What would happen then? This seems to be a valid input scenario according to the xsd schema.

KristianShishoev: It must take take parameters from provider-target section but i added test just in case.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/validators/MtaArchiveValidatorTest.java
KristianShishoev: Code must cover both cases but just in case i added tests for both scenarios.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/cf/v1_0/CloudModelBuilderTest.java
KristianShishoev: Because all targets are described in targets.json and in tests all targets are referenced from this file.
So this implicit target will never work because this create method requires org and space parameters.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/liquibase/TransformFilterColumn.java
```diff
@@ -0,0 +1,183 @@
...
+
+    Map<Long, String> transformData(Map<Long, String> retrievedData) {
+
+        Map<Long, String> transformedData = new HashMap<Long, String>();
+        for (Map.Entry<Long, String> entry : retrievedData.entrySet()) {
+
+            if (StringUtils.isEmpty(entry.getValue())) {
+                continue;
+            }
+            JsonElement initialFilterJsonElement = new JsonParser().parse(entry.getValue());
```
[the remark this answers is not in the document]
KristianShishoev: No need. The code is clear in this way also.

PR #7, opened 2017-10-17
Allow two or more resources to point to the same existing service

DenitsaKostova:
Change-Id: Ieec59c05497312697e94b00f30615a9876d8d0fb


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v1_0/ApplicationsCloudModelBuilder.java
```diff
@@ -134,7 +134,7 @@ protected CloudApplicationExtended getApplication(Module module) throws SLExcept
         List<String> resolvedIdleUris = xsPlaceholderResolver.resolve(idleUris);
         List<String> customUris = new UrisClassifier(xsPlaceholderResolver).getCustomUris(deployedModule);
         List<String> fullResolvedUris = ListUtil.merge(resolvedUris, customUris);
-        List<String> services = getApplicationServices(module, true);
+        List<String> services = new ArrayList<String>(new TreeSet<String>(getApplicationServices(module, true)));
```
nvvalchev: It would be better to do the duplication removal directly in the getApplicationServices method.

PR #15, opened 2017-10-25
Aggregate iteration variables in Map.

ttsokov:


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/AbstractXS2SubProcessStarterStep.java
```diff
@@ -18,9 +20,13 @@
     protected ExecutionStatus pollStatusInternal(DelegateExecution context) throws Exception {
         try {
             String userId = StepsUtil.determineCurrentUser(context, getStepLogger());
-
             int nextIndex = (int) context.getVariable(getIndexVariable());
-            context.setVariable(getIterationVariableName(), getIterationVariable(context, nextIndex));
+            Map<String, Object> iterationVariablesMap = getIterationVariablesMap(context, nextIndex);
+
+            iterationVariablesMap.keySet().forEach(iterationVariableName -> {
```
enchobelezirev: Using this map structure here is a little bit too much. Could you introduce a big variable in which to store the required variables? Something like master variable?

ttsokov: In this way this "master" variable will introduce inconsistency with the current iteration logic. 
Another way is to refactor the iteration logic to support Map variables, which are necessary for some sub processes.

ttsokov:
Unnecessary PR.

PR #28, opened 2017-11-10
Add healthCheckHttpEndpoint to UpdateAppStep

DenitsaKostova:
Change-Id: Icb118f0b4831cd920df6d278104c39d5078ad507

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UpdateAppStep.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UpdateAppStep.java
@@ -219,12 +219,13 @@ public class UpdateAppStep extends CreateAppStep {
         String stack = staging.getStack();
         Integer healthCheckTimeout = staging.getHealthCheckTimeout();
         String healthCheckType = staging.getHealthCheckType();
+        String healthCheckHttpEndpoint = staging.getHealthCheckHttpEndpoint();
         Boolean sshEnabled = staging.isSshEnabled();
         return (buildpackUrl != null && !buildpackUrl.equals(existingStaging.getBuildpackUrl()))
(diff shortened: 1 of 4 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UpdateAppStep.java
```diff
@@ -219,12 +219,13 @@ private boolean hasChanged(StagingExtended staging, Staging existingStaging) {
         String stack = staging.getStack();
         Integer healthCheckTimeout = staging.getHealthCheckTimeout();
         String healthCheckType = staging.getHealthCheckType();
+        String healthCheckHttpEndpoint = staging.getHealthCheckHttpEndpoint();
         Boolean sshEnabled = staging.isSshEnabled();
         return (buildpackUrl != null && !buildpackUrl.equals(existingStaging.getBuildpackUrl()))
             || (command != null && !command.equals(existingStaging.getCommand()))
             || (stack != null && !stack.equals(existingStaging.getStack()))
             || (healthCheckTimeout != null && !healthCheckTimeout.equals(existingStaging.getHealthCheckTimeout()))
-            || (healthCheckType != null) || (sshEnabled != null);
```
enchobelezirev: shouldn't these three null cheks be transformed to !<new-staging-parameter>.equals(<existing-staging-parameter>)? :)

DenitsaKostova: Currently not, because now they are part of StagingExtended but soon a fix for this will be provided where they will be part of Staging and this check will be implemented :)

enchobelezirev: aha. I thought that this is the refactoring

PR #38, opened 2017-11-15
Needed for the adoption in the internal project

mptsonev:


com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/ParameterMetadata.java
```diff
@@ -58,7 +58,7 @@ public ParameterMetadata build() {
             result.setId(id);
             result.setRequired(CommonUtil.getOrDefault(required, false));
             result.setDefaultValue(defaultValue);
-            result.setType(type);
+            result.setType(CommonUtil.getOrDefault(type, ParameterType.STRING));
```
mptsonev: This is done in order to preserve the previous behaviour of the ParameterMetadata builder, where if no type is specified then the type is defaulted to String.

PR #53, opened 2017-11-29
Add zip bomb validation

KristianShishoev:
Change-Id: I6d2cc1683c83aaab58d25c04891643f43be8cfd5

Change-Id: Ic3de91486384f365ed755da5a554f9d9065bc755
JIRA:LMCROSSITXSADEPLOY-405

### DIFF
--- a/com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
+++ b/com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
@@ -1,0 +1,91 @@
+package com.sap.cloud.lm.sl.cf.client.util;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.text.MessageFormat;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.sap.cloud.lm.sl.cf.client.message.Messages;
+
+public class StreamManager {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(StreamManager.class);
+    private InputStream inputStream;
+    public static final String ARCHIVE_ENTRY_SEPARATOR = "/";
+
+    public StreamManager(InputStream inputStream) {
+        this.inputStream = inputStream;
+    }
+
+    public File saveStreamToFile(String entryName) throws IOException {
+        
+        File file = createTempFile(entryName);
+        LOGGER.debug(MessageFormat.format(Messages.SAVING_INPUT_STREAM_TEMP_FILE, file.getPath()));
+        try (FileOutputStream outputStream = new FileOutputStream(file)) {
+            IOUtils.copy(inputStream, outputStream);
+        }
+        LOGGER.debug(MessageFormat.format(Messages.INPUT_STREAM_SAVED_IN_TEMP_FILE, file.getPath()));
+        return file;
+    }
+
+    private File createTempFile(String entryName) throws IOException {
+
+        int dotIndex = entryName.lastIndexOf('.');
+        String fileName = (dotIndex > 0) ? entryName.substring(0, dotIndex) : entryName;
+        String suffix = (dotIndex > 0) ? entryName.substring(dotIndex) : null;
+        if (fileName.length() < 3) {
+            fileName = fileName + "000";
+        }
+        return File.createTempFile(fileName, suffix);
+    }
+
+    public void saveStreamToDirectory(String entryName, String directoryName, Path tempDirectory) throws IOException {
+        StreamUtil.validateEntry(entryName);
+        Path filePath = StreamUtil.resolveTempEntryPath(entryName, directoryName, tempDirectory);
+        StreamUtil.createParentDirectories(filePath);
+        createFile(filePath);
+    }
+
+    public File saveZipStreamToDirectory(String entryName, long maxZipEntrySize) throws IOException {
+        Path path = StreamUtil.getTempDirectory(entryName);
+        File dir = path.toFile();
+        LOGGER.debug(MessageFormat.format(Messages.SAVING_INPUT_STREAM_TO_TMP_DIR, dir.getPath()));
+        saveEntriesToTempDirectory(path, entryName, maxZipEntrySize);
+        LOGGER.debug(MessageFormat.format(Messages.INPUT_STREAM_SAVED_TO_TMP_DIR, dir.getPath()));
+        return dir;
+    }
+
+    private void saveEntriesToTempDirectory(Path tempDirectory, String rootEntryName, long maxZipEntrySize)
+        throws IOException, FileNotFoundException {
+        ZipInputStream zipInputStream = (ZipInputStream) inputStream;
+        for (ZipEntry zipEntry; (zipEntry = zipInputStream.getNextEntry()) != null;) {
+            String currentEntryName = zipEntry.getName();
+            if (!zipEntry.getName().startsWith(rootEntryName)) {
+                continue;
+            }
+            StreamUtil.validateZipEntrySize(zipEntry, maxZipEntrySize);
+            StreamUtil.validateEntry(currentEntryName);
+            Path filePath = StreamUtil.resolveTempEntryPath(currentEntryName, rootEntryName, tempDirectory);
+            if (zipEntry.getName().endsWith(ARCHIVE_ENTRY_SEPARATOR)) {
+                Files.createDirectories(filePath);
+            } else {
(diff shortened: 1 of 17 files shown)


com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(StreamManager.class);
+    private InputStream inputStream;
+    public static final String ARCHIVE_ENTRY_SEPARATOR = "/";
+
+    public StreamManager(InputStream inputStream) {
+        this.inputStream = inputStream;
+    }
+
+    public File saveStreamToFile(String entryName) throws IOException {
+        int i = entryName.lastIndexOf('.');
```
enchobelezirev: What is 'i'?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+        String fileName = (i > 0) ? entryName.substring(0, i) : entryName;
+        String suffix = (i > 0) ? entryName.substring(i) : null;
+        if (fileName.length() < 3) {
+            fileName = fileName + "000";
```
enchobelezirev: "000" what magic is this?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+        }
```
enchobelezirev: From line 31 to 35, will it be good to extract it in separate method ?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+        File file = File.createTempFile(fileName, suffix);
+        LOGGER.debug(MessageFormat.format("Saving input stream to temporary file \"{0}\"...", file.getPath()));
```
enchobelezirev: extract "Saving input stream to temporary file \"{0}\"..." to constant

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+        try (FileOutputStream outputStream = new FileOutputStream(file)) {
+            IOUtils.copy(inputStream, outputStream);
+        }
+        LOGGER.debug(MessageFormat.format("Input stream saved to temporary file \"{0}\"", file.getPath()));
```
enchobelezirev: The same here .. "Input stream saved to temporary file \"{0}\""

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+    public File saveZipStreamToDirectory(String entryName, long maxZipEntrySize) throws IOException {
+        Path path = StreamUtil.getTempDirectory(entryName);
+        File dir = path.toFile();
+        LOGGER.debug(MessageFormat.format("Saving input stream to temporary directory \"{0}\"...", dir.getPath()));
+        saveEntries(path, entryName, maxZipEntrySize);
+        LOGGER.debug(MessageFormat.format("Input stream saved to temporary directory \"{0}\"", dir.getPath()));
+        return dir;
+    }
+
+    private void saveEntries(Path tempDirectory, String rootEntryName, long maxZipEntrySize) throws IOException, FileNotFoundException {
```
enchobelezirev: The method name saveEntries is not really applicable for the behavior of the method. Should we rename it to saveEntriesToTempLocation?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+        ZipInputStream zis = (ZipInputStream) inputStream;
```
enchobelezirev: zis is too algorithmic. Could we rename it it archiveInputStream or smthing similar?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+        for (ZipEntry e; (e = zis.getNextEntry()) != null;) {
+            String currentEntryName = e.getName();
+            if (!e.getName().startsWith(rootEntryName)) {
+                continue;
+            }
+            StreamUtil.validateZipEntrySize(e, maxZipEntrySize);
+            StreamUtil.validateEntry(currentEntryName);
```
enchobelezirev: I suppose that the validateEntry and the validateZipEntrySize both throw an exception. What will happen with the inputstream then? Will it be good to introduce a finally block?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+            if (e.getName().endsWith(ARCHIVE_ENTRY_SEPARATOR)) {
+                Files.createDirectories(filePath);
+            } else {
+                createFile(filePath);
+            }
+        }
+    }
+
+    private void createFile(Path fileToCreate) throws IOException {
+        Files.createFile(fileToCreate);
```
enchobelezirev: This returns a Path object, couldn't it be used?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+        try (OutputStream outputStream = Files.newOutputStream(fileToCreate)) {
+            IOUtils.copy(inputStream, outputStream);
```
enchobelezirev: woudn't be easier to use Files::copy(Path source, OutputStream out) ?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    public InputStream getInputStream() {
+        return inputStream;
+    }
+
+    public void setInputStream(InputStream inputStream) {
```
enchobelezirev: Why do we have a setter when the inputStream is passed from the constructor?

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamUtil.java
```diff
@@ -20,23 +20,23 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.sap.cloud.lm.sl.cf.client.message.Messages;
+import com.sap.cloud.lm.sl.common.ContentException;
+
 public class StreamUtil {
```
enchobelezirev: How this is different from the StreamManager?

KristianShishoev: This class has only static methods. StreamManager has constructor with parameter InputStream. This is the reason to split this StreamUtil class.

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
KristianShishoev: This streams must be handled by the method that created them.

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/StreamManager.java
KristianShishoev: No significat difference because Files::copy also requires to close stream manually.

PR #73, opened 2018-01-03
Cleanup pom

ddonchev:
removing a dead config from parent pom.xml


pom.xml
```diff
@@ -58,7 +58,6 @@
         <jackson.version>2.5.0</jackson.version>
         <semver4j.version>2.1.0</semver4j.version>
         <jgit.version>4.0.1.201506240215-r</jgit.version>
-        <java.container.security.version>0.19.1</java.container.security.version>
```
enchobelezirev: This should be moved to the internal project because in the Xs2CustomTokenServices - the dependency is needed in order to validate the user's tokens :)

PR #84, opened 2018-01-17
Protect against potential DoS attack

mptsonev:
An attacker might try to execute a request too large for the backend to
process. This commit adds a limit of 1MB on a PUT/POST request size.


mptsonev:
Works with curl too.

PR #90, opened 2018-01-26
Add support for executed application state

enchobelezirev:
Change-Id: Iee19eadc4451ccd779a185a36304df029d42f38c


ttsokov:
In future we should test the task execution of HDI Deployer apps.

PR #97, opened 2018-02-02
Bugfix Optional resources

theghost5800:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/CloudModelBuilderUtil.java
```diff
@@ -76,12 +76,18 @@ public static String buildImplicitDeployTargetName(String org, String space) {
...
-        if (resource.getType() == null) {
+        if (!compareWithServiceTypes(resource, propertiesAccessor)) {
             return false;
         }
         return !ResourceType.EXISTING_SERVICE_KEY.toString().equals(getResourceType(resource, propertiesAccessor));
     }
 
+    private static boolean compareWithServiceTypes(Resource resource, PropertiesAccessor propertiesAccessor) {
+        String resourceType = getResourceType(resource, propertiesAccessor);
+        return ResourceType.MANAGED_SERVICE.toString().equals(resourceType) || ResourceType.EXISTING_SERVICE.toString().equals(resourceType)
```
mptsonev: I don't like this. It would be better if there is a list of service types returned from the ResourceType enum and the check to be if the computed resourceType is contained in that list. Whatsmore, the getResourceType method seems like should return a ResourceType rather than a String. Could this be revised?

PR #114, opened 2018-02-28
Refactor deletetion of file chunks in clean up job

KristianShishoev:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/FileUtils.java
```diff
@@ -24,6 +27,19 @@ public static void copyFile(Path fromPath, Path toPath) throws IOException {
         }
         Files.copy(fromPath, toPath, StandardCopyOption.REPLACE_EXISTING, StandardCopyOption.COPY_ATTRIBUTES);
     }
+    
+    public static List<String> splitFilesInChunks(List<String> fileIds) {
```
enchobelezirev: Why this method is in FileUtils? The class here is supposed to work only with files. This method is working with strings which are named fileIds :) and it is splitting them if they contain "," otherwise it is taking them whole.
Maybe extracting it in some other util class is a good idea, what do you think? :)

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/jobs/CleanUpJobTest.java
```diff
@@ -0,0 +1,27 @@
...
+    public void testSplitAllFilesInChunks() {
+        
+        Map<String, List<String>> spaceToFileIds = new HashMap<>();
+        List<String> fileIds = new ArrayList<String>();
+        fileIds.add(null);
+        fileIds.add("9f87be64-6519-4576-b426-42548840f2ec");
+        fileIds.add("9f87be64-6519-4516-b426-42543845f2az,9f87ne64-6519-1234-b426-42548840f2gh,9f87be64-1239-4567-b426-34548840f2oq");
+        spaceToFileIds.put("space", fileIds);
+        Map<String, List<String>> splitAllFilesInChunks = cleanUpJob.splitAllFilesInChunks(spaceToFileIds);
+        assertEquals(5, splitAllFilesInChunks.get("space").size());
```
enchobelezirev: Here maybe an assertion, whether the files are the one that should be, will be good?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/jobs/CleanUpJob.java
```diff
@@ -248,18 +249,27 @@ private void removeProcessLogs(List<String> oldFinishedOperationsIds) {
...
+            int removedOldFiles = fileService.deleteAllByFileIds(spaceToFileChunkIds);
             LOGGER.info("Deleted MTA files: " + removedOldFiles);
         } catch (FileStorageException e) {
             throw new SLException(e);
         }
     }
 
+    public Map<String, List<String>> splitAllFilesInChunks(Map<String, List<String>> spaceToFileIds) {
+        Map<String, List<String>> spaceToFileChunks = new HashMap<String, List<String>>();
+        for (String space : spaceToFileIds.keySet()) {
```
enchobelezirev: Is there a way to achieve this using a Stream::map function? It will be more clear. If not - leave it as it is :)

KristianShishoev: In my opinion it will become very difficult to understand and also difficult to implement.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/jobs/CleanUpJobTest.java
KristianShishoev: Done.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/FileUtils.java
KristianShishoev: Good remark. Moved to NameUtil.

PR #115, opened 2018-03-01
Split 'task_id' column in 'progress_message' table

nictas:


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ProcessStepHelper.java
```diff
@@ -30,18 +30,18 @@
...
     private ProgressMessageService progressMessageService;
-    protected ProcessLoggerProviderFactory processLoggerProviderFactory;
-    private StepIndexProvider stepIndexProvider;
+    private ProcessLoggerProviderFactory processLoggerProviderFactory;
+    private TaskIndexProvider taskIndexProvider;
 
-    private String indexedStepName;
-    private int stepIndex;
+    String taskId;
+    String taskIndex;
```
enchobelezirev: Why not private, used in tests?

nictas: Yup.

PR #124, opened 2018-03-10
Fix issues in FilesApi

mptsonev:
The WebApplicationException does not take in consideration the cause in the constructor. This makes the message of the exception disappear from the response.


enchobelezirev:
Is this mean that when an application is thrown, only the message will be shown? Will we loose the stacktrace and more importantly, do we need it? :)
Also, as we have started some improvements, maybe it will be good idea to analyze the exception and based on it, to throw the right WebApplicationException(BadRequestException, NotAuthorizedException,ForbiddenException,NotFoundException, etc.)? What would you say?

mptsonev:
We don't really need the stacktrace, since this is thrown in the REST resource to the client. May be it's a good idea to log the stacktrace in order to preserve it. 

Throwing a more meaningful WebApplicationException, however, is a challenge, since the exception would most likely be an IOException. Therefore, I think the message of the exception is enough in order to understand the actual issue.

PR #134, opened 2018-03-16
Parse JSONs to Java objects properly

boyan-velinov:
Jira; LMCROSSITXSADEPLOY-943

### DIFF
--- a/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/dto/serialization/v1/DeployTargetDto.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/dto/serialization/v1/DeployTargetDto.java
@@ -16,7 +16,7 @@ import com.google.gson.annotations.Expose;
 import com.google.gson.annotations.JsonAdapter;
 import com.google.gson.annotations.SerializedName;
 import com.sap.cloud.lm.sl.cf.core.dto.persistence.PersistentObject;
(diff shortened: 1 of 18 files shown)


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/dto/serialization/v1/DeployTargetDto.java
```diff
@@ -48,7 +48,7 @@
     private String description;
 
     @Expose
-    @JsonAdapter(PropertiesAdapterFactory.class)
```
nictas: These should no longer be necessary, right?

com.sap.cloud.lm.sl.cf.core/src/test/resources/com/sap/cloud/lm/sl/cf/core/cf/detect/deployed-components-10.json
```diff
@@ -12,9 +12,9 @@
           "services": [],
           "providedDependencyNames": [],
           "deployAttributes": {
-            "success-marker": "SUCCESS",
```
nictas: Why are these reordered? Is this something to do with the type of the map? Is it HashMap, because if it is this ordering might not work on my machine. :)

boyan-velinov: Adapter uses TreeMap so it re-order entries in the Map. So it should not be different on any machine

PR #160, opened 2018-04-30
Stop applications if their execution fails

nictas:


ddonchev:
Deleting comments, yeah!

PR #182, opened 2018-05-25
LMCROSSITXSADEPLOY-1071

nictas:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/SupportedParameters.java
```diff
@@ -155,10 +156,10 @@
         .unmodifiableSet(new HashSet<>(Arrays.asList(SERVICE_NAME, SERVICE, SERVICE_PLAN, SERVICE_ALTERNATIVES, SERVICE_PROVIDER,
             SERVICE_VERSION, SERVICE_CONFIG, SERVICE_CONFIG_PATH, SERVICE_TAGS, SERVICE_KEY_NAME)));
 
-    public static final Set<String> APP_ATTRIBUTES = Collections.unmodifiableSet(
-        new HashSet<>(Arrays.asList(EXECUTE_APP, SUCCESS_MARKER, FAILURE_MARKER, STOP_APP, CHECK_DEPLOY_ID, REGISTER_SERVICE_URL,
-            REGISTER_SERVICE_URL_SERVICE_NAME, REGISTER_SERVICE_URL_SERVICE_URL, CREATE_SERVICE_BROKER, SERVICE_BROKER_NAME,
-            SERVICE_BROKER_USERNAME, SERVICE_BROKER_PASSWORD, SERVICE_BROKER_URL, SERVICE_BROKER_SPACE_SCOPED, DEPENDENCY_TYPE, NO_START)));
+    public static final Set<String> APP_ATTRIBUTES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(EXECUTE_APP, SUCCESS_MARKER,
```
enchobelezirev: formating ?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
```diff
@@ -285,7 +282,16 @@ protected String getIndexVariable() {
...
-        return configuration.getUploadAppTimeout();
+        CloudApplication app = StepsUtil.getApp(context);
+        int uploadTimeout = extractUploadTimeoutFromAppAttributes(app, DEFAULT_APP_UPLOAD_TIMEOUT);
+        getStepLogger().debug(Messages.UPLOAD_APP_TIMEOUT, uploadTimeout);
+        return uploadTimeout;
+    }
+
+    private int extractUploadTimeoutFromAppAttributes(CloudApplication app, int defaultAppUploadTimeout) {
+        ApplicationAttributes appAttributes = ApplicationAttributes.fromApplication(app);
+        Number uploadTimeout = appAttributes.get(SupportedParameters.UPLOAD_TIMEOUT, Number.class, defaultAppUploadTimeout);
```
enchobelezirev: I am wondering whether this is consistent with the CF push approach? There is no upload-timeout but instead there is just timeout or start timeout which are applied either for just one operation or the whole push operation?

From the user POV, it will be easier to configure the timeout for each application, but on the other hand, is it necessary?

Why is bad to leave it to be configured from the ENV of deploy-service?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollUploadAppStatusExecution.java
```diff
@@ -15,32 +15,17 @@
...
-        // Get the next cloud application from the context
-        final CloudApplication app = StepsUtil.getApp(execution.getContext());
+    public AsyncExecutionState execute(ExecutionWrapper execution) {
+        CloudApplication app = StepsUtil.getApp(execution.getContext());
 
         try {
             execution.getStepLogger()
                 .debug(Messages.CHECKING_UPLOAD_APP_STATUS, app.getName());
-            String status = (String) execution.getContext().getVariable(Constants.VAR_UPLOAD_STATE);
-            if (AsyncExecutionState.ERROR.name()
```
enchobelezirev: I suppose this was the case "This should not happen" :D

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollUploadAppStatusExecution.java
```diff
@@ -15,32 +15,17 @@
...
-            if (uploadToken == null) {
-                String message = format(Messages.APP_UPLOAD_TIMED_OUT, app.getName());
-                execution.getStepLogger()
-                    .error(message);
-                return AsyncExecutionState.ERROR;
-            }
 
             CloudFoundryOperations client = execution.getCloudFoundryClient();
 
+            String uploadToken = (String) execution.getContext()
```
enchobelezirev: Could it be null here?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
```diff
@@ -53,63 +53,26 @@ public StepPhase executeAsyncStep(ExecutionWrapper execution) throws FileStorage
...
-            execution.getContext()
-                .setVariable(Constants.VAR_UPLOAD_STATE, status.name());
+        } catch (CloudFoundryException cfe) {
+            CloudControllerException e = new CloudControllerException(cfe);
+            getStepLogger().error(e, Messages.ERROR_UPLOADING_APP, app.getName());
+            logException(execution.getContext(), e);
+            throw e;
+        } catch (SLException e) {
+            getStepLogger().error(e, Messages.ERROR_UPLOADING_APP, app.getName());
+            throw e;
```
enchobelezirev: Why not logging the SLException using the logException method?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
```diff
@@ -53,63 +53,26 @@ public StepPhase executeAsyncStep(ExecutionWrapper execution) throws FileStorage
...
             String uploadToken = asyncUploadFiles(execution, client, app, appArchiveId, fileName);
+            getStepLogger().debug(Messages.STARTED_ASYNC_UPLOAD_OF_APP_0, app.getName());
             execution.getContext()
                 .setVariable(Constants.VAR_UPLOAD_TOKEN, uploadToken);
-            getStepLogger().debug("Started async upload of application \"{0}\"", fileName, app.getName());
-            status = AsyncExecutionState.RUNNING;
-        } finally {
-            execution.getContext()
-                .setVariable(Constants.VAR_UPLOAD_STATE, status.name());
+        } catch (CloudFoundryException cfe) {
```
enchobelezirev: Another question, before we used to catch Throwable and FileStorageException, for the FSE I suppose it is fine to catch just a SLException, but what about the Throwable, could there be such case?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/SupportedParameters.java
nictas: Yup, + the new UPLOAD_TIMEOUT attribute.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollUploadAppStatusExecution.java
nictas: It could've happened if the previous step was using synchronous upload and it failed, but since synchronous upload is no longer used, yes this should not happen. :D

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
nictas: logException() logs the exception in the process logs and as a progress message. getStepLogger().error() does the same. I think we were using logException(), because this logic was previously in the runnable.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
nictas: Any exception that is not handled in this class is handled in SyncActivitiStep. There's a catch Throwable there. :P

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
enchobelezirev: Aha, and why on 69 row, there is logException method usage? :D

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollUploadAppStatusExecution.java
nictas: I was wondering that as well. The only way it could be null, is if the controller client returned null from its asyncUploadApplication() method. I don't think we should handle that scenario. If it returned null, then it may be able to also handle it in its getUploadStatus() method.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
nictas: Nice catch! I must have missed it. :)

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
nictas: It is not necessarily bad, it's just that this is a better and more flexible approach. It allows users to specify smaller timeouts for applications they know are small in size, while also allowing them to put bigger timeouts on large applications (SAPUI5, for example).

cf push does not have an upload timeout at all, but since we do, it's better if the user can configure it without involving us.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
enchobelezirev: verno be :d

PR #184, opened 2018-06-01
Adopt new methods in cf-java-client

KristianShishoev:


com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/CloudFoundryClientExtended.java
```diff
@@ -63,7 +63,19 @@ public void withTimeoutExecutor(TimeoutExecutor timeoutExecutor) {
             throw fromException(cause.getMessage(), cause, HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
-
+    
+    @Override
+    public List<String> getSpaceManagers2(UUID spaceGuid) {
+        try {
+            return executeWithTimeout(() -> executeWithRetry(() -> super.getSpaceManagers(spaceGuid).stream()
```
nictas: Not sure if we should retry this request in case of 404 errors... If we do and the user doesn't "see" the space, because he doesn't have permissions, then he'll have to wait 15 seconds to see a 403 response from us. @nvvalchev What do you think?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/MockCloudFoundryClient.java
```diff
@@ -751,4 +751,19 @@ public void associateManagerWithSpace(String orgName, String spaceName, String s
         return null;
     }
 
+    @Override
```
nvvalchev: Why not delete the MockCloudFoundryClient.java? It is not used anymore as I see.

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/CloudFoundryClientExtended.java
nvvalchev: I see we apply the same strategy for all other methods in CloudFoundryClientExtended. Here things are a little bit different. The entry point for authorization being getSpaceManagers2/Developers2/Auditors2 this means we need to be careful here. This is a entry point for polling, which in case of sporadic 404 from controller would break the client side.

The following should also be considered:
- is there any chance the controller to return 404?
- could the retry in this case be pushed to CLI side?

Short answer is that the handling is as before and not changing it would keep the old behavior.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/MockCloudFoundryClient.java
KristianShishoev: Ok i will have that in mind and try to remove in new commit.

PR #185, opened 2018-06-01
LMCROSSITXSADEPLOY-1033

nictas:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/message/Messages.java
```diff
@@ -147,7 +147,9 @@
     public static final String GATHER_STATISTICS = "Gather statistics: {0}";
     public static final String HEALTH_CHECK_CONFIGURATION = "Health check configuration: {0}";
     public static final String MAIL_API_URL = "Mail API URL: {0}";
-    public static final String CONTROLLER_OPERATIONS_TIMEOUT = "CFG_CONTROLLER_OPERATIONS_TIMEOUT: {0}";
+    public static final String CONTROLLER_OPERATIONS_TIMEOUT = "Controller operations timeout: {0}";
+    public static final String APPLICATION_ID = "Application ID: {0}";
```
enchobelezirev: ID ? or GUID?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/configuration/ActivitiConfiguration.java
```diff
@@ -0,0 +1,96 @@
+package com.sap.cloud.lm.sl.cf.web.configuration;
```
enchobelezirev: I am with a mixed feelings here... :(
On one side, the creation using the Java Spring configuration fills me with happiness :D
but on the other, it is now a little more complicated to create the activiti shits with such Java configuration?

What do you say?

Can we use the Java Configuration in order to make some magic tricks here? As we can do in the FileSystemFileServiceFacotryBean(or w/e was)

But on the other hand, the next commit uses it in order to create the job-id in more precise way. Anyway, I am staying with mixed feelings JUST in this case, otherwise I prefer JAVA CONFIG

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/configuration/ActivitiConfiguration.java
```diff
@@ -93,4 +100,19 @@ protected void scale(DefaultJobExecutor jobExecutor) {
...
+    public String jobExecutorId(ApplicationConfiguration applicationConfiguration) {
+        String applicationGuid = applicationConfiguration.getApplicationGuid();
+        Integer applicationInstanceIndex = applicationConfiguration.getApplicationInstanceIndex();
+        if (applicationGuid == null || applicationInstanceIndex == null) {
+            return randomIdGenerator.get();
+        }
+        return generateJobExecutorId(applicationGuid, applicationInstanceIndex);
+    }
+
+    private String generateJobExecutorId(String applicationGuid, int applicationInstanceIndex) {
```
enchobelezirev: isnt this building id not generating?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/configuration/ActivitiConfiguration.java
nictas: I'm also not the biggest fan of this class, but it does "fix" a couple of issues, which were present in the XML configuration:
  1. Renaming classes will automatically rename them here as well.
  2. Removes the duplication in the two activiti-context.xml files (one in this project and one in the internal).

The downside is that the class is more verbous than the XML, but considering the other benefits, I'm willing to live with it. :)

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfiguration.java
```diff
@@ -714,53 +716,49 @@ private Boolean shouldUseXSAuditLoggingFromEnvironment() {
...
-            LOGGER.warn(format(Messages.SPACE_ID_NOT_SPECIFIED, DEFAULT_SPACE_ID));
-        } catch (ParsingException e) {
-            LOGGER.warn(format(Messages.INVALID_VCAP_APPLICATION, vcapApplication), e);
+        Map<String, Object> vcapApplication = getVcapApplication();
+        Object orgName = vcapApplication.get("organization_name");
+        if (orgName != null) {
+            LOGGER.info(format(Messages.ORG_NAME, orgName));
+            return orgName.toString();
         }
-        return null;
```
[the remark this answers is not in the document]
nictas: I thought it was a mistake as well, but after taking a second look, the rest of our logic relies on this being 'null', which is inconsistent with getSpaceId(), but still, I'll make sure we return 'null' here. Nice catch! :)

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/message/Messages.java
nictas: GUID was consistent with getSpaceGuid(), but I think I'll rename both methods to getSpaceId() and getApplicationId() respectively, so that they're consistent with their log messages.

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/configuration/ActivitiConfiguration.java
nictas: Mmmm, maybe? I don't really know. :D Part of it is generated, part of it is built. Since 2/3rds of the ID are "built" I'll rename it to buildJobExecutorId.

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/configuration/ActivitiConfiguration.java
enchobelezirev: Yes, I haven't thought about the benefits. I want to live with it, too

PR #188, opened 2018-06-06
Update NOTICE

christopherclark:


NOTICE
```diff
@@ -1 +1,9 @@
-Copyright (c) 2015 SAP SE or an SAP affiliate company. All rights reserved.
+Copyright (c) 2018-Present CloudFoundry.org Foundation, Inc. All Rights Reserved.
+This project contains software that is Copyright (c) 2015-2018 SAP SE or an SAP affiliate company.
```
nvvalchev: Should we keep the SAP reference here?

PR #198, opened 2018-06-18
Add limitation to application upload size (Zipbomb vulnerability fix)

valentinEmpy:


com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/message/Messages.java
```diff
@@ -76,7 +76,7 @@
     public static final String CANT_RUN_TASK_ON_APP = "Could not run task \"{0}\" on application \"{1}\"";
     public static final String CANT_CANCEL_TASK = "Could not cancel task with GUID \"{0}\"";
     public static final String CANT_UPDATE_SERVICE_PLAN = "Could not update service plan of service \"{0}\": \"{1}\"";
-    public static final String ERROR_SIZE_OF_UNCOMPRESSED_FILE_EXCEEDS_MAX_SIZE_LIMIT = "The size \"{0}\" of file \"{1}\" exceeds max size limit \"{2}\"";
+
```
nictas: Delete this empty line.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        try (InputStreamProducer streamProducer = new InputStreamProducer(appArchiveStream, fileName)) {
+            return extract(streamProducer, callback);
+        } catch (IOException | SLException e) {
+            callback.onError(e);
+            throw e;
+        } catch (Exception e) {
+            callback.onError(e);
+            return null;
```
nictas: Why return null here? If this null goes back to line 88 of UploadAppStep, then a NPE will be thrown on line 109.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    public Path extract(ExtractStatusCallback callback) throws IOException, SLException {
+        if (callback == null) {
+            callback = ExtractStatusCallback.NONE;
+        }
+
+        try (InputStreamProducer streamProducer = new InputStreamProducer(appArchiveStream, fileName)) {
+            return extract(streamProducer, callback);
+        } catch (IOException | SLException e) {
```
nictas: Why are IOExceptions and SLExceptions handled differently from the others?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+        InputStream stream = streamProducer.getNextInputStream();
+        if (stream == null) {
+            throw new ContentException(com.sap.cloud.lm.sl.mta.message.Messages.CANNOT_FIND_ARCHIVE_ENTRY, streamProducer.getFileName());
+        }
+
+        if (isFile(streamProducer.getFileName())) {
+            return saveToFile(streamProducer, callback);
+        }
+
+        if (streamProducer.getFileName()
```
nictas: Why does this mean that the stream is a directory? Even if that really is the case can this be made clearer?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    public static Path getTempDirectory(String entryName) throws IOException {
+        String dirName = entryName;
+        if (dirName.endsWith(ARCHIVE_FILE_SEPARATOR)) {
+            dirName = dirName.substring(0, dirName.length() - 1);
+        }
+        dirName = dirName.replace('/', '-');
+        Path dirPath = Paths.get(dirName);
+        if (Files.exists(dirPath)) {
```
nictas: Why check whether it exists? And BTW, wouldn't Files.createTempDirectory create a directory with a random string appended to the prefix that is the "dirName" variable. If yes, then this check doesn't make sense, because you're checking whether directory A exists, but if it doesn't you're creating directory B.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+        return prefix;
+    }
+
+    private String getSuffix(String nameOfFile) {
+        int indexOfDot = nameOfFile.lastIndexOf('.');
+        return (indexOfDot > 0) ? nameOfFile.substring(indexOfDot) : null;
+    }
+
+    public static Path getTempDirectory(String entryName) throws IOException {
```
nictas: If you decide to remove the if (Files.exists(dirPath)) { check from this method, then rename it to createTempDirectory.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private boolean isFile(String fileName) {
+        return !fileName.endsWith(ARCHIVE_FILE_SEPARATOR);
+    }
+
+    private String getPrefix(String nameOfFile) {
+        int indexOfDot = nameOfFile.lastIndexOf('.');
+        String prefix = (indexOfDot > 0) ? nameOfFile.substring(0, indexOfDot) : nameOfFile;
+        if (prefix.length() < 3) {
+            prefix = prefix + "000";
```
nictas: What is the purpose of this?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+        } catch (IOException | SLException e) {
+            callback.onError(e);
+            throw e;
+        } catch (Exception e) {
+            callback.onError(e);
+            return null;
+        }
+    }
+
+    private Path extract(InputStreamProducer streamProducer, ExtractStatusCallback callback) throws IOException, SLException {
```
nictas: This method is not symmetric to the one above it. The first uses a field (appArchiveStream) while this one uses a method argument. Either use a method argument in both methods, or a field.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+        try (OutputStream outputStream = Files.newOutputStream(filePath)) {
+            copy(streamProducer.getInputStream(), outputStream, callback);
+        }
+        LOGGER.debug(MessageFormat.format("Input stream saved to temporary file \"{0}\"", filePath));
+        return filePath;
+    }
+
+    private Path saveToDirectory(InputStreamProducer streamProducer, ExtractStatusCallback callback) throws IOException, SLException {
+        Path dirPath = getTempDirectory(streamProducer.getFileName());
+        callback.onFileCreated(dirPath);
```
nictas: The method name here is a bit misleading. It is called "onFileCreated", but here you're giving a directory instead of a file.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+        LOGGER.debug(MessageFormat.format("Input stream saved to temporary directory \"{0}\"", dirPath));
+        return dirPath;
+    }
+
+    protected Path saveToDirectoryUnordered(InputStreamProducer streamProducer, ExtractStatusCallback callback)
+        throws IOException, SLException {
+        Path dirPath = getTempDirectory(streamProducer.getFileName());
+        callback.onFileCreated(dirPath);
+        do {
+            if (isFile(streamProducer.getStreamEntryName())) {
```
nictas: I don't understand this. Why do this check again, when a similar one is made on line 59. What is the difference between getFileName and getStreamEntryName? How is saveToDirectoryUnordered different than saveToDirectory.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ExtractStatusCallback.java
```diff
@@ -0,0 +1,48 @@
+package com.sap.cloud.lm.sl.cf.process.util;
+
+import java.nio.file.Path;
+
+import com.sap.cloud.lm.sl.common.SLException;
+
+public interface ExtractStatusCallback {
```
nictas: Why use a callback at all? Why not just give the file size limit to ApplicationArchiveExtractor and let him do the check?

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractorTest.java
```diff
@@ -0,0 +1,219 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import com.sap.cloud.lm.sl.cf.core.util.ApplicationConfiguration;
+import com.sap.cloud.lm.sl.cf.process.message.Messages;
+import com.sap.cloud.lm.sl.common.ContentException;
+import com.sap.cloud.lm.sl.common.SLException;
+
+@RunWith(Enclosed.class)
```
nictas: Use JUnit 5 here. The test will be MUCH simpler.

nictas:
I left a lot of comments, but I'm not sure whether the code I'm commenting on is yours or was just moved from another class. If it was moved then you can ignore my comments.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ExtractStatusCallback.java
valentinEmpy: If the file size limit is to be passed to ApplicationArchiveExtractor, it'd mean it has to be aborted (the extraction) and an exception is to be thrown, then the methods will obviously return null and any track of the file will be lost, so cleanup will not be possible. One way to handle this is to cleanup when throwing the exception (in ApplicationArchiveExtractor) but that breaks clean code standards as all exceptions are to be handled in a central place (UploadAppStep). So Callback is used to track the file on creation, and when an error occurs the file is cleaned up. It also is consistent with uploading to the controller (which also uses a Callback). Both callbacks are in UploadAppStep, one for extraction, other for uploading, both handling errors.

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/message/Messages.java
valentinEmpy: The empty line exists for separation reasons, as the above messages are error messages, and the ones below are debug messages. Readability increases.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
valentinEmpy: It is required by File.createTempFile(prefix, suffix) that the prefix is at least 3 characters long otherwise it throws an exception.
This assignment wasn't implemented by me, it was just moved from another class.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
valentinEmpy: If the application is a single file (not a directory), then onFileCreated will be called with the path to the file (this case is handled by saveToFile). Accordingly, same will happen if it is not a single file but directory with multiple files. It will return the path to the root folder (which is a file by itself). I do not find any other name more appropriate?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
valentinEmpy: This is recursive and is called for each file from the directory when extracting, unlike the one on line 59 which checks only a single.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
valentinEmpy: This check is not primarily to ensure that it is a directory. This ensures that the current position of the stream is at a File that is indeed the application binaries for this module. (It happens to be a directory because the check above this if, ensures it).

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ExtractStatusCallback.java
nictas: I'm not sure it would break clean code standards, as the exception itself is handled in UploadAppStep. IMO, the clean-up of a resource, which is created by class A should be the responsibility of class A, at least in case of an error. And even if this does break clean code standards - so what? Isn't it more complex to have a callback infrastructure + an implementation instead of a simple "if" in ApplicationArchiveExtractor with the corresponding clean up? How is the clean code variant "cleaner"? It is more flexible - that's for sure, but do we need this flexibility?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
nictas: Okay.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveExtractor.java
nictas: I still don't understand what's the difference between saveToDirectory and saveToDirectoryUnordered. Maybe I could, by spending several minutes reading it, but the code should express the idea more clearly. There should be no "Aha!" moment when reading the code.

PR #199, opened 2018-06-19
Add monitor resource for free space on file system service

KristianShishoev:


theghost5800:
This feature has been implemented already

PR #210, opened 2018-07-17
Remove TimeoutExecutor

nictas:
The default timeout is so high that nothing will ever timeout, which
makes the executor itself pointless.

The initial purpose of the executor was to prevent Activiti step
timeouts and provide more helpful error messages, but several things
make the executor useless:

1) The default timeout has been increased to 15 minutes, which means
that nothing will ever timeout.
2) Activiti's default step timeout has been increased from 5 to 30
minutes.
3) If the number of calls to the executor within a step is greater than
2, then the step may still timeout, if each call takes close to 14
minutes.

BUG: NGPBUG-60694


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfiguration.java
```diff
@@ -267,16 +265,17 @@ public String getConfiguratioName() {
...
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
     }
 
     private Set<String> getNotSensitiveConfigVariables() {
         return new HashSet<>(Arrays.asList(CFG_TYPE, CFG_TARGET_URL, CFG_DB_TYPE, CFG_PLATFORMS, CFG_PLATFORMS_V2, CFG_PLATFORMS_V3,
             CFG_TARGETS, CFG_TARGETS_V2, CFG_TARGETS_V3, CFG_MAX_UPLOAD_SIZE, CFG_MAX_MTA_DESCRIPTOR_SIZE, CFG_MAX_MANIFEST_SIZE,
             CFG_MAX_RESOURCE_FILE_SIZE, CFG_SCAN_UPLOADS, CFG_USE_XS_AUDIT_LOGGING, CFG_DUMMY_TOKENS_ENABLED, CFG_BASIC_AUTH_ENABLED,
-            CFG_GLOBAL_AUDITOR_USER, CFG_XS_CLIENT_CORE_THREADS, CFG_XS_CLIENT_MAX_THREADS, CFG_XS_CLIENT_QUEUE_CAPACITY,
-            CFG_XS_CLIENT_KEEP_ALIVE, CFG_CONTROLLER_POLLING_INTERVAL, CFG_UPLOAD_APP_TIMEOUT, CFG_SKIP_SSL_VALIDATION,
+            CFG_GLOBAL_AUDITOR_USER, CFG_CONTROLLER_POLLING_INTERVAL, CFG_UPLOAD_APP_TIMEOUT, CFG_SKIP_SSL_VALIDATION,
```
ddonchev: Shouldn't the new auditlog pool variables be listed here?

PR #211, opened 2018-07-17
Enable CTS+ on CF

ttsokov:
JIRA: LMCROSSITXSADEPLOY-695


enchobelezirev:
What avout the CTS_DEPLOY diagram? Should we move it too?

ttsokov:
Currently it is not necessary to expose CTS_DEPLOY diagram. Later we will move back the CTS ProcessType and CtsDeployMetadata into the internal project.

PR #262, opened 2018-09-17
Fix Start/End ProcessListeners to not set process finished state earlier

ttsokov:
…ier.

If there is process with subprocess, there are registered
Start and EndProcessListeners for the parent process and for the subprocess.
In this case the EndProcessListener of the subprocess set the parent
process (correlationId) in finished state earlier, before it is fully
executed. This leads to earlier report for finished state and missing
progress messages in the client. This commit fixes this bug, because
only EndProcessListener of the parent process set the state to finished,
when it is actually fully executed.

JIRA: LMCROSSITXSADEPLOY-1065

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/StartProcessListener.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/StartProcessListener.java
@@ -14,7 +14,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 
 import com.sap.cloud.lm.sl.cf.core.dao.OperationDao;
 import com.sap.cloud.lm.sl.cf.core.util.ApplicationConfiguration;
-import com.sap.cloud.lm.sl.cf.process.Constants;
 import com.sap.cloud.lm.sl.cf.process.message.Messages;
 import com.sap.cloud.lm.sl.cf.process.metadata.ProcessTypeToOperationMetadataMapper;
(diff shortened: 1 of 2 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/EndProcessListener.java
```diff
@@ -56,6 +56,14 @@ protected void writeLogs(DelegateExecution context) throws IOException, FileStor
 
     @Override
     protected void notifyInternal(DelegateExecution context) throws FileStorageException {
+        String correlationId = StepsUtil.getCorrelationId(context);
+        String processInstanceId = context.getProcessInstanceId();
+
+        // Check if this is subprocess
```
enchobelezirev: Why not making the check in the abstract class?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/StartProcessListener.java
```diff
@@ -48,6 +48,15 @@ protected void notifyInternal(DelegateExecution context) {
             correlationId = context.getProcessInstanceId();
             context.setVariable(Constants.VAR_CORRELATION_ID, correlationId);
         }
+
+        correlationId = StepsUtil.getCorrelationId(context);
```
enchobelezirev: Same question here?

PR #298, opened 2018-10-23
Handle incomplete/incorrect MTA metadata in app env

nictas:
JIRA: LMCROSSITXSADEPLOY-1124


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/detect/ApplicationMtaMetadataParser.java
```diff
@@ -31,28 +34,38 @@ public static ApplicationMtaMetadata parseAppMetadata(CloudApplication app) {
 
     private static ApplicationMtaMetadata attemptToParseAppMetadata(CloudApplication app) {
         Map<String, String> appEnv = app.getEnvAsMap();
-        DeployedMtaMetadata mtaMetadata = parseMtaMetadata(appEnv);
+        DeployedMtaMetadata mtaMetadata = parseMtaMetadata(app, appEnv);
```
ddonchev: Why not pass only the app reference without the appEnv map?

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/cf/detect/DeployedComponentsDetectorTest.java
```diff
@@ -1,110 +1,86 @@
...
-    }
+        private Map<Object, Object> env;
+        private String name;
 
-    private List<CloudApplication> toCloudApplications(List<SimpleApplication> simpleApplications) {
-        return simpleApplications.stream()
-            .map((application) -> application.toCloudApplication())
-            .collect(Collectors.toList());
-    }
+        private CloudApplication toCloudApplication() {
```
ddonchev: cool :)

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/detect/ApplicationMtaMetadataParser.java
nictas: getEnvAsMap() is a rather expensive operation, which is why it's extracted at the top of the method.

PR #305, opened 2018-10-25
Prepare migration for flowable for large DBs

boyan-velinov:
Jira: LMCROSSITXSADEPLOY-1252


com.sap.cloud.lm.sl.cf.core/src/main/resources/com/sap/cloud/lm/sl/cf/core/db/changelog/db-changelog-prepare_for_flowable.xml
```diff
@@ -0,0 +1,216 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9"
```
enchobelezirev: What about the sequences? Will they be fine?

boyan-velinov: Yes, the order is defined by order of changeSets

PR #307, opened 2018-10-29
validate service parameters

IvankaTashkova:

### DIFF
--- a/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v1/SpecialResourceTypesRequiredParameters.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v1/SpecialResourceTypesRequiredParameters.java
@@ -1,0 +1,34 @@
+package com.sap.cloud.lm.sl.cf.core.cf.v1;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import com.sap.cloud.lm.sl.cf.core.model.SupportedParameters;
+
+public enum SpecialResourceTypesRequiredParameters {
+
+    MANAGED_SERVICE_REQUIRED_PARAMETERS(ResourceType.MANAGED_SERVICE, SupportedParameters.SERVICE,
+        SupportedParameters.SERVICE_PLAN), USER_PROVIDED_SERVICE_REQUIRED_PARAMETERS(ResourceType.USER_PROVIDED_SERVICE,
+            SupportedParameters.SERVICE_CONFIG), EXISTING_SERVICE_REQUIRED_PARAMETERS(
+                ResourceType.EXISTING_SERVICE), EXISTING_SERVICE_KEY_REQUIRED_PARAMETERS(ResourceType.EXISTING_SERVICE_KEY);
+
+    private final ResourceType resourceType;
+    private final HashSet<String> requiredParameters = new HashSet<>();
+
+    SpecialResourceTypesRequiredParameters(ResourceType resourceType, String... requiredParameter) {
+        this.resourceType = resourceType;
+        for (String supportedParameters : requiredParameter) {
+            requiredParameters.add(supportedParameters);
+        }
+    }
+
+    public static Set<String> getRequiredParameters(ResourceType resourceType) {
+        for (SpecialResourceTypesRequiredParameters value : SpecialResourceTypesRequiredParameters.values()) {
+            if (value.resourceType.equals(resourceType)) {
+                return value.requiredParameters;
+            }
+        }
+        return new HashSet<>();
+    }
+
+}
--- a/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/SpecialResourceTypesRequiredParametersUtil.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/SpecialResourceTypesRequiredParametersUtil.java
@@ -1,0 +1,29 @@
+package com.sap.cloud.lm.sl.cf.core.util;
+
+import java.util.Map;
+import java.util.Set;
+
+import com.sap.cloud.lm.sl.cf.core.activiti.Messages;
+import com.sap.cloud.lm.sl.cf.core.cf.v1.ResourceType;
+import com.sap.cloud.lm.sl.cf.core.cf.v1.SpecialResourceTypesRequiredParameters;
+import com.sap.cloud.lm.sl.common.ContentException;
+
+public class SpecialResourceTypesRequiredParametersUtil {
+
+    public static void checkRequiredParameters(ResourceType resourceType, Map<String, Object> parameters) {
+        Set<String> requiredParameters = getRequiredParameters(resourceType);
+        if (requiredParameters.isEmpty()) {
+            return;
+        }
+        for (String parameter : requiredParameters) {
+            if (!parameters.containsKey(parameter)) {
+                throw new ContentException(Messages.NO_VALUE_GIVEN_FOR_REQUIRED_PARAMETERS);
+            }
+        }
+    }
+
+    protected static Set<String> getRequiredParameters(ResourceType resourceType) {
+        return SpecialResourceTypesRequiredParameters.getRequiredParameters(resourceType);
+    }
+
+}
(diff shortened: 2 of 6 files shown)


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v1/SpecialResourceTypesRequiredParameters.java
```diff
@@ -0,0 +1,34 @@
...
+
+public enum SpecialResourceTypesRequiredParameters {
+
+    MANAGED_SERVICE_REQUIRED_PARAMETERS(ResourceType.MANAGED_SERVICE, SupportedParameters.SERVICE,
+        SupportedParameters.SERVICE_PLAN), USER_PROVIDED_SERVICE_REQUIRED_PARAMETERS(ResourceType.USER_PROVIDED_SERVICE,
+            SupportedParameters.SERVICE_CONFIG), EXISTING_SERVICE_REQUIRED_PARAMETERS(
+                ResourceType.EXISTING_SERVICE), EXISTING_SERVICE_KEY_REQUIRED_PARAMETERS(ResourceType.EXISTING_SERVICE_KEY);
+
+    private final ResourceType resourceType;
+    private final HashSet<String> requiredParameters = new HashSet<>();
```
nictas: Use Set<String> as a type for "requiredParameters" (or even Collection<String>) if you don't need any of the HashSet specific methods. This makes it easy to change the used implementation, if necessary (from HashSet to TreeSet, for example).

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v1/SpecialResourceTypesRequiredParameters.java
```diff
@@ -0,0 +1,34 @@
...
+
+    SpecialResourceTypesRequiredParameters(ResourceType resourceType, String... requiredParameter) {
```
nictas: Why package-private? Enum's should have private constructors. See https://stackoverflow.com/questions/7113363/private-enum-constructor.

Also, rename "requiredParameter" to "requiredParameters", since the variable technically can contain multiple parameters.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v1/SpecialResourceTypesRequiredParameters.java
```diff
@@ -0,0 +1,34 @@
+package com.sap.cloud.lm.sl.cf.core.cf.v1;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import com.sap.cloud.lm.sl.cf.core.model.SupportedParameters;
+
+public enum SpecialResourceTypesRequiredParameters {
```
nictas: I would actually merge this into the ResourceType enum. Then, the getRequiredParameters() method would be much more simple. Actually, it will be just a getter. :)

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v1/SpecialResourceTypesRequiredParameters.java
```diff
@@ -0,0 +1,34 @@
...
+        }
+    }
+
+    public static Set<String> getRequiredParameters(ResourceType resourceType) {
+        for (SpecialResourceTypesRequiredParameters value : SpecialResourceTypesRequiredParameters.values()) {
+            if (value.resourceType.equals(resourceType)) {
+                return value.requiredParameters;
+            }
+        }
+        return new HashSet<>();
```
nictas: This wouldn't be necessary if you merge this class into the ResourceType enum, but if somewhere you need to return an empty collection, just use Collections.emptySet(). It isn't a big deal, but in the long term it saves a lot of unnecessary garbage collection, because Collections.emptySet() does not create a new object.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/SpecialResourceTypesRequiredParametersUtil.java
```diff
@@ -0,0 +1,29 @@
...
+import com.sap.cloud.lm.sl.cf.core.activiti.Messages;
+import com.sap.cloud.lm.sl.cf.core.cf.v1.ResourceType;
+import com.sap.cloud.lm.sl.cf.core.cf.v1.SpecialResourceTypesRequiredParameters;
+import com.sap.cloud.lm.sl.common.ContentException;
+
+public class SpecialResourceTypesRequiredParametersUtil {
+
+    public static void checkRequiredParameters(ResourceType resourceType, Map<String, Object> parameters) {
+        Set<String> requiredParameters = getRequiredParameters(resourceType);
+        if (requiredParameters.isEmpty()) {
```
nictas: This "if" is not necessary. The for-each below would handle this case.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/SpecialResourceTypesRequiredParametersUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+import com.sap.cloud.lm.sl.cf.core.cf.v1.SpecialResourceTypesRequiredParameters;
+import com.sap.cloud.lm.sl.common.ContentException;
+
+class SpecialResourceTypesRequiredParametersUtilTest {
+
+    @Test
+    void getRequiredParametersForManagedService() {
+        ResourceType resourceType = ResourceType.MANAGED_SERVICE;
+        Set<String> expectedRequiredParameters = SpecialResourceTypesRequiredParameters.getRequiredParameters(resourceType);
+        Set<String> actualRequiredParameters = SpecialResourceTypesRequiredParametersUtil.getRequiredParameters(resourceType);
```
nictas: You should only test the public methods of a class (black box testing). Private and protected methods are an implementation detail and change more regularly than the public methods, which would make their tests very fragile.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/SpecialResourceTypesRequiredParametersUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertEquals(expectedRequiredParameters, actualRequiredParameters);
+    }
+    
+    @Test
+    void checkRequiredParametersForManagedServiceWithNoParameters() {
+        Map<String, Object> dummyParameters = new HashMap<>();
+        ResourceType resourceType = ResourceType.MANAGED_SERVICE;
+        Assertions.assertThrows(ContentException.class, () -> SpecialResourceTypesRequiredParametersUtil.checkRequiredParameters(resourceType, dummyParameters));
```
nictas: Nice!

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/SpecialResourceTypesRequiredParametersUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Test
+    void checkRequiredParametersForExistingServiceWithNoParameters() {
+        Map<String, Object> dummyParameters = new HashMap<>();
+        ResourceType resourceType = ResourceType.EXISTING_SERVICE;
+        Assertions.assertDoesNotThrow(() -> SpecialResourceTypesRequiredParametersUtil.checkRequiredParameters(resourceType, dummyParameters));
+    }
+
```
nictas: Add tests for user-provided services.

nictas:
I'm closing this pull request, as there's another for the same bug fix:
https://github.com/cloudfoundry-incubator/multiapps-controller/pull/309

PR #310, opened 2018-10-30
Replace OpenJPA with EclipseLink

nictas:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/dao/ConfigurationSubscriptionDtoDao.java
```diff
@@ -27,9 +26,8 @@
 @Component
 public class ConfigurationSubscriptionDtoDao {
 
-    @Autowired
-    @Qualifier("configurationSubscriptionEntityManagerFactory")
-    protected EntityManagerFactory emf;
+    @Inject
+    protected EntityManagerFactory entityManagerFactory;
```
enchobelezirev: Use constructor?

PR #316, opened 2018-11-05
Adapt to deployed-after attribute. Introduce parallel deployments. Refactor process logging.

valentinEmpy:
#### Description: 

Introduces parallel deployments and refactors process logging

LMCROSSITXSADEPLOY-24


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+import com.sap.cloud.lm.sl.cf.client.lib.domain.CloudApplicationExtended;
+import com.sap.cloud.lm.sl.cf.core.security.serialization.SecureSerializationFacade;
+import com.sap.cloud.lm.sl.cf.process.message.Messages;
+
+@Component("computeNextAppsStep")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class ComputeNextAppsStep extends SyncActivitiStep {
```
enchobelezirev: Refactor a little

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v3/ApplicationsCloudModelBuilder.java
```diff
@@ -53,44 +47,13 @@ public ApplicationsCloudModelBuilder(DeploymentDescriptor deploymentDescriptor,
     protected HandlerFactory createHandlerFactory() {
         return new HandlerFactory(MTA_MAJOR_VERSION);
     }
-    
+
     @Override
     protected CloudApplicationExtended getApplication(com.sap.cloud.lm.sl.mta.model.v1.Module module) {
-        List<Map<String, Object>> parametersList = parametersChainBuilder.buildModuleChain(module.getName());
```
enchobelezirev: Remove it from the previous commit

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerHandler.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final String LOG_FILE_EXTENSION = ".log";
+
+    @Inject
+    private ProcessLogsPersistenceService processLogsPersistenceService;
+
+    private Map<String, ProcessLogger> loggers = new HashMap<>();
+
+    private static ProcessLoggerHandler loggerHandler;
+
+    public static ProcessLoggerHandler getInstance() {
```
enchobelezirev: Why making it Singleton - you are using Spring?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerHandler.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final org.slf4j.Logger LOGGER = LoggerFactory.getLogger(ProcessLoggerHandler.class);
+    private static final String LOG_LAYOUT = "#2.0#%d{yyyy MM dd HH:mm:ss.SSS}#%d{XXX}#%p#%c#%n%X{MsgCode}#%X{CSNComponent}#%X{DCComponent}##%X{DSRCorrelationId}#%X{Application}#%C#%X{User}#%X{Session}#%X{Transaction}#%X{DSRRootContextId}#%X{DSRTransaction}#%X{DSRConnection}#%X{DSRCounter}#%t##%X{ResourceBundle}#%n%m#%n%n";
+    public static final String DEFAULT_LOG_NAME = "MAIN_LOG";
+    private static final String DEFAULT_LOG_DIR = "logs";
+    private static final String LOG_FILE_EXTENSION = ".log";
+
+    @Inject
+    private ProcessLogsPersistenceService processLogsPersistenceService;
+
+    private Map<String, ProcessLogger> loggers = new HashMap<>();
```
enchobelezirev: ConcurrentHashMap??

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerHandler.java
```diff
@@ -0,0 +1,168 @@
...
+        try {
+            logger.appendLogContent(context, processLogsPersistenceService);
+        } catch(FileStorageException e) {
+            LOGGER.warn(MessageFormat.format(Messages.COULD_NOT_PERSIST_LOGS_FILE, e.getMessage()), e);
+        } finally {
+            logger.deleteLogFile();
+        }
+    }
+
+    public ProcessLogsPersistenceService getProcessLogsPersistenceService() {
```
enchobelezirev: https://pics.me.me/delet-di-s-32240102.png

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLogsPersistenceService.java
```diff
@@ -111,13 +112,55 @@ private void appendLog(final String space, final String namespace, final String
...
+                        FileUploader.removeFile(fileUpload);
+                    }
+                }
             }
+
         };
         processFileContent(new DefaultFileDownloadProcessor(space, fileId, fileProcessor));
     }
 
+    private FileUpload uploadFile(final InputStream in, InputStream logFile) throws FileStorageException {
```
enchobelezirev: What is this? Why doing this?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v1/ApplicationsCloudModelBuilder.java
```diff
@@ -99,10 +100,21 @@ protected ApplicationEnvironmentCloudModelBuilder createApplicationEnvironmentCl
...
+        }
+        return apps;
+    }
+
+    private List<CloudApplicationExtended> resolveModules(Set<String> mtaModulesInArchive, Set<String> allMtaModules,
+        Set<String> deployedModules, SortedSet<String> unresolvedMtaModules) {
+        List<CloudApplicationExtended> apps = new ArrayList<>();
+        for (Module module : handler.getModulesForDeployment(deploymentDescriptor, SupportedParameters.PARALLEL_DEPLOYMENTS,
+            SupportedParameters.DEPENDENCY_TYPE, DEPENDENCY_TYPE_HARD)) {
             if (!mtaModulesInArchive.contains(module.getName()) || module.getType() == null) {
```
enchobelezirev: I am hurt when I see this logic below. As you are changing it a bit, is there any way to refactor it a bit?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v3/ApplicationsCloudModelBuilder.java
```diff
@@ -25,69 +31,89 @@
...
     @Override
     protected HandlerFactory createHandlerFactory() {
         return new HandlerFactory(MTA_MAJOR_VERSION);
     }
-    
+
+    @Override
+    protected CloudApplicationExtended getApplication(com.sap.cloud.lm.sl.mta.model.v1.Module module) {
+        CloudApplicationExtended app = super.getApplication(module);
+        Set<String> deployedAfter = emptyIfNull(((Module) module).getDeployedAfter()).stream()
```
enchobelezirev: Extract the Set conversion to a separate variable in order to be more readable

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v3/ApplicationsCloudModelBuilder.java
```diff
@@ -25,69 +31,89 @@
...
+
     public ApplicationsCloudModelBuilder(DeploymentDescriptor deploymentDescriptor, CloudModelConfiguration configuration,
         DeployedMta deployedMta, SystemParameters systemParameters, XsPlaceholderResolver xsPlaceholderResolver, String deployId,
         UserMessageLogger userMessageLogger) {
         super(deploymentDescriptor, configuration, deployedMta, systemParameters, xsPlaceholderResolver, deployId, userMessageLogger);
     }
-    
+
+    @Override
+    public DeploymentMode getDeploymentMode() {
```
enchobelezirev: Little confused, why the ApplicationCloudModelBuilder has a method getDeploymentMode? Isn't it meant to be placed in a more central place?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+import com.sap.cloud.lm.sl.cf.client.lib.domain.CloudApplicationExtended;
+import com.sap.cloud.lm.sl.cf.core.security.serialization.SecureSerializationFacade;
+import com.sap.cloud.lm.sl.cf.process.message.Messages;
+
+@Component("computeNextAppsStep")
```
enchobelezirev: maybe computeNextAppsForDeployment?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    protected StepPhase executeStep(ExecutionWrapper execution) throws Exception {
+        getStepLogger().debug(Messages.COMPUTING_NEXT_APPS_FOR_PARALLEL_DEPLOYMENT);
+        List<CloudApplicationExtended> allApplicationsToDeploy = StepsUtil.getAppsToDeploy(execution.getContext());
+        List<CloudApplicationExtended> completedApplications = StepsUtil.getCompletedAppsInParallel(execution.getContext());
+
+        // Set next iteration data
+        List<CloudApplicationExtended> appsForNextIteration = getApplicationsForNextIteration(allApplicationsToDeploy,
+            completedApplications.stream()
```
enchobelezirev: Extract to a separate variable

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,55 @@
...
+                    .equals(app.getModuleName())));
+        StepsUtil.setIsLastParallelIteration(execution.getContext(), isLastIteration);
+        getStepLogger().debug(Messages.COMPUTED_NEXT_APPS_FOR_PARALLEL_DEPLOYMENT);
+        return StepPhase.DONE;
+    }
+
+    private List<CloudApplicationExtended> getApplicationsForNextIteration(List<CloudApplicationExtended> allApplicationsToDeploy,
+        List<String> completedModules) {
+        return allApplicationsToDeploy.stream()
+            .filter(app -> !completedModules.contains(app.getModuleName()) && (app.getDeployedAfter()
```
enchobelezirev: Could first, the completed to be removed from the list of applications?
Then, extract all applications which have deployedAfter which contains one of the completedModules?
Isn't it be simpler that way?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,55 @@
...
+        // Set next iteration data
+        List<CloudApplicationExtended> appsForNextIteration = getApplicationsForNextIteration(allApplicationsToDeploy,
+            completedApplications.stream()
+                .map(CloudApplicationExtended::getModuleName)
+                .collect(Collectors.toList()));
+        getStepLogger().debug(Messages.APPS_FOR_NEXT_PARALLEL_ITERATION, secureSerializer.toJson(appsForNextIteration));
+        StepsUtil.setAppsToDeployInParallel(execution.getContext(), appsForNextIteration);
+
+        // Mark next iteration data as completed
+        completedApplications.addAll(appsForNextIteration);
```
enchobelezirev: Why do it here?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,55 @@
...
+        StepsUtil.setCompletedAppsInParallel(execution.getContext(), completedApplications);
+
+        boolean isLastIteration = allApplicationsToDeploy.stream()
```
enchobelezirev: Whaaat ? Why we need to check whether this is the last operation?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/AbstractFileService.java
```diff
@@ -139,7 +139,7 @@ private FileEntry addFile(String space, String namespace, String name,
         return fileEntry;
     }
 
-    private void scanUpload(FileUpload file) throws FileStorageException {
+    protected void scanUpload(FileUpload file) throws FileStorageException {
```
enchobelezirev: why protected?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLogsPersistenceService.java
```diff
@@ -111,13 +112,55 @@ private void appendLog(final String space, final String namespace, final String
...
+            if (fileUploadProcessor.shouldScanFile()) {
+                scanUpload(fileUpload);
+            }
+            return fileUpload;
+        } finally {
+            IOUtils.closeQuietly(inputStream);
+        }
+    }
+
+    private void addFile(final String space, final String namespace, final String logName, Connection connection, FileUpload fileUpload)
```
enchobelezirev: I believe such method already exists in the Abstract class?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/SupportedParameters.java
```diff
@@ -34,6 +34,7 @@
     public static final String GENERATED_PASSWORD = "generated-password";
     public static final String DEFAULT_IDLE_DOMAIN = "default-idle-domain";
     public static final String TIMESTAMP = "timestamp";
+    public static final String PARALLEL_DEPLOYMENTS = "parallel-deployments";
```
nictas: Rename this to "enable-parallel-deployments". This would be consistent with the "enable-ssh" parameter. Actually, after reviewing the ComputeNextAppsStep - rename this to "enable-parallel-operations", since "operations" groups "deployments" and "restarts".

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
```diff
@@ -80,6 +80,9 @@
     public static final String VAR_SERVICE_KEYS_TO_CREATE = "serviceKeysToCreate";
     public static final String VAR_SERVICE_KEYS_CREDENTIALS_TO_INJECT = "serviceKeysCredentialsToInject";
     public static final String VAR_APPS_TO_DEPLOY = "appsToDeploy";
+    public static final String VAR_DEPLOYMENT_MODE = "deploymentMode";
+    public static final String VAR_COMPLETED_APPS_IN_PARALLEL = "completedAppsInParallel";
```
nictas: What does "completed" mean? Shouldn't it be "deployed"?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/message/Messages.java
```diff
@@ -332,6 +332,9 @@
     public static final String SERVICE_BINDINGS_EXISTS = "Service bindings \"{0}\" exists";
     public static final String PREPARING_APPS_DEPLOYMENT = "Preparing applications deployment...";
     public static final String APPS_DEPLOYMENT_PREPARED = "Applications deployment prepared";
+    public static final String COMPUTING_NEXT_APPS_FOR_PARALLEL_DEPLOYMENT = "Computing applications for next parallel deployment...";
+    public static final String APPS_FOR_NEXT_PARALLEL_ITERATION = "Applications for next parallel iteration: {0}";
```
nictas: Merge this message into the one below and use a consistent term: either "deployment" or "iteration"

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
nictas: Nice catch! I did not even consider that we could use the parallel feature for restarts during blue-green deployments. However, if you're using it for that purpose, then you also need to change the messages. Right now they include the term "deployment", which is not correct in that case.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,52 @@
...
+
+        // Mark next iteration data as computed
+        completedApplications.addAll(appsForNextIteration);
+        StepsUtil.setCompletedAppsInParallel(execution.getContext(), completedApplications);
+
+        getStepLogger().debug(Messages.COMPUTED_NEXT_APPS_FOR_PARALLEL_DEPLOYMENT);
+        return StepPhase.DONE;
+    }
+
+    private List<CloudApplicationExtended> getApplicationsForNextIteration(List<CloudApplicationExtended> allApplicationsToDeploy,
```
nictas: Rename to "computeApplicationsForNextIteration".

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PrepareAppsRestartStep.java
```diff
@@ -22,6 +24,8 @@ protected StepPhase executeStep(ExecutionWrapper execution) {
             .setVariable(Constants.EXECUTE_ONE_OFF_TASKS, false);
         StepsUtil.setUseIdleUris(execution.getContext(), false);
         StepsUtil.setSkipUpdateConfigurationEntries(execution.getContext(), false);
+        
+        StepsUtil.setCompletedAppsInParallel(execution.getContext(), Collections.emptyList());
```
nictas: This default should be in the getter IMO. That way people would never forget to set it.

valentinEmpy: I am not sure I understood this statement but the context already includes all applications after the deployment, so this is necessary to be set to empty in order for restart iteration in bg-deploy to reiterate them.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/SupportedParameters.java
```diff
@@ -34,6 +34,7 @@
     public static final String GENERATED_PASSWORD = "generated-password";
     public static final String DEFAULT_IDLE_DOMAIN = "default-idle-domain";
     public static final String TIMESTAMP = "timestamp";
+    public static final String ENABLE_PARALLEL_DEPLOYMENTS = "enable-parallel-deployments";
```
nictas: Why "deployments"? I thought we agreed upon "operations".

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,51 @@
...
+    protected StepPhase executeStep(ExecutionWrapper execution) throws Exception {
+        getStepLogger().debug(Messages.COMPUTING_NEXT_APPS_FOR_PARALLEL_ITERATION);
+        List<CloudApplicationExtended> allApplicationsToDeploy = StepsUtil.getAppsToDeploy(execution.getContext());
+        List<CloudApplicationExtended> completedApplications = StepsUtil.getIteratedAppsInParallel(execution.getContext());
+        List<String> completedModuleNames = completedApplications.stream()
+            .map(CloudApplicationExtended::getModuleName)
+            .collect(Collectors.toList());
+
+        // Set next iteration data
+        List<CloudApplicationExtended> appsForNextIteration = computeApplicationsForNextIteration(allApplicationsToDeploy,
```
nictas: Use a consistent term please - either "apps" or "applications".

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ComputeNextAppsStep.java
```diff
@@ -0,0 +1,51 @@
...
+
+        getStepLogger().debug(Messages.COMPUTED_NEXT_APPS_FOR_PARALLEL_ITERATION, secureSerializer.toJson(appsForNextIteration));
+        return StepPhase.DONE;
+    }
+
+    private List<CloudApplicationExtended> computeApplicationsForNextIteration(List<CloudApplicationExtended> allApplicationsToDeploy,
+        List<String> completedModules) {
+        allApplicationsToDeploy.removeIf(app -> completedModules.contains(app.getModuleName()));
+        return allApplicationsToDeploy.stream()
+            .filter(app -> (app.getDeployedAfter()
```
nictas: Extract the predicate in this filter in a separate method.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PrepareAppsRestartStep.java
```diff
@@ -22,6 +24,8 @@ protected StepPhase executeStep(ExecutionWrapper execution) {
             .setVariable(Constants.EXECUTE_ONE_OFF_TASKS, false);
         StepsUtil.setUseIdleUris(execution.getContext(), false);
         StepsUtil.setSkipUpdateConfigurationEntries(execution.getContext(), false);
+        
+        StepsUtil.setIteratedAppsInParallel(execution.getContext(), Collections.emptyList());
```
nictas: Why is this still here? You should be able to remove it after providing a default value in the getter.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -406,6 +407,36 @@ public static void setAppsToDeploy(DelegateExecution context, List<CloudApplicat
...
         context.setVariable(Constants.VAR_APPS_TO_DEPLOY, cloudApplicationsAsStrings);
     }
+    
+    @SuppressWarnings("unchecked")
+    public static List<CloudApplicationExtended> getIteratedAppsInParallel(DelegateExecution context) {
+        Object appsDeployedInParallel = context.getVariable(Constants.VAR_ITERATED_APPS_IN_PARALLEL);
+        if (appsDeployedInParallel == null) {
+            return new ArrayList<>();
+        }
+        List<String> cloudApplicationsAsStrings = (List<String>) appsDeployedInParallel;
```
nictas: Do the cast on the first line of the method. null can be cast to anything.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -406,6 +407,36 @@ public static void setAppsToDeploy(DelegateExecution context, List<CloudApplicat
             .collect(Collectors.toList());
         context.setVariable(Constants.VAR_APPS_TO_DEPLOY, cloudApplicationsAsStrings);
     }
+    
+    @SuppressWarnings("unchecked")
+    public static List<CloudApplicationExtended> getIteratedAppsInParallel(DelegateExecution context) {
+        Object appsDeployedInParallel = context.getVariable(Constants.VAR_ITERATED_APPS_IN_PARALLEL);
+        if (appsDeployedInParallel == null) {
+            return new ArrayList<>();
```
nictas: This line makes the method know more than it should. It "knows" that people would add things to this collection. Replace this line with "return Collections.emptyList();" and use ListUtils.union in the step where you're adding more apps to the collection.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -406,6 +407,36 @@ public static void setAppsToDeploy(DelegateExecution context, List<CloudApplicat
...
+    
+    @SuppressWarnings("unchecked")
+    public static List<CloudApplicationExtended> getIteratedAppsInParallel(DelegateExecution context) {
+        Object appsDeployedInParallel = context.getVariable(Constants.VAR_ITERATED_APPS_IN_PARALLEL);
+        if (appsDeployedInParallel == null) {
+            return new ArrayList<>();
+        }
+        List<String> cloudApplicationsAsStrings = (List<String>) appsDeployedInParallel;
+        return cloudApplicationsAsStrings.stream()
+            .map(app -> (CloudApplicationExtended) JsonUtil.fromJson(app, CloudApplicationExtended.class))
```
nictas: Why do you need the cast? Isn't JsonUtil.fromJson casting the result to the type of the "class" object? If you don't need it, please use a method reference instead of a lambda.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerHandler.java
```diff
@@ -0,0 +1,155 @@
...
+        logger.addAppender(createAppender(logger.getLevel(), logFile, customLayout));
+        return new ProcessLogger(logger, logFile, logName, processId, activityId);
+    }
+
+    private Logger createLogger(String name) {
+        Logger logger;
+        logger = Logger.getLogger(name);
+        Level level = logger.getParent()
+            .getLevel();
+        logger.setLevel(level);
```
nictas: But isn't this the default?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerHandler.java
```diff
@@ -0,0 +1,155 @@
...
+        return logger;
+    }
+
+    private File getLocalFile(String logId, String name, String nameSuffix) {
+        return new java.io.File(DEFAULT_LOG_DIR, getLocalLogFileName(logId, name, nameSuffix));
```
nictas: Why are you specifying the full class name when the class is imported?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    private Logger createLogger(String name) {
+        Logger logger;
+        logger = Logger.getLogger(name);
+        Level level = logger.getParent()
+            .getLevel();
+        logger.setLevel(level);
+        return logger;
+    }
+
+    private File getLocalFile(String logId, String name, String nameSuffix) {
```
nictas: Why are you calling the processId "logId" and logName "name" here? I'm having a tough time following the flow of information through the methods.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerHandler.java
```diff
@@ -0,0 +1,155 @@
...
+        for (ProcessLogger logger : getExistingLoggers(getCorrelationId(context), context.getCurrentActivityId())) {
+            appendLoggerContent(context, logger, processLogsPersistenceService);
+            loggers.remove(logger.getName());
+        }
+    }
+
+    private List<ProcessLogger> getExistingLoggers(String processId, String activityId) {
+        return loggers.values()
+            .stream()
+            .filter(logger -> logger.getProcessId()
```
nictas: Extract the predicate in this filter in a separate method.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerHandler.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+        appender.setLayout(layout);
+        appender.setFile(logFile.getAbsolutePath());
+        appender.setThreshold(level);
+        appender.setAppend(true);
+        appender.activateOptions();
+        return appender;
+    }
+
+    public void appendProcessLog(DelegateExecution context) {
```
nictas: This method does not belong in this factory class.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerProviderFactory.java
```diff
@@ -1,212 +0,0 @@
...
-import org.apache.log4j.PatternLayout;
-import org.apache.log4j.spi.NOPLogger;
-import org.apache.log4j.varia.NullAppender;
-import org.flowable.engine.delegate.DelegateExecution;
-import org.slf4j.LoggerFactory;
-
-import com.sap.cloud.lm.sl.cf.persistence.message.Constants;
-import com.sap.cloud.lm.sl.cf.persistence.message.Messages;
-
-public class ProcessLoggerProviderFactory {
```
nictas: https://memegenerator.net/img/instances/62582777.jpg

May the lord (Dido Hristov) have mercy on your soul.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ProcessStepHelper.java
```diff
@@ -29,34 +29,30 @@
...
 
-        processLoggerProviderFactory.removeAll();
-        // Write the log messages:
-        try {
-            processLoggerProviderFactory.append(context);
-        } catch (IOException | FileStorageException e) {
-            LOGGER.warn(MessageFormat.format(Messages.COULD_NOT_PERSIST_LOGS_FILE, e.getMessage()), e);
-        }
-
+        processLoggerHandler.appendProcessLog(context);
```
nictas: Why are the loggers not removed? Are they kept in the cache forever? This would eventually lead to a memory leak.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/SupportedParameters.java
valentinEmpy: For us, as developers it makes more sense to indeed change it to "operations", because we know the bg-deploy internally uses the app-deploy process to restart apps but for users I think it makes more sense to define it as "deployments" because they start a bg-deploy process and simply throughout the output they cannot make a difference between the prevalidation deployment and postvalidation "deployment". If I was a user, I would interpret the whole process as simply "deployment". And this parameter heavily impacts user experience. Wouldn't changing it to operations bring questions like "This is set to parallel ops, then why isn't the creation of services parallel too? It is an operation...". Same question applies for a lot of other stuff. What do you think?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PrepareAppsRestartStep.java
valentinEmpy: Read the previous comment on this. This is not a default value. It is there to "reset" the iteration from the deployment operation in order for restart call activity to reiterate apps. Getter would return default value only if there was nothing in the context. In this case there is (all apps) and we need to reset that.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/executors/SqlQueryExecutor.java
```diff
@@ -0,0 +1,65 @@
...
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.List;
+
+import javax.sql.DataSource;
+
+import com.sap.cloud.lm.sl.cf.persistence.query.SqlQuery;
+import com.sap.cloud.lm.sl.cf.persistence.util.JdbcUtil;
+
+public class SqlQueryExecutor {
```
nictas: After a long discussion with @enchobelezirev we came to the conclusion that this refactoring is not necessary and that the issue can be solved in a simpler way. Let's sync this in person tomorrow.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v3/ApplicationsCloudModelBuilder.java
```diff
@@ -25,69 +31,90 @@
...
     protected HandlerFactory createHandlerFactory() {
         return new HandlerFactory(MTA_MAJOR_VERSION);
     }
-    
+
+    @Override
+    protected CloudApplicationExtended getApplication(com.sap.cloud.lm.sl.mta.model.v1.Module module) {
+        CloudApplicationExtended app = super.getApplication(module);
+        List<String> deployedAfter = emptyIfNull(((Module) module).getDeployedAfter());
+        Set<String> deployedAfterSet = deployedAfter.stream()
```
boyan-velinov: You can avoid streams when creating new HashSet<String>(deployedAfter);

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v3/ApplicationsCloudModelBuilder.java
```diff
@@ -25,69 +31,90 @@
...
     @Override
     protected HandlerFactory createHandlerFactory() {
         return new HandlerFactory(MTA_MAJOR_VERSION);
     }
-    
+
+    @Override
+    protected CloudApplicationExtended getApplication(com.sap.cloud.lm.sl.mta.model.v1.Module module) {
+        CloudApplicationExtended app = super.getApplication(module);
+        List<String> deployedAfter = emptyIfNull(((Module) module).getDeployedAfter());
```
nictas: Here you're converting a set to a list. Use SetUtils.emptyIfNull instead of ListUtils.emptyIfNull. And don't static import it.

Btw, [??](http://s3-eu-west-1.amazonaws.com/malikafavre/project_large_images/69_image1.png?mtime=1433245318)'th comment. :D

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PrepareAppsRestartStep.java
nictas: Aaaaaahhhh, okaay! Now it makes sense. However, this is also a bit dirty, because this step is basically cleaning up all the shit from the previous steps... Can't this be avoided?

Btw, I meant that if the getter returned a default (Collections.emptyList()) in case the value in the context was null, then you wouldn't need to set it to Collections.emptyList() explicitly here just to avoid a NPE. However, I was under the assumption that this is the first place where we used this variable, which is apparently not true.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/SupportedParameters.java
nictas: I can see where you're coming from. Some other people have also given alternative names, so let's all vote and decide the final name after the Scrum meeting tomorrow.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v3/ApplicationsCloudModelBuilder.java
valentinEmpy: I could use SetUtils.emptyIfNull if getDeployedAfter returned a set, but that is not the case, therefore the usage of ListUtils.emptyIfNull. And the conversion to set should happen after the value is ensured not to be null.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
valentinEmpy: No, I need to explicitly assign it, fromJson ensures that the string will be deserialized to CloudApplicationExtended, but the return type is not the same. Assignment is needed. Check out the method.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLogger.java
```diff
@@ -0,0 +1,83 @@
...
+    public ProcessLogger(Logger logger, File log, String logName, String processId, String activityId) {
+        super(logger.getName());
+        this.logger = logger;
+        this.log = log;
+        this.logName = logName;
+        this.processId = processId;
+        this.activityId = activityId;
+    }
+
+    @Override
```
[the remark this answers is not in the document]
valentinEmpy: A decorator-like ProcessLogger in order to easily track which logger is for which process/activity, and which file it logs to. Additionally added functionalities for deletion of file etc..
The methods need to be overriden because of the way Logger is created (Logger.getLogger), adding the appender, etc.. Overriden in order to use the originally created logger.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLogsPersistenceService.java
```diff
@@ -116,79 +78,97 @@ public String findFileId(String space, String namespace, String fileName) throws
...
+                updateLogFile(space, fileId, localLog, remoteLogStream);
             }
         };
         processFileContent(new DefaultFileDownloadProcessor(space, fileId, fileProcessor));
     }
 
+    private void storeLogFile(final String space, final String namespace, final String remoteLogName, File localLog)
+        throws FileStorageException {
+        InputStream inputStream = null;
+        try {
```
nictas: Use try-with-resources.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLogsPersistenceService.java
```diff
@@ -116,79 +78,97 @@ public String findFileId(String space, String namespace, String fileName) throws
...
+            inputStream = new FileInputStream(localLog);
+            FileInfo localLogFileInfo = new FileInfo(localLog, BigInteger.valueOf(localLog.length()),
```
nictas: Extract this and the following two lines in a separate method.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLogsPersistenceService.java
```diff
@@ -116,79 +78,97 @@ public String findFileId(String space, String namespace, String fileName) throws
...
+            throw new FileStorageException(MessageFormat.format(Messages.ERROR_STORING_LOG_FILE, localLog.getName()), e);
+        } finally {
+            IOUtils.closeQuietly(inputStream);
+        }
+    }
+
+    private void updateLogFile(String space, String fileId, File localLog, InputStream remoteLogStream) throws FileStorageException {
+        InputStream localLogStream = null;
+        InputStream appendedLog = null;
+        try {
```
nictas: Use try-with-resources.

PR #318, opened 2018-11-05
Validate service parameters fix

IvankaTashkova:
#### Description: 
Adding serviceName to exception message


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/SpecialResourceTypesRequiredParametersUtil.java
```diff
@@ -12,7 +12,7 @@ public static void checkRequiredParameters(ResourceType resourceType, Map<String
         Set<String> requiredParameters = resourceType.getRequiredParameters();
         for (String parameter : requiredParameters) {
             if (!parameters.containsKey(parameter)) {
-                throw new ContentException(Messages.MISSING_REQUIRED_PARAMETERS, parameter);
+                throw new ContentException(Messages.MISSING_REQUIRED_PARAMETERS,resourceType.getName(), parameter);
```
nictas: This must be the service name and not the resource type's name.

PR #319, opened 2018-11-05
Remove service bindings and service keys that prevent service deletion

IvankaTashkova:
#### Description: 
Unbind service bindings and delete service keys in order to allow deletion of the services.

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
@@ -9,9 +9,11 @@ import org.cloudfoundry.client.lib.CloudControllerException;
 import org.cloudfoundry.client.lib.CloudException;
 import org.cloudfoundry.client.lib.CloudOperationException;
 import org.cloudfoundry.client.lib.CloudServiceBrokerException;
+import org.cloudfoundry.client.lib.domain.CloudApplication;
 import org.cloudfoundry.client.lib.domain.CloudService;
 import org.cloudfoundry.client.lib.domain.CloudServiceBinding;
 import org.cloudfoundry.client.lib.domain.CloudServiceInstance;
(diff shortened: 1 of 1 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -63,17 +65,35 @@ private void deleteService(CloudControllerClient client, String serviceName) {
...
+            deleteServiceKeyFromService(client, serviceName, serviceKey);
         }
 
         getStepLogger().info(Messages.DELETING_SERVICE, serviceName);
         client.deleteService(serviceName);
         getStepLogger().debug(Messages.SERVICE_DELETED, serviceName);
     }
 
+    private void deleteServiceKeyFromService(CloudControllerClient client, String serviceName, ServiceKey serviceKey) {
+        getStepLogger().info(Messages.DELETING_SERVICE_KEY_FOR_SERVICE, serviceKey, serviceName);
```
boyan-velinov: Should be serviceKey.getName() for proper logging

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -63,17 +65,35 @@ private void deleteService(CloudControllerClient client, String serviceName) {
 
     private void attemptToDeleteService(CloudControllerClient client, CloudServiceInstance serviceInstance, String serviceName) {
         List<CloudServiceBinding> bindings = serviceInstance.getBindings();
+        List<CloudApplication> applications = client.getApplications();
```
boyan-velinov: This will return all applications in the space, not only the application that are bound to the service. It will end up in error in unbindServiceFromApplication().
You can check CreateOrUpdateServicesStep::getBoundAppNames which has similar logic.

PR #322, opened 2018-11-08
Fix default port calulation

boyan-velinov:
Jira: LMCROSSITXSADEPLOY-1298


com.sap.cloud.lm.sl.cf.core/enhance.xml
```diff
@@ -0,0 +1,26 @@
+<project name="jpa_enhance_builder" default="enhance">
```
nictas: https://pics.me.me/delet-dis-34029456.png

boyan-velinov: Done

PR #329, opened 2018-11-13
Warn for leftover services after undeploy

IvankaTashkova:
#### Description: 
Warn user with a simple message for leftover services after undeploy when  --delete-services flag is not used.

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/message/Messages.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/message/Messages.java
@@ -136,7 +136,6 @@ public class Messages {
     public static final String ERROR_DELETING_OPTIONAL_SERVICE = "Error deleting optional service \"{0}\" from offering \"{1}\" and plan \"{2}\": {3}";
     public static final String ERROR_MONITORING_CREATION_OF_SERVICES = "Error monitoring creation of services";
     public static final String ERROR_MAPPING_SERVICE_NAMES_TO_GUIDS = "Error mapping services \"{0}\" to guids";
-    
 
     // WARN log messages
     public static final String CANNOT_RETRIEVE_SERVICE_INSTANCE_OF_OPTIONAL_SERVICE = "Cannot retrieve service instance of optional service {0}";
@@ -171,6 +170,7 @@ public class Messages {
     public static final String ERROR_DURING_CLEAN_UP_0 = "Error during clean-up: {0}";
     public static final String COULD_NOT_DELETE_HISTORIC_PROCESS_0 = "Could not delete historic process \"{0}\"";
     public static final String COULD_NOT_ABORT_OPERATION_0 = "Could not abort operation \"{0}\"";
+    public static final String SKIP_SERVICES_DELETION = "Skipping deletion of services, because the command line option \"--delete-services\" is not specified.";
 
     // INFO log messages
     public static final String MTA_NOT_FOUND = "An MTA with id \"{0}\" does not exist";
@@ -466,5 +466,6 @@ public class Messages {
     public static final String REMOVING_EXPIRED_TOKENS_FROM_TOKEN_STORE = "Removing expired tokens from the token store...";
     public static final String DELETING_DATA_FOR_NON_EXISTING_USERS = "Deleting data for no-longer existing users...";
     public static final String REGISTERED_CLEANERS_IN_CLEAN_UP_JOB_0 = "Registered cleaners in clean-up job: {0}";
+    public static final String MISSING_SERVICES_TO_DELETE = "Missing services to delete.";
 
 }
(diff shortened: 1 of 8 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -52,6 +53,14 @@ protected StepPhase executeAsyncStep(ExecutionWrapper execution) throws Exceptio
...
 
+            boolean deleteServicesFlag = (boolean) execution.getContext()
+                .getVariable(Constants.PARAM_DELETE_SERVICES);
+            
+            if(!deleteServicesFlag) {
+                getStepLogger().warn(Messages.DETECTED_LEFTOVER_SERVICES);
+                return StepPhase.DONE;
+            }
+            
             List<String> servicesToDelete = StepsUtil.getServicesToDelete(execution.getContext());
```
boyan-velinov: First check if servicesToDelete is not empty, then:
-if yes - log debug message that there are not services to delete and finish the step
-if not - check the parameter PARAM_DELETE_SERVICES
-- if not - log warn and finish the step
-- if yes -  continue with actual deletion

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/message/Messages.java
```diff
@@ -466,5 +466,6 @@
     public static final String REMOVING_EXPIRED_TOKENS_FROM_TOKEN_STORE = "Removing expired tokens from the token store...";
     public static final String DELETING_DATA_FOR_NON_EXISTING_USERS = "Deleting data for no-longer existing users...";
     public static final String REGISTERED_CLEANERS_IN_CLEAN_UP_JOB_0 = "Registered cleaners in clean-up job: {0}";
+    public static final String MISSING_SERVICES_TO_DELETE = "Missing services to delete.";
```
boyan-velinov: Will it be displayed in console?

I think we should refrain from printing such message.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/message/Messages.java
```diff
@@ -171,6 +170,7 @@
     public static final String ERROR_DURING_CLEAN_UP_0 = "Error during clean-up: {0}";
     public static final String COULD_NOT_DELETE_HISTORIC_PROCESS_0 = "Could not delete historic process \"{0}\"";
     public static final String COULD_NOT_ABORT_OPERATION_0 = "Could not abort operation \"{0}\"";
+    public static final String SKIP_SERVICES_DELETION = "Option '--delete-services' is not used. Skip deletion of services: \"{0}\".";
```
nictas: Use "Skipping deletion of services, because the command line option \"--delete-services\" is not specified.".

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -1081,4 +1081,10 @@ public static void setServicesGuids(DelegateExecution context, Map<String, Strin
         }.getType());
     }
 
+    public static boolean getDeleteServicesFlag(ExecutionWrapper execution) {
```
nictas: By convention names of boolean methods should be questions, so rename this method to "shouldDeleteServices".

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -54,6 +54,17 @@ protected StepPhase executeAsyncStep(ExecutionWrapper execution) throws Exceptio
 
             List<String> servicesToDelete = StepsUtil.getServicesToDelete(execution.getContext());
 
+            if (servicesToDelete.isEmpty()) {
+                getStepLogger().debug(Messages.MISSING_SERVICES_TO_DELETE);
+                return StepPhase.DONE;
+            }
+
+            if (!StepsUtil.getDeleteServicesFlag(execution)) {
+                getStepLogger().warn(Messages.SKIP_SERVICES_DELETION, String.join("\", \"", servicesToDelete));
```
nictas: Final decision: Do not log the services names as part of the message, because the message may become too long as a result. We considered logging them only on "debug", so that the user will not see the long message, but even that is unnecessary because they are already logged in BuildCloudUndeployModelStep.

Sorry for all the back and forth discussions and changing requirements.

PR #330, opened 2018-11-13
Adding new constant for tracking polling retries.

petko-boyadzhiev:
Adding new constant to track polling count for task list build.


petko-boyadzhiev:
Unnecessary

PR #337, opened 2018-11-20
Enhance FlowableFacade

enchobelezirev:
#### Description: 
This commit enhances the way of getting the dead letter jobs for a process based on its processId or correlationId.
Also, adds possibility to check if the process is running and only then to abort it.

### DIFF
--- a/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/FlowableFacade.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/FlowableFacade.java
@@ -57,10 +57,24 @@ public class FlowableFacade {
     }
 
     public State getProcessInstanceState(String processInstanceId) {
-        if (isProcessInstanceDeleted(processInstanceId)) {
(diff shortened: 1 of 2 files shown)


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/FlowableFacade.java
```diff
@@ -101,18 +101,19 @@ private boolean hasHistoricProcessInstanceNotEnded(String processId) {
...
-        ProcessInstance processInstance = processEngine.getRuntimeService()
-            .createProcessInstanceQuery()
-            .processInstanceId(processId)
-            .singleResult();
+        ProcessInstance processInstance = getProcessInstance(processId);
         return processInstance != null && !processInstance.isEnded();
     }
 
     private boolean isProcessInstanceDeleted(String processId) {
+        return getProcessInstance(processId) == null && hasDeleteReason(processId);
```
nictas: This check for whether the process instance is null or not is repeated in a lot of places. Please move it to the beginning of the getProcessInstanceState() method and remove it from the rest of the private methods.

PR #339, opened 2018-11-20
Use app name for idle route hostname generation

ddonchev:
During bg-deploy app idle routes are generated with values including the
org & space name and the module name.
Collision occurs when two mtas having matching module names but having
different app_names configured for those modules, get deployed.
By chaning the app-name to be taken in the evaluation, instead of the module name,
 the collision will be avoided.

JIRA:LMCROSSITXSADEPLOY-1321


ddonchev:
A case has to be taken care of - this change would break existing applications using the module name in their routes.

A fix could be to dynamically compute the default host based on the existing hosts of app.

nictas:
The nature of the task has changed. This PR is therefore not relevant anymore.

PR #340, opened 2018-11-21
Temporary disable trigger for Postgre LO related to process_log

boyan-velinov:


theghost5800:
This change was included in this PR https://github.com/cloudfoundry-incubator/multiapps-controller/pull/345

PR #341, opened 2018-11-21
Prepare services to be deleted

IvankaTashkova:
#### Description: 
Unbind service bindings and delete service keys in order to allow deletion of the services.


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -1080,6 +1082,17 @@ public static void setServicesGuids(DelegateExecution context, Map<String, Strin
         return JsonUtil.fromJson(jsonString, new TypeToken<Map<String, String>>() {
         }.getType());
     }
+    
+    public static CloudApplication getBoundApplication(List<CloudApplication> applications, UUID appGuid) {
+        return applications.stream()
+            .filter(app -> {
+                return app.getMeta()
```
nictas: Extract the predicate in the filter to a separate method (hasGuid or something of the sort).

PR #347, opened 2018-11-28
Delete services

IvankaTashkova:


com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStepTest.java
```diff
@@ -173,7 +176,12 @@ private void prepareContext() {
     }
 
     private void prepareClient() {
+        Mockito.when(dummyUserProvidedService.isUserProvided())
+            .thenReturn(false);
```
VaskoBozhurski: why is isUserProvided returning false for a cloud service called dummyUserProvidedService? If the service is not really user provided maybe it needs a new name?

PR #348, opened 2018-11-28
Fix computing of modulesToUndeploy

IvankaTashkova:


valentinEmpy:
Won't this change mean that in bg-deploy the old version of the application won't ever be deleted?
Please test this.
This will indeed get the correct modules which have to be kept (which refer to the new application), but won't it mess up which modules have to be undeployed?
I guess the naming is confusing here and causes misunderstandings. This will make more sense if we think in the context of applications instead of modules. But I assume that if there is no change in the MTA in bg-deploy, all old versions (-blue) modules should be undeployed, and all new versions (-green) should be kept.
Let's sync tomorrow in person. :)

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/BuildCloudUndeployModelStep.java
```diff
@@ -81,49 +81,71 @@ protected StepPhase executeStep(ExecutionWrapper execution) {
...
             .collect(Collectors.toList());
     }
 
+    private boolean shouldDeleteService(List<DeployedMtaModule> modulesToKeep, String service,
+        List<CloudApplicationExtended> appsToDeploy) {
+        return modulesToKeep.stream()
+            .noneMatch(module -> module.getServices()
+                .contains(service))
+                && appsToDeploy.stream()
+                    .map(CloudApplicationExtended::getServices)
```
valentinEmpy: Either use map on both expressions or not at all.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/BuildCloudUndeployModelStep.java
```diff
@@ -81,49 +81,71 @@ protected StepPhase executeStep(ExecutionWrapper execution) {
...
             .collect(Collectors.toList());
     }
 
+    private boolean shouldUndeployModule(DeployedMtaModule deployedModule, Set<String> mtaModules, List<String> appsToDeploy) {
+        if (!mtaModules.contains(deployedModule.getModuleName())) {
+            return true;
+        }
+        // The deployed module may be in the list of MTA modules, but the actual application that was created from it may have a
+        // different name:
+        boolean appNameMatchesExisting = appsToDeploy.contains(deployedModule.getAppName());
```
valentinEmpy: Directly return instead of doing an assignment and then using a "logical not" operator.
Delete the empty line on the bottom of the method.

PR #349, opened 2018-11-28
Fix deleteServiceKeys for user-provided services

IvankaTashkova:


com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStepTest.java
```diff
@@ -173,7 +176,12 @@ private void prepareContext() {
     }
 
     private void prepareClient() {
+        Mockito.when(dummyUserProvidedService.isUserProvided())
```
VaskoBozhurski: if dummyUserProvidedService is not a user provided service maybe it needs a better name?

PR #375, opened 2018-12-24
One-off tasks support for CF

nictas:
JIRA: LMCROSSITXSADEPLOY-292


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/message/Messages.java
```diff
@@ -69,7 +69,7 @@
     public static final String INVALID_TCP_ROUTE = "TCP route can either be TCP or TCPS";
     public static final String NO_SERVICE_PLAN_FOUND = "Could not create service instance \"{0}\". Service plan \"{1}\" from service offering \"{2}\" was not found.";
     public static final String NO_CONTENT_TO_DEPLOY = "No content to deploy";
-    public static final String UNABLE_TO_PARSE_PARAMETER = "Unable to parse parameter {0} with value {1}";
+    public static final String UNABLE_TO_PARSE_PARAMETER = "Unable to parse memory string \"{0}\"";
```
ddonchev: .

PR #377, opened 2019-01-10
Separate stage and start of applications on CF

IvanBorislavovDimitrov:

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
@@ -155,6 +155,7 @@ public class Constants {
     public static final String VAR_STEP_START_TIME = "stepStartTime_";
     public static final String VAR_DEPLOYED_MTA_COLOR = "deployedMtaColor";
     public static final String VAR_MTA_COLOR = "mtaColor";
+    public static final String VAR_BUILD_GUID = "buildGuid";
 
     public static final String TOOL_TYPE = "tool_type";
     public static final String FEEDBACK_MAIL = "feedback_form";
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
@@ -2,24 +2,27 @@ package com.sap.cloud.lm.sl.cf.process.steps;
 
 import static java.text.MessageFormat.format;
 
+import java.util.UUID;
+
 import org.cloudfoundry.client.lib.CloudControllerClient;
 import org.cloudfoundry.client.lib.CloudControllerException;
 import org.cloudfoundry.client.lib.CloudOperationException;
-import org.cloudfoundry.client.lib.StartingInfo;
 import org.cloudfoundry.client.lib.domain.CloudApplication;
 import org.cloudfoundry.client.lib.domain.PackageState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.http.HttpStatus;
-
-import com.sap.cloud.lm.sl.cf.client.XsCloudControllerClient;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
 import com.sap.cloud.lm.sl.cf.core.cf.clients.RecentLogsRetriever;
+import com.sap.cloud.lm.sl.cf.core.util.ApplicationConfiguration;
 import com.sap.cloud.lm.sl.cf.persistence.services.ProcessLoggerProvider;
 import com.sap.cloud.lm.sl.cf.process.Constants;
 import com.sap.cloud.lm.sl.cf.process.message.Messages;
+import com.sap.cloud.lm.sl.cf.process.util.ApplicationStager;
+import com.sap.cloud.lm.sl.cf.process.util.StagingState;
+import com.sap.cloud.lm.sl.cf.process.util.StagingState.StagingLogs;
 import com.sap.cloud.lm.sl.cf.process.util.XMLValueFilter;
 import com.sap.cloud.lm.sl.common.SLException;
-import com.sap.cloud.lm.sl.common.util.Pair;
 
 public class PollStageAppStatusExecution implements AsyncExecution {
 
@@ -27,8 +30,11 @@ public class PollStageAppStatusExecution implements AsyncExecution {
 
     private RecentLogsRetriever recentLogsRetriever;
 
-    public PollStageAppStatusExecution(RecentLogsRetriever recentLogsRetriever) {
+    private ApplicationStager applicationStager;
+    
+    public PollStageAppStatusExecution(RecentLogsRetriever recentLogsRetriever, ApplicationStager applicationStager) {
         this.recentLogsRetriever = recentLogsRetriever;
+        this.applicationStager = applicationStager;
     }
 
     @Override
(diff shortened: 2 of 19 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/StagingResolver.java
```diff
@@ -0,0 +1,25 @@
+package com.sap.cloud.lm.sl.cf.process.util;
+
+import org.cloudfoundry.client.lib.CloudControllerClient;
+import org.cloudfoundry.client.lib.StartingInfo;
+import org.cloudfoundry.client.lib.domain.PackageState;
+
+import com.sap.cloud.lm.sl.cf.process.steps.ExecutionWrapper;
+import com.sap.cloud.lm.sl.common.util.Pair;
+
+public abstract class StagingResolver {
```
nictas: You don't need this abstract class. Move the logic to CFStagingResolver, rename it to StagingResolver and extend it for XSA.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/StagingResolverCreator.java
```diff
@@ -0,0 +1,21 @@
+package com.sap.cloud.lm.sl.cf.process.util;
+
+import org.cloudfoundry.client.lib.CloudControllerClient;
+
+public enum StagingResolverCreator {
```
nictas: Use Spring instead of this factory. See application-context.xml for an example.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/XSStagingResolver.java
```diff
@@ -0,0 +1,64 @@
...
+import org.cloudfoundry.client.lib.StartingInfo;
+import org.cloudfoundry.client.lib.domain.PackageState;
+import org.springframework.http.HttpStatus;
+
+import com.sap.cloud.lm.sl.cf.process.Constants;
+import com.sap.cloud.lm.sl.cf.process.steps.ExecutionWrapper;
+import com.sap.cloud.lm.sl.cf.process.steps.StepsUtil;
+import com.sap.cloud.lm.sl.common.util.Pair;
+
+public class XSStagingResolver extends StagingResolver {
```
nictas: Move this to our internal project.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/XSStagingResolver.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public XSStagingResolver(CloudControllerClient client) {
+        super(client);
+    }
+
+    @Override
+    public Pair<PackageState, String> getStagingState(ExecutionWrapper execution) {
```
nictas: This is not a clean interface. getStagingState() returns the staging state, but it also logs what happened during the staging. Separate this in two methods: getStagingState() and getStagingLogs().

nictas: Don't use Pair. Create a new class that holds the state and an error message.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
```diff
@@ -1,121 +1,95 @@
...
+            }
+
+            execution.getStepLogger()
+                .info(Messages.APP_STAGED, app.getName());
+
+            String appId = client.getApplication(app.getName())
+                .getMeta()
+                .getGuid()
+                .toString();
+            stagingResolver.finilizeStep(execution, appId);
```
nictas: Use a better name than "finalizeStep".

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PrepareAppsRestartStep.java
```diff
@@ -1,33 +1,33 @@
-package com.sap.cloud.lm.sl.cf.process.steps;
```
nictas: Check whether you've set autocrlf in your Git configuration (https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration).

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StageAppStep.java
```diff
@@ -1,68 +1,94 @@
...
+        }
+    }
+
+    private StepPhase stageAppCF(ExecutionWrapper execution, CloudControllerClient client, CloudApplication app) {
+        String uploadToken = (String) execution.getContext()
+            .getVariable(Constants.VAR_UPLOAD_TOKEN);
+        if (uploadToken == null) {
+            return StepPhase.DONE;
+        }
+        String packageGuid = uploadToken.substring(uploadToken.lastIndexOf("/") + 1);
```
nictas: Naaah, this is not a good idea. Make a more structured UploadToken if necessary, but don't use substring(lastIndexOf()).

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/CFStagingResolver.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public Pair<PackageState, String> getStagingState(ExecutionWrapper execution) {
+        String buildGuid = (String) execution.getContext().getVariable(Constants.PARAM_BUILD_GUID);
+        if (buildGuid == null) {
+            return new Pair<>(PackageState.STAGED, null);
+        }
+        CloudBuild build = getClient().getBuild(buildGuid);
+        PackageState packageState = null;
+        String stagingError = null;
+        switch (build.getState()) {
```
nictas: Extract this switch in a separate method.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
```diff
@@ -37,18 +43,32 @@ public AsyncExecutionState execute(ExecutionWrapper execution) {
         CloudControllerClient client = execution.getControllerClient();
 
         try {
+            stagingResolver.setClient(client);
```
nictas: No. If you're going to change the state of a Spring component, then the component shouldn't be a component at all (there are some exceptions for prototype components though). Modifying a singleton that is used in multiple threads is **extremely** dangerous. What if two processes start using the same object at the same time. The second would override the client of the first and then the StagingResolver will work with the second client for both threads - thus giving one user control over the other user's space.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StageAppStep.java
```diff
@@ -45,6 +63,24 @@ protected StepPhase executeAsyncStep(ExecutionWrapper execution) {
         }
     }
 
+    private StepPhase stageAppCF(ExecutionWrapper execution, CloudControllerClient client, CloudApplication app) {
```
nictas: Why is the logic different for CF and XSA? Isn't that the whole purpose of the `StagingResolver`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
```diff
@@ -79,17 +79,17 @@ public StepPhase executeAsyncStep(ExecutionWrapper execution) throws FileStorage
...
+        UploadToken uploadToken = new UploadToken();
         final DelegateExecution context = execution.getContext();
         FileDownloadProcessor uploadFileToControllerProcessor = new DefaultFileDownloadProcessor(StepsUtil.getSpaceId(context),
             appArchiveId, appArchiveStream -> {
                 Path filePath = null;
                 long maxSize = configuration.getMaxResourceFileSize();
                 try {
                     filePath = extractFromMtar(appArchiveStream, fileName, maxSize);
-                    upload(execution, client, app, filePath, uploadTokenBuilder);
+                    upload(execution, client, app, filePath, uploadToken);
```
nictas: Why doesn't the `upload` method return the upload token?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/StagingResolver.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.UUID;
+
+import org.cloudfoundry.client.lib.CloudControllerClient;
+import org.cloudfoundry.client.lib.domain.CloudBuild;
+import org.cloudfoundry.client.lib.domain.PackageState;
+
+import com.sap.cloud.lm.sl.cf.process.Constants;
+import com.sap.cloud.lm.sl.cf.process.steps.ExecutionWrapper;
+
+public class StagingResolver {
```
nictas: Now that I think about it, `Resolver` does not express the purpose of the class very clearly. Consider using something like `ApplicationStager`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
```diff
@@ -49,6 +49,7 @@
     public static final String PARAM_SKIP_OWNERSHIP_VALIDATION = "skipOwnershipValidation";
     public static final String PARAM_MODULES_FOR_DEPLOYMENT = "modulesForDeployment";
     public static final String PARAM_RESOURCES_FOR_DEPLOYMENT = "resourcesForDeployment";
+    public static final String PARAM_BUILD_GUID = "buildGuid";
```
nictas: Rename this to `VAR_BUILD_GUID`. The `PARAM` prefix is only used for variables, which are given to the process during its startup.

PR #395, opened 2019-01-23
Fix possible null pointer

petko-boyadzhiev:


nictas:
Fix is applied in https://github.com/cloudfoundry-incubator/multiapps-controller/pull/409.

PR #414, opened 2019-02-06
Content deployment logs

IvankaTashkova:


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -76,9 +76,9 @@
 
 public class StepsUtil {
 
-    private static org.apache.log4j.Logger getAppLogger(DelegateExecution context, String appName,
+    protected static org.apache.log4j.Logger getCustomLogger(DelegateExecution context, String customName,
```
boyan-velinov: You can keep it more simple with:
getCustomLogger -> getLogger
customName -> name

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -905,7 +904,7 @@ public static void setStepPhase(DelegateExecution context, StepPhase stepPhase)
         return (T) context.getVariable(name);
     }
 
-    private static String getPrefix(Logger logger) {
+    public static String getPrefix(Logger logger) {
```
boyan-velinov: rename it to getLoggerPrefix

PR #418, opened 2019-02-12
Fix setting of service parameters from file for service create/update

petko-boyadzhiev:


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineServiceCreateUpdateServiceActionsStep.java
```diff
@@ -77,14 +77,25 @@ protected StepPhase executeStep(ExecutionWrapper execution) throws Exception {
 
         List<ServiceAction> actions = determineActions(controllerClient, spaceId, serviceToProcess, existingService, serviceDefaultTags,
             serviceKeys, execution);
+        
+        setServiceParameters(serviceToProcess, actions, execution.getContext());
+        
         StepsUtil.setServiceActionsToExecute(actions, execution.getContext());
         StepsUtil.isServiceUpdated(false, execution.getContext());
```
enchobelezirev: Swap places of false and execution.getContext()

PR #420, opened 2019-02-13
Retrieve recent app logs in fail safe manner

boyan-velinov:

### DIFF
--- a/com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/ExecutionRetrier.java
+++ b/com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/ExecutionRetrier.java
@@ -18,9 +18,15 @@ public class ExecutionRetrier {
     private static final long DEFAULT_RETRY_COUNT = 3;
     private static final long DEFAULT_WAIT_TIME_BETWEEN_RETRIES_IN_MILLIS = 5000;
 
+    private boolean failSafe;
     private long retryCount = DEFAULT_RETRY_COUNT;
     private long waitTimeBetweenRetriesInMillis = DEFAULT_WAIT_TIME_BETWEEN_RETRIES_IN_MILLIS;
-
+    
+    public ExecutionRetrier failSafe() {
+        this.failSafe = true;
+        return this;
+    }
+    
     public ExecutionRetrier withRetryCount(long retryCount) {
         this.retryCount = retryCount;
         return this;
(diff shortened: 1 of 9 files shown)


com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/ExecutionRetrier.java
```diff
@@ -18,9 +18,15 @@
     private static final long DEFAULT_RETRY_COUNT = 3;
     private static final long DEFAULT_WAIT_TIME_BETWEEN_RETRIES_IN_MILLIS = 5000;
 
+    private boolean failSafe;
     private long retryCount = DEFAULT_RETRY_COUNT;
     private long waitTimeBetweenRetriesInMillis = DEFAULT_WAIT_TIME_BETWEEN_RETRIES_IN_MILLIS;
-
+    
+    public ExecutionRetrier withFailSafe(boolean failSafe) {
```
nictas: I would've preferred it if the method was called failSafe() without any arguments and just assigned true to the field. It's default value is false anyway.

com.sap.cloud.lm.sl.cf.client/src/main/java/com/sap/cloud/lm/sl/cf/client/util/ExecutionRetrier.java
```diff
@@ -47,7 +53,14 @@ public ExecutionRetrier withWaitTimeBetweenRetriesInMillis(long waitTimeBetweenR
             }
             sleep(waitTimeBetweenRetriesInMillis);
         }
-        return supplier.get();
+        try {
+            return supplier.get();
+        } catch (Exception e) {
```
nictas: This is a bit inconsistent. The try-catch block above retries on ResourceAccessException and CloudOperationException, but this ignores every type of exception. What if the 2nd retry throws an IOException? Why would the last retry be resistant to IOExceptions, but the second wouldn't?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/clients/CustomControllerClientErrorHandler.java
```diff
@@ -28,6 +28,16 @@ public CustomControllerClientErrorHandler(ExecutionRetrier retrier) {
         this.retrier = retrier;
     }
 
+    public <T> T handleErrorsOrReturnResult(Supplier<T> supplier, boolean failSafe, HttpStatus... httpStatusesToIgnore) {
```
nictas: Where is the failSafe argument used here?

PR #425, opened 2019-02-15
Remove taskExecutionId from ProgressMessage

enchobelezirev:
#### Description: 
Currently, the task exection id was used when the tasklist building of polling tasks was required(StartAppTask, etc.).
With the old process engine(Activiti) when some step has failed, this executionId is removed at the step startup.
With flowable, that is not possible as when some step is being executed, there is no access to the jobs of the step from the step itself. Thus, when some big Exception is thrown which kills the JVM(like OOM), the executionId is not cleared from the database and kconsequent errors are produced.

Change-Id: Ibf90eb43e9f61e54c3720d9f43b0c23883d4999c


com.sap.cloud.lm.sl.cf.persistence/src/main/resources/com/sap/cloud/lm/sl/cf/persistence/db/changelog/db-changelog-alter_progress_message_drop_task_execution_id_column.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    logicalFilePath="com/sap/cloud/lm/sl/persistence/db/changelog/db-changelog-alter_progress_message_add_task_execution_id_column.xml"
```
nictas: https://img.memecdn.com/copypasta_o_7243351.jpg
https://pics.me.me/delet-dis-34029456.png

com.sap.cloud.lm.sl.cf.persistence/src/main/resources/com/sap/cloud/lm/sl/cf/persistence/db/changelog/db-changelog-alter_progress_message_drop_task_execution_id_column.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nictas: Use the Eclipse formatter with the following settings:
Line width: 120
Indent using spaces
Indentation size: 4

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ProcessStepHelper.java
```diff
@@ -57,14 +53,13 @@ protected void postExecuteStep(DelegateExecution context, StepPhase state) {
     }
 
     void preExecuteStep(DelegateExecution context, StepPhase initialPhase) {
-        init(context, initialPhase);
+        String taskId = taskIdProvider.getTaskId(context);// context.getCurrentActivityId();
```
nictas: Why do we need the comment?

PR #428, opened 2019-02-21
Fix retry of polling steps

enchobelezirev:
#### Description: 
When a step which has a timeout configured, hits the timeout period and is not re-triggered in that time frame, the step fails with the following error: **Step XXX has timed out**

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/TimeoutAsyncFlowableStep.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/TimeoutAsyncFlowableStep.java
@@ -27,11 +27,16 @@ public abstract class TimeoutAsyncFlowableStep extends AsyncFlowableStep {
 
     private long getStepStartTime(DelegateExecution context) {
         Long stepStartTime = (Long) context.getVariable(getStepStartTimeVariable());
-        if (stepStartTime == null) {
+        if (stepStartTime == null || isInRetry(context)) {
             stepStartTime = System.currentTimeMillis();
             context.setVariable(getStepStartTimeVariable(), stepStartTime);
         }
+
         return stepStartTime;
+    }
+
+    private boolean isInRetry(DelegateExecution context) {
+        return StepsUtil.getStepPhase(context) == StepPhase.RETRY;
     }
 
     private String getStepStartTimeVariable() {
--- a/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/FlowableFacade.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/FlowableFacade.java
@@ -333,17 +333,25 @@ public class FlowableFacade {
     public void shutdownJobExecutor(long secondsToWaitOnShutdown) {
         setSecondsToWaitOnJobExecutorShutdown(secondsToWaitOnShutdown);
         LOGGER.info(Messages.SHUTTING_DOWN_FLOWABLE_JOB_EXECUTOR);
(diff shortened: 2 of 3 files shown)


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/SetRetryPhaseAdditionalProcessAction.java
```diff
@@ -0,0 +1,35 @@
...
+
+import javax.inject.Inject;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class SetRetryPhaseAdditionalProcessAction implements AdditionalProcessAction {
+
+    private static final String RETRY_STEP_PHASE = "RETRY";
+    private static final String STEP_PHASE = "stepPhase";
```
nictas: Here we define the same constant for the second time. This can be a problem in the future.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/TimeoutAsyncFlowableStep.java
```diff
@@ -28,9 +28,19 @@ private boolean hasTimedOut(DelegateExecution context) {
...
         }
+
+        if (StepsUtil.getStepPhase(context) == StepPhase.RETRY) {
+            return retrieveAndSetStartTime(context);
+        }
+
+        return stepStartTime;
+    }
+
+    private Long retrieveAndSetStartTime(DelegateExecution context) {
```
nictas: Generally, methods with "and" in their name should be refactored, because they do more than one thing (SRP).

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/TimeoutAsyncFlowableStep.java
```diff
@@ -28,9 +28,19 @@ private boolean hasTimedOut(DelegateExecution context) {
     private long getStepStartTime(DelegateExecution context) {
         Long stepStartTime = (Long) context.getVariable(getStepStartTimeVariable());
         if (stepStartTime == null) {
-            stepStartTime = System.currentTimeMillis();
-            context.setVariable(getStepStartTimeVariable(), stepStartTime);
+            return retrieveAndSetStartTime(context);
         }
+
+        if (StepsUtil.getStepPhase(context) == StepPhase.RETRY) {
```
nictas: Merge this if statement with the one above. Then, inline the retrieveAndSetStartTime() method in the body of the statement.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/SetRetryPhaseAdditionalProcessAction.java
enchobelezirev: I had doubts in myself about this and moving the constant in the cf.core project did not seem to be a good idea for me?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/TimeoutAsyncFlowableStep.java
enchobelezirev: Thought for this... the code is in this style due to more readability - will check how it will look like inlined

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/TimeoutAsyncFlowableStep.java
enchobelezirev: Correct. Will rename it

PR #432, opened 2019-02-25
Sonar fixes

enchobelezirev:
#### Description: 
Sonar fixes

### DIFF
--- a/com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/executors/SqlQueryExecutor.java
+++ b/com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/executors/SqlQueryExecutor.java
@@ -17,8 +17,9 @@ public class SqlQueryExecutor {
     }
 
     public <R> R executeWithAutoCommit(SqlQuery<R> sqlQuery) throws SQLException {
-        Connection connection = dataSource.getConnection();
+        Connection connection = null;
         try {
+            connection = dataSource.getConnection();
             connection.setAutoCommit(true);
             return sqlQuery.execute(connection);
         } catch (SQLException e) {
@@ -30,8 +31,9 @@ public class SqlQueryExecutor {
     }
 
     public <R> R execute(SqlQuery<R> sqlQuery) throws SQLException {
-        Connection connection = dataSource.getConnection();
+        Connection connection = null;
         try {
+            connection = dataSource.getConnection();
             connection.setAutoCommit(false);
             R result = sqlQuery.execute(connection);
             JdbcUtil.commit(connection);
@@ -40,8 +42,14 @@ public class SqlQueryExecutor {
             JdbcUtil.rollback(connection);
             throw e;
         } finally {
-            connection.setAutoCommit(true);
+            setAutocommitSafely(connection);
             JdbcUtil.closeQuietly(connection);
+        }
+    }
+
+    private void setAutocommitSafely(Connection connection) throws SQLException {
+        if (connection != null) {
+            connection.setAutoCommit(true);
         }
     }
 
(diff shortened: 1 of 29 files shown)


com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/executors/SqlQueryExecutor.java
```diff
@@ -30,8 +31,9 @@ public SqlQueryExecutor(DataSource dataSource) {
     }
 
     public <R> R execute(SqlQuery<R> sqlQuery) throws SQLException {
-        Connection connection = dataSource.getConnection();
+        Connection connection = null;
         try {
+            connection = dataSource.getConnection();
```
nictas: Why do we need to initialize the connection in the try block?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/executors/SqlQueryExecutor.java
```diff
@@ -17,8 +17,9 @@ public SqlQueryExecutor(DataSource dataSource) {
     }
 
     public <R> R executeWithAutoCommit(SqlQuery<R> sqlQuery) throws SQLException {
-        Connection connection = dataSource.getConnection();
+        Connection connection = null;
         try {
+            connection = dataSource.getConnection();
```
nictas: Why do we need to initialize the connection in the try block?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -129,6 +129,10 @@ private void unbindService(CloudControllerClient client, String serviceName) {
         logBindings(bindings);
         for (CloudServiceBinding binding : bindings) {
             CloudApplication application = StepsUtil.getBoundApplication(client.getApplications(), binding.getAppGuid());
+            if (application == null) {
+                throw new IllegalStateException(
```
nictas: Why not put the exception inside the method by calling `.orElseThrow()` on the Optional there?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/configuration/UAAClientConfiguration.java
```diff
@@ -34,7 +34,7 @@ private URL readTokenEndpoint(URL targetURL) {
         try {
             String infoURL = targetURL.toString() + "/v2/info";
             ResponseEntity<String> infoResponse = new RestTemplate().getForEntity(infoURL, String.class);
-            if (infoResponse == null) {
+            if (infoResponse.getBody() == null) {
```
nictas: Can the response itself be `null`?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/executors/SqlQueryExecutor.java
enchobelezirev: Because it throws SQLException and will not be handled properly?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/executors/SqlQueryExecutor.java
enchobelezirev: Same comment from above

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
enchobelezirev: Also thought of this... the reason I did it this way is because the methods in StepsUtils are not throwing some exceptions(as far as I saw) and decided to do it here?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/configuration/UAAClientConfiguration.java
enchobelezirev: As far as I observed the code of the Spring, the response is always built. However, based on the Response type(in this case String.class), the body is decided whether to be included in the response or not.
You could check ResponseEntityResponseExtractor which is component responsible for getting the response body out of the HTTP response

PR #441, opened 2019-03-01
Store single token per user in cache

theghost5800:
make username column in oauth_acces_token table unique and do update if
new token is provided from the user

JIRA:LMCROSSITXSADEPLOY-1430


theghost5800:
This fix is not required right now.

PR #442, opened 2019-03-05
Reduce log size

ddonchev:
Those two classes generate about 10% of the content in the app log

 Issue: none


ddonchev:
cf oqs passed with no issues.

PR #459, opened 2019-03-19
Fix - tasks not executed if their app exists but not staged

IvanBorislavovDimitrov:
1. We attempt to deploy it, but the staging of its corresponding application fails for some reason.
2. We run another deployment after we've fixed the cause of the staging failure.
3. The application will be restaged due to failed staging.

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStep.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStep.java
@@ -8,6 +8,7 @@ import org.cloudfoundry.client.lib.CloudControllerException;
 import org.cloudfoundry.client.lib.CloudOperationException;
 import org.cloudfoundry.client.lib.domain.CloudApplication;
 import org.flowable.engine.delegate.DelegateExecution;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
@@ -20,8 +21,11 @@ import com.sap.cloud.lm.sl.cf.core.cf.apps.ApplicationStartupStateCalculator;
 import com.sap.cloud.lm.sl.cf.core.cf.apps.ApplicationStateAction;
 import com.sap.cloud.lm.sl.cf.core.cf.apps.ChangedApplicationActionCalcultor;
 import com.sap.cloud.lm.sl.cf.core.cf.apps.UnchangedApplicationActionCalculator;
+import com.sap.cloud.lm.sl.cf.core.util.ApplicationConfiguration;
 import com.sap.cloud.lm.sl.cf.process.Constants;
 import com.sap.cloud.lm.sl.cf.process.message.Messages;
+import com.sap.cloud.lm.sl.cf.process.util.ApplicationStager;
+import com.sap.cloud.lm.sl.cf.process.util.ApplicationStagerFactory;
 import com.sap.cloud.lm.sl.common.SLException;
 
 @Component("determineDesiredStateAchievingActionsStep")
@@ -30,9 +34,13 @@ public class DetermineDesiredStateAchievingActionsStep extends SyncFlowableStep 
 
     protected Supplier<ApplicationStartupStateCalculator> appStateCalculatorSupplier = ApplicationStartupStateCalculator::new;
 
+    @Autowired
+    protected ApplicationConfiguration configuration;
+
     @Override
     protected StepPhase executeStep(ExecutionWrapper context) {
         CloudApplication app = StepsUtil.getApp(context.getContext());
+
         try {
             return attemptToExecuteStep(context);
         } catch (CloudOperationException coe) {
@@ -46,24 +54,26 @@ public class DetermineDesiredStateAchievingActionsStep extends SyncFlowableStep 
     }
 
     private StepPhase attemptToExecuteStep(ExecutionWrapper execution) {
-        CloudApplication app = StepsUtil.getApp(execution.getContext());
-        ApplicationStartupState currentState = computeCurrentState(execution, app);
-        getStepLogger().debug(Messages.CURRENT_STATE, app.getName(), currentState);
+        String appName = StepsUtil.getApp(execution.getContext())
+            .getName();
+        CloudControllerClient client = execution.getControllerClient();
+        CloudApplication app = client.getApplication(appName);
+        ApplicationStartupState currentState = computeCurrentState(app);
+        getStepLogger().debug(Messages.CURRENT_STATE, appName, currentState);
         ApplicationStartupState desiredState = computeDesiredState(execution.getContext(), app);
-        getStepLogger().debug(Messages.DESIRED_STATE, app.getName(), desiredState);
-
+        getStepLogger().debug(Messages.DESIRED_STATE, appName, desiredState);
+        ApplicationStager applicationStager = ApplicationStagerFactory.createApplicationStager(configuration.getPlatformType());
         Set<ApplicationStateAction> actionsToExecute = getActionsCalculator(execution.getContext()).determineActionsToExecute(currentState,
-            desiredState);
-        getStepLogger().debug(Messages.ACTIONS_TO_EXECUTE, app.getName(), actionsToExecute);
+            desiredState, applicationStager.isApplicationStagedCorrectly(execution, app));
+        getStepLogger().debug(Messages.ACTIONS_TO_EXECUTE, appName, actionsToExecute);
 
         StepsUtil.setAppStateActionsToExecute(execution.getContext(), actionsToExecute);
         return StepPhase.DONE;
     }
 
-    private ApplicationStartupState computeCurrentState(ExecutionWrapper execution, CloudApplication app) {
-        CloudControllerClient client = execution.getControllerClient();
+    private ApplicationStartupState computeCurrentState(CloudApplication app) {
         return appStateCalculatorSupplier.get()
-            .computeCurrentState(client.getApplication(app.getName()));
+            .computeCurrentState(app);
     }
 
     private ApplicationStartupState computeDesiredState(DelegateExecution context, CloudApplication app) {
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
@@ -1,5 +1,6 @@
 package com.sap.cloud.lm.sl.cf.process.util;
 
+import java.util.Comparator;
 import java.util.UUID;
 
 import org.cloudfoundry.client.lib.CloudControllerClient;
@@ -7,6 +8,7 @@ import org.cloudfoundry.client.lib.domain.CloudApplication;
 import org.cloudfoundry.client.lib.domain.CloudBuild;
 import org.cloudfoundry.client.lib.domain.PackageState;
 import org.cloudfoundry.client.lib.domain.UploadToken;
+import org.cloudfoundry.client.lib.util.JsonUtil;
 import org.flowable.engine.delegate.DelegateExecution;
 
 import com.sap.cloud.lm.sl.cf.process.Constants;
@@ -16,9 +18,10 @@ import com.sap.cloud.lm.sl.cf.process.steps.StepPhase;
 import com.sap.cloud.lm.sl.cf.process.steps.StepsUtil;
 
 public class ApplicationStager {
-    
+
     public StagingState getStagingState(ExecutionWrapper execution, CloudControllerClient client) {
-        UUID buildGuid = (UUID) execution.getContext().getVariable(Constants.VAR_BUILD_GUID);
+        UUID buildGuid = (UUID) execution.getContext()
+            .getVariable(Constants.VAR_BUILD_GUID);
         if (buildGuid == null) {
             return new StagingState(PackageState.STAGED, null);
         }
@@ -26,7 +29,32 @@ public class ApplicationStager {
 
         return getStagingState(build);
     }
-    
+
+    public boolean isApplicationStagedCorrectly(ExecutionWrapper execution, CloudApplication cloudApplication) {
+        CloudBuild cloudBuild = execution.getControllerClient()
+            .getBuildsForApplication(cloudApplication.getMeta()
+                .getGuid())
+            .stream()
+            .max(Comparator.comparing(build -> build.getMeta()
+                .getCreated()))
+            .orElse(null);
+        if (cloudBuild == null) {
+            execution.getStepLogger()
+                .debug(Messages.NO_BUILD_FOUND_FOR_APPLICATION, cloudApplication.getName());
+            return false;
+        }
+        if (cloudBuild.getState() == CloudBuild.BuildState.STAGED && cloudBuild.getDroplet() != null && cloudBuild.getError() == null) {
+            return true;
+        }
+        execution.getStepLogger()
+            .info(Messages.APPLICATION_NOT_STAGED_CORRECTLY, cloudApplication.getName());
+        execution.getStepLogger()
+            .debug(Messages.LAST_BUILD, cloudBuild.getName());
+        execution.getStepLogger()
+            .debug(JsonUtil.convertToJson(cloudBuild));
+        return false;
+    }
+
     private StagingState getStagingState(CloudBuild build) {
         PackageState packageState = null;
         String stagingError = null;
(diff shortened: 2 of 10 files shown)


com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStepTest.java
```diff
@@ -124,16 +133,25 @@
...
     private RestartParameters appRestartParameters;
 
     @Mock
     private ApplicationStartupStateCalculator appStateCalculator;
 
+    @Mock
+    private CloudControllerClient client;
+
+    @Mock
+    private CloudApplicationExtended cloudApplicationExtended;
```
nictas: Usually, the mocks we create via the `@Mock` annotation are objects with behavior (clients, factories, etc.) and not POJOs. POJOs we typically mock wherever necessary directly with `Mockito.mock()`. That's because objects with behavior are more likely to get reused in multiple methods in the test, while POJOs are usually used in just one place and it's always a good idea to [limit the scope of a variable as much as possible](https://softwareengineering.stackexchange.com/questions/307346/why-is-it-good-programming-practice-to-limit-scope).
Additionally, mocks with `@Mock` are used whenever they are injected into the tested class via `@InjectMocks` (see the parent class of this test).

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStepTest.java
```diff
@@ -154,10 +173,16 @@ public void setUp() {
         context.setVariable(Constants.PARAM_NO_START, false);
         when(appStateCalculator.computeCurrentState(any())).thenReturn(currentAppState);
         when(appStateCalculator.computeDesiredState(any(), eq(false))).thenReturn(desiredAppState);
-        step.appStateCalculatorSupplier = () -> appStateCalculator;
-        CloudApplicationExtended app = new CloudApplicationExtended(null, "dummy");
+        step.appStateCalculatorSupplier = ApplicationStartupStateCalculator::new;
+        CloudApplicationExtended app = new CloudApplicationExtended(null, DUMMY);
         app.setRestartParameters(appRestartParameters);
         context.setVariable(Constants.VAR_APP_TO_DEPLOY, JsonUtil.toJson(app));
+        when(execution.getControllerClient()).thenReturn(client);
```
nictas: This line should not be necessary, if you reuse the client mock from the parent class.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStep.java
```diff
@@ -47,23 +50,33 @@ protected StepPhase executeStep(ExecutionWrapper context) {
 
     private StepPhase attemptToExecuteStep(ExecutionWrapper execution) {
         CloudApplication app = StepsUtil.getApp(execution.getContext());
-        ApplicationStartupState currentState = computeCurrentState(execution, app);
+        CloudControllerClient client = execution.getControllerClient();
+        CloudApplication cloudApplication = client.getApplication(app.getName());
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Renamed. We need to retrieve it because the application returned from StepsUtil.getApp(...) only contains the application name. All of the other fields are null.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
```diff
@@ -16,17 +16,29 @@
...
         }
         CloudBuild build = client.getBuild(buildGuid);
 
         return getStagingState(build);
     }
-    
+
+    public boolean isAplicationStagedCorrectly(CloudControllerClient client, CloudApplication cloudApplication) {
+        if (cloudApplication == null) {
+            return true;
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: It actually can't be -> refactored

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStagerFactory.java
```diff
@@ -18,4 +18,8 @@ public static ApplicationStager createApplicationStager(PlatformType platformTyp
 
         throw new IllegalStateException("Invalid platform type!");
     }
+    
+    public static ApplicationStager getDefaultApplicationStager() {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Removed

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStep.java
```diff
@@ -47,23 +50,33 @@ protected StepPhase executeStep(ExecutionWrapper context) {
...
             desiredState);
+        addStageStateIfNeeded(client, cloudApplication, actionsToExecute);
         getStepLogger().debug(Messages.ACTIONS_TO_EXECUTE, app.getName(), actionsToExecute);
 
         StepsUtil.setAppStateActionsToExecute(execution.getContext(), actionsToExecute);
         return StepPhase.DONE;
     }
+    
+    private void addStageStateIfNeeded(CloudControllerClient client, CloudApplication cloudApplication, Set<ApplicationStateAction> actionsToExecute) {
+        ApplicationStager defaultApplicationStager = ApplicationStagerFactory.getDefaultApplicationStager();
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
```diff
@@ -16,17 +16,29 @@
...
+        }
+        return client.getAllBuildsForApplication(cloudApplication.getMeta()
+            .getGuid())
+            .stream()
+            .anyMatch(cloudBuild -> cloudBuild.getState() != null && cloudBuild.getState() == CloudBuild.BuildState.STAGED
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStepTest.java
```diff
@@ -124,16 +133,25 @@
...
     private final Set<ApplicationStateAction> expectedAppStateActions;
-
+    private final List<CloudBuild> cloudBuilds;
     private RestartParameters appRestartParameters;
 
     @Mock
     private ApplicationStartupStateCalculator appStateCalculator;
 
+    @Mock
+    private CloudControllerClient client;
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStepTest.java
IvanBorislavovDimitrov: Removed

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStepTest.java
IvanBorislavovDimitrov: Thanks for the info, done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStep.java
```diff
@@ -47,23 +55,33 @@ protected StepPhase executeStep(ExecutionWrapper context) {
...
             desiredState);
+        addStageStateIfNeeded(client, retrievedCloudApplication, actionsToExecute);
         getStepLogger().debug(Messages.ACTIONS_TO_EXECUTE, app.getName(), actionsToExecute);
 
         StepsUtil.setAppStateActionsToExecute(execution.getContext(), actionsToExecute);
         return StepPhase.DONE;
     }
+    
+    private void addStageStateIfNeeded(CloudControllerClient client, CloudApplication cloudApplication, Set<ApplicationStateAction> actionsToExecute) {
+        ApplicationStager defaultApplicationStager = ApplicationStagerFactory.createApplicationStager(configuration.getPlatformType());
```
nictas: This is no longer the default application stager. Rename the field to `applicationStager`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStep.java
```diff
@@ -47,23 +55,33 @@ protected StepPhase executeStep(ExecutionWrapper context) {
 
     private StepPhase attemptToExecuteStep(ExecutionWrapper execution) {
         CloudApplication app = StepsUtil.getApp(execution.getContext());
-        ApplicationStartupState currentState = computeCurrentState(execution, app);
+        CloudControllerClient client = execution.getControllerClient();
+        CloudApplication retrievedCloudApplication = client.getApplication(app.getName());
```
nictas: The presence of two CloudApplication objects in the same method is still confusing IMO. Replace line 57 with:
```
// The rest of the fields in the object are null, so we need to retrieve them from the controller.
String appName = StepsUtil.getApp(execution.getContext()).getName();
```
Also, rename `retrievedCloudApplication` to `app` afterwards.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
```diff
@@ -16,17 +16,30 @@
...
 
         return getStagingState(build);
     }
-    
+
+    public boolean isAplicationStagedCorrectly(CloudControllerClient client, CloudApplication cloudApplication) {
+        CloudBuild cloudBuild = client.getBuildsForApplication(cloudApplication.getMeta()
+            .getGuid())
+            .stream()
+            .max((cloudBuild1, cloudBuild2) -> cloudBuild1.getCreatedAt()
```
nictas: Try replacing this line with:
```
.max(CloudBuild::compareTo)
```

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStep.java
```diff
@@ -46,24 +55,39 @@ protected StepPhase executeStep(ExecutionWrapper context) {
...
         StepsUtil.setAppStateActionsToExecute(execution.getContext(), actionsToExecute);
         return StepPhase.DONE;
     }
 
-    private ApplicationStartupState computeCurrentState(ExecutionWrapper execution, CloudApplication app) {
-        CloudControllerClient client = execution.getControllerClient();
+    private void addStageStateIfNeeded(CloudControllerClient client, CloudApplication cloudApplication,
+        Set<ApplicationStateAction> actionsToExecute) {
+        PlatformType platformType = configuration.getPlatformType();
+        if (platformType != PlatformType.CF) {
```
nictas: I take it that this is not implemented for XSA, which is why you had the `getDefaultApplicationStager` method in the first place?
If that's the case, we have two options:
1. Make some minimal implementation in the XsApplicationStager that always returns `true` and track a backlog item for a proper implementation.
2. Actually implement the method in the XsApplicationStager.

I would prefer the second option if it's not too much effort. Either way, though, this `if` should not exist.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
```diff
@@ -16,17 +17,30 @@
...
 
         return getStagingState(build);
     }
-    
+
+    public boolean isApplicationStagedCorrectly(CloudControllerClient client, CloudApplication cloudApplication) {
+        CloudBuild cloudBuild = client.getBuildsForApplication(cloudApplication.getMeta()
+            .getGuid())
+            .stream()
+            .max(Comparator.comparing(build -> build.getMeta()
```
nictas: Nice! :)

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStep.java
```diff
@@ -46,24 +55,39 @@ protected StepPhase executeStep(ExecutionWrapper context) {
...
+            return;
+        }
+        ApplicationStager applicationStager = ApplicationStagerFactory.createApplicationStager(platformType);
+        if (!applicationStager.isApplicationStagedCorrectly(client, cloudApplication)) {
+            actionsToExecute.add(ApplicationStateAction.STAGE);
```
nictas: I'm beginning to think that it would be a good idea if we were to print a warning to the user saying that his application was not staged correctly last time and that we're going to restage it. It would also be good to log the most recent build, so that if there's an open ticket we can investigate through the logs downloaded by "cf dmol".

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStepTest.java
```diff
@@ -1,177 +1,420 @@
...
+    public static class DetermineAppRestartTest extends DetermineDesiredStateAchievingActionsStepBaseTest {
+
+
+        @Parameters
+        public static List<Object[]> getParameters() throws ParseException {
+            return Arrays.asList(new Object[][] {
+                // @formatter:off
+                // (0)
+                {
+                    ApplicationStartupState.STARTED, ApplicationStartupState.STARTED, false, Stream.of(ApplicationStateAction.STAGE, ApplicationStateAction.START, ApplicationStateAction.STOP).collect(Collectors.toSet()), Collections.emptyList(), true, true, true, true, true, true
```
nictas: The current and desired states in this test are always `ApplicationStartupState.STARTED`. This means that they are not actual parameters but constants. Extract them as such.

Also, the `hasAppChanged` field is also always `false`. Extract it as well.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineDesiredStateAchievingActionsStepTest.java
```diff
@@ -1,177 +1,420 @@
...
+                },
+                // (2)
+                {
+                    ApplicationStartupState.STARTED, ApplicationStartupState.STARTED, false, Stream.of(ApplicationStateAction.STAGE, ApplicationStateAction.START, ApplicationStateAction.STOP).collect(Collectors.toSet()), Collections.emptyList(), false, true, false, false, true, false
+                },
+                // (3)
+                {
+                    ApplicationStartupState.STARTED, ApplicationStartupState.STARTED, false, Stream.of(ApplicationStateAction.STAGE, ApplicationStateAction.START, ApplicationStateAction.STOP).collect(Collectors.toSet()), Collections.emptyList(), false, false, true, false, false, true
+                },
+                // (4)
```
nictas: Add a comment explaining why the expected actions always contain `ApplicationStateAction.STAGE`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
```diff
@@ -16,17 +18,45 @@
...
+            execution.getStepLogger()
+                .debug(Messages.NO_BUILD_FOUND_FOR_APPLICATION, cloudApplication.getName());
+            return false;
+        }
+        if (cloudBuild.getState() == CloudBuild.BuildState.STAGED && cloudBuild.getDroplet() != null && cloudBuild.getError() == null) {
+            return true;
+        }
+        execution.getStepLogger()
+            .info(Messages.APPLICATION_NOT_STAGED_CORRECTLY, cloudApplication.getName());
+        execution.getStepLogger()
```
nictas: Why print the same message twice?

IvanBorislavovDimitrov: Fixed

PR #461, opened 2019-03-21
Refactor getters and setters in StepsUtil

nictas:

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
@@ -6,9 +6,7 @@ import java.lang.reflect.Type;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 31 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -1177,4 +1129,37 @@ public static String getString(VariableScope scope, String name, String defaultV
         return value == null ? null : factory.apply(value);
     }
 
+    public static void setAsJsonBinary(VariableScope scope, String name, Object value) {
```
enchobelezirev: Why is the variableScope used and not the DelegateExecution?

enchobelezirev:
I saw them in the StepsUtil... cant delete the comment https://farm8.staticflickr.com/7411/10482984273_a428e23b10_m.jpg

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
nictas: Two reasons:
# Getters/Setters do not need the additional functionality `DelegateExecution` brings. Therefore, using it is kinda like using `ArrayList` instead of `List` everywhere.
# `VariableScope` is much more readable as a name when seen in a getter/setter.

PR #473, opened 2019-04-05
Add error message for service creation failure

boyan-velinov:
When creation of service fails it is not clear for customer where the deployment fails.
This happens because services are created asynchronously and logs for various services are present.

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateServiceStep.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateServiceStep.java
@@ -1,5 +1,6 @@
 package com.sap.cloud.lm.sl.cf.process.steps;
 
+import java.text.MessageFormat;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateServiceStep.java
```diff
@@ -65,9 +65,10 @@
             .createService(client, service, StepsUtil.getSpaceId(context));
     }
 
-
     private void processServiceCreationFailure(CloudServiceExtended service, CloudOperationException e) {
         if (!service.isOptional()) {
+            getStepLogger().warn(Messages.ERROR_CREATING_SERVICE, service.getName(), service.getLabel(), service.getPlan(),
```
petko-boyadzhiev: Why passing more arguments than needed ?
And why not use the extended message.

boyan-velinov: It is leftover and the last argument should not be passed.

PR #476, opened 2019-04-08
Fix test for CreateAppStepWithDocker not executed

IvanBorislavovDimitrov:
The tests for CreateAppStep (test with docker) were not executed

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/CreateAppStepBaseTest.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/CreateAppStepBaseTest.java
@@ -1,0 +1,39 @@
+package com.sap.cloud.lm.sl.cf.process.steps;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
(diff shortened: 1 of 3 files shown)


com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/CreateAppStepBaseTest.java
```diff
@@ -0,0 +1,41 @@
...
+import com.sap.cloud.lm.sl.cf.client.lib.domain.CloudApplicationExtended;
+import com.sap.cloud.lm.sl.cf.client.lib.domain.CloudServiceExtended;
+import com.sap.cloud.lm.sl.cf.core.cf.PlatformType;
+
+public abstract class CreateAppStepBaseTest extends SyncFlowableStepTest<CreateAppStep> {
+
+    protected StepInput stepInput;
+    protected CloudApplicationExtended application;
+    
+    protected abstract void validateClient();
```
nictas: There's no reason to have a `protected abstract` method when it's not being used anywhere in the class. The child classes may not want to validate the interactions with the client. Let them do what they want. :)

IvanBorislavovDimitrov: Okay. Done

PR #484, opened 2019-04-17
make deploy fail if service can't be recreated

VaskoBozhurski:
#### Description: 

This change fixes the issue where
DetermineServiceCreateUpdateServiceActionsStep incorrectly picks the
ServiceAction when the service would fit recreate criteria, but the
delete-services cmd option is not provided by the user.

An example of such case would be if the existing service has a label of
postresql and the new one is mondodb. Instead of failing the deployment
because the service can't be recreated into the new label the step
incorrectly determined that an update of plans could be possible.

This should also fix cases where the existing service is broken and is
missing label/plan.

#### Issue: 
Bug: NGPBUG-83966


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DetermineServiceCreateUpdateServiceActionsStep.java
```diff
@@ -222,6 +231,15 @@ private boolean shouldRecreate(CloudServiceExtended service, CloudService existi
         return haveDifferentTypes || haveDifferentLabels;
     }
 
+    private String computeExistingServiceType(CloudService existingService) {
```
petko-boyadzhiev: Rename

VaskoBozhurski: Done.

PR #490, opened 2019-04-22
change FileStorage.addFile to use File

VaskoBozhurski:
#### Description: 

This changes the FileStorage addFile interface method to take File instead
of InputStream. Previously if an addFile operation failed mid-way any
retry functionality would be impossible as the stream would not be
rewinded back to the start.

#### Issue: 
Bug: LMCROSSITXSADEPLOY-1488

### DIFF
--- a/com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ObjectStoreFileStorage.java
+++ b/com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ObjectStoreFileStorage.java
@@ -1,5 +1,6 @@
 package com.sap.cloud.lm.sl.cf.persistence.services;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.text.MessageFormat;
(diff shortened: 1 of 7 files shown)


com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ObjectStoreFileStorage.java
```diff
@@ -43,15 +44,14 @@ public ObjectStoreFileStorage(BlobStore blobStore, String container) {
...
+    public void addFile(FileEntry fileEntry, File file) throws FileStorageException {
         String entryName = fileEntry.getId();
         long fileSize = fileEntry.getSize()
             .longValue();
         Blob blob = blobStore.blobBuilder(entryName)
-            .payload(fileStream)
+            .payload(file)
             .contentDisposition(fileEntry.getName())
             .contentType(MediaType.OCTET_STREAM.toString())
-            .contentLength(fileSize)
```
VaskoBozhurski: this is automatically set by BlobBuilder.payload(File file) so it's no longer needed.

PR #495, opened 2019-04-25
Introduce application.before-stop.live/idle and make undeployment parallel

enchobelezirev:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/HookPhase.java
```diff
@@ -3,14 +3,19 @@
 import java.text.MessageFormat;
 
 public enum HookPhase {
-    APPLICATION_BEFORE_STOP, APPLICATION_AFTER_STOP, NONE;
+    APPLICATION_BEFORE_STOP_LIVE, APPLICATION_BEFORE_STOP_IDLE, APPLICATION_AFTER_STOP, NONE;
 
     public static HookPhase fromString(String hookPhase) {
         if (hookPhase.equals("application.after-stop")) {
```
nictas: Shouldn't this have the same suffixes? ".idle"/".live"

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/ScaleAppStepTest.java
```diff
@@ -59,8 +59,8 @@ public void testExecute() throws Exception {
 
     private void prepareContext() {
         context.setVariable(Constants.VAR_MODULES_INDEX, 0);
-        context.setVariable("appToDeploy", JsonUtil.toJson(application.toCloudApplication()));
-        //TODO
+        context.setVariable("appToProcess", JsonUtil.toJson(application.toCloudApplication()));
```
nictas: Can we use the constant here?

PR #498, opened 2019-05-07
Refactor hooks execution to be synchronized

enchobelezirev:
#### Description: 
This PR refactors the way the hooks are executed. It introduces call activity which by default is synched in order to wait for all hooks to be executed and then to complete the step execution


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HookProcessGetter.java
```diff
@@ -0,0 +1,20 @@
...
+@Component("hookProcessGetter")
+public class HookProcessGetter {
+
+    public String get(String hookForExecutionString) {
+        Hook hookForExecution = JsonUtil.fromJson(hookForExecutionString, Hook.class);
+        if (TasksHookParser.HOOK_TYPE_TASKS.equals(hookForExecution.getType())) {
+            return Constants.EXECUTE_HOOK_TASKS_SUB_PROCESS_ID;
+        }
+
+        return null;
```
nictas: Throw some kind of exception here.

PR #504, opened 2019-05-13
adopt v3 application binary upload

VaskoBozhurski:
#### Description: 

use the v3 client to upload application binaries so that they are no
longer loaded in-memory before being sent to the Cloud Controller.

#### Issue: 
Bug: LMCROSSITXSADEPLOY-1507


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveReader.java
```diff
@@ -54,45 +44,22 @@ public ZipEntry getFirstZipEntry(ApplicationArchiveContext applicationArchiveCon
         return zipEntry;
     }
 
-    protected void addDataToApplicationResources(String zipEntryName, ApplicationResources applicationResources,
-        ApplicationArchiveContext applicationArchiveContext) throws IOException {
+    protected void addDataToApplicationResources(String zipEntryName, ApplicationArchiveContext applicationArchiveContext)
```
theghost5800: Rename method name

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
```diff
@@ -92,43 +89,38 @@ public StepPhase executeAsyncStep(ExecutionWrapper execution) throws FileStorage
         return StepPhase.POLL;
     }
 
-    private ApplicationResources getApplicationResources(ExecutionWrapper execution, String appArchiveId, String fileName)
-        throws FileStorageException {
-        ApplicationResources applicationResources = new ApplicationResources();
+    private String getNewApplicationDigest(ExecutionWrapper execution, String appArchiveId, String fileName) throws FileStorageException {
         DelegateExecution context = execution.getContext();
+        StringBuilder digestStringBuilder = new StringBuilder();
```
theghost5800: Why StringBuilder is used?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
```diff
@@ -92,43 +89,38 @@ public StepPhase executeAsyncStep(ExecutionWrapper execution) throws FileStorage
         return StepPhase.POLL;
     }
 
-    private ApplicationResources getApplicationResources(ExecutionWrapper execution, String appArchiveId, String fileName)
-        throws FileStorageException {
-        ApplicationResources applicationResources = new ApplicationResources();
+    private String getNewApplicationDigest(ExecutionWrapper execution, String appArchiveId, String fileName) throws FileStorageException {
```
theghost5800: My idea for double reading was to calculate file resources digests and check for already uploaded files and in the second iteration to save files which are missing in the cache. Now double reading is not needed and calculation of digest of the whole content could be done with one reading while writing to zip file. If we keep this behaviour we should add logic to finish UploadAppStep if both digests are equal and skip uploading to the controller.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveReaderTest.java
```diff
@@ -57,39 +52,13 @@
...
-        expectedZipEntries.stream()
-            .forEach(zipEntry -> {
-                normalizedZipEntries.add(FilenameUtils.normalize(zipEntry));
-            });
-        return normalizedZipEntries;
-    }
-
     @ParameterizedTest
     @MethodSource
     public void testFailingReadResources(String mtar, String fileName, String expectedException, long maxFileUploadSize) {
```
theghost5800: rename method

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveReader.java
VaskoBozhurski: renamed

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationArchiveReaderTest.java
VaskoBozhurski: done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UploadAppStep.java
VaskoBozhurski: It's a hack to pass the string value out of the FileDownloadProcessor since it can't change the reference to an outside variable.

PR #506, opened 2019-05-14
Metadata

petko-boyadzhiev:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/detect/ApplicationMtaMetadataParser.java
```diff
@@ -35,19 +37,24 @@ public static ApplicationMtaMetadata parseAppMetadata(CloudApplication app) {
...
+                          .allMatch(Objects::isNull)) {
             return null;
         }
         if (metadataFields.stream()
-            .anyMatch(Objects::isNull)) {
+                          .anyMatch(Objects::isNull)) {
             throw new ParsingException(Messages.MTA_METADATA_FOR_APP_0_IS_INCOMPLETE, app.getName());
         }
+        List<DeployedMtaResource> services = serviceNames.stream()
+                                                         .map(n -> DeployedMtaResource.builder()
```
enchobelezirev: Good you have renamed services to serviceNames but WTF is n?

PR #526, opened 2019-05-28
Leave traces of polled build state in deployer logs. NGPBUG-86867

ddonchev:
NGPBUG-86867
There are currently no traces of the polled build state during staging. The PR should fix this.


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
```diff
@@ -45,7 +45,7 @@ public AsyncExecutionState execute(ExecutionWrapper execution) {
                 .debug(Messages.CHECKING_APP_STATUS, app.getName());
 
             StagingState state = applicationStager.getStagingState(execution, client);
-
+            LOGGER.info("Polled staging state: {} for app: {}", state.getState(), app.getName());
```
enchobelezirev: In that way, the logs of the deploy-service in kibana will be a lot and some logs could be dropped.
If we really need this, we could do it in the catch blocks where the exception is risen.

PR #528, opened 2019-06-03
fix login for global auditor user

VaskoBozhurski:
#### Description: 

The global auditor user has been changed to need origin:uaa when
authenticating which caused previous regular login to fail. This change
adopts the new java client that includes support for origin.

#### Issue: 
Bug: LMCROSSITXSADEPLOY-1539

### DIFF
--- a/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/security/data/termination/DataTerminationService.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/security/data/termination/DataTerminationService.java
@@ -34,8 +34,9 @@ import com.sap.cloud.lm.sl.mta.model.AuditableConfiguration;
 @Component
 public class DataTerminationService {
 
+    private static final String AUTH_ORIGIN = "uaa";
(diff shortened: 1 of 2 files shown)


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/security/data/termination/DataTerminationService.java
```diff
@@ -120,13 +121,13 @@ protected CloudControllerClientImpl getCFClient() {
     private List<String> getDeleteSpaceEvents() {
         CloudControllerClientImpl cfClient = getCFClient();
         CFOptimizedEventGetter cfOptimizedEventGetter = new CFOptimizedEventGetter(cfClient);
-        return cfOptimizedEventGetter.findEvents(SPACE_DELETE_EVENT_TYPE, getDateBeforeTwoDays());
+        return cfOptimizedEventGetter.findEvents(SPACE_DELETE_EVENT_TYPE, getDateBeforeDays(GET_EVENTS_DAYS_BEFORE));
     }
 
-    private String getDateBeforeTwoDays() {
```
VaskoBozhurski: The method name used to say 2 days, but the GET_EVENTS_DAYS_BEFORE was actually set to 1. Changed the name and made the method more generic to avoid similar inconsistencies again.

PR #537, opened 2019-06-06
path of manifest does not seem right,

stevegroom:


stevegroom:
I made a mistake - this PR is not valid. Withdrawn.

PR #553, opened 2019-06-15
Allow escaping of references and placeholders

nictas:
JIRA: LMCROSSITXSADEPLOY-1586


com.sap.cloud.lm.sl.cf.core/src/test/resources/com/sap/cloud/lm/sl/cf/core/helpers/mtad-properties-resolver-test/mtad-with-escaped-references.json
```diff
@@ -0,0 +1,93 @@
+{
```
enchobelezirev: Isn't those tests already covered in the multiapps project?

PR #563, opened 2019-06-27
Update flowable user management

radito3:
#### Description: 
Add setting of process variable in process actions.
Update determineCurrentUser.
Remove unused messages.

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/ResumeProcessAction.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/ResumeProcessAction.java
@@ -31,6 +31,7 @@ public class ResumeProcessAction extends ProcessAction {
                 .isEmpty())
             .collect(Collectors.toList());
 
+        updateUser(userId, superProcessInstanceId);
         for (String processAtReceiveTask : processesAtReceiveTask) {
             triggerProcessInstance(userId, processAtReceiveTask);
         }
@@ -39,8 +40,9 @@ public class ResumeProcessAction extends ProcessAction {
     private void triggerProcessInstance(String userId, String processId) {
         List<Execution> executionsAtReceiveTask = flowableFacade.findExecutionsAtReceiveTask(processId);
         if (!executionsAtReceiveTask.isEmpty()) {
-            executionsAtReceiveTask.stream()
-                .forEach(execution -> flowableFacade.trigger(userId, execution.getId()));
+            for (Execution execution : executionsAtReceiveTask) {
+                flowableFacade.trigger(userId, execution.getId());
+            }
             return;
         }
         LOGGER.warn(MessageFormat.format("Process with id {0} is in undetermined process state", processId));
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/ErrorProcessListener.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/ErrorProcessListener.java
@@ -5,6 +5,8 @@ import javax.inject.Inject;
 import org.flowable.common.engine.api.delegate.event.AbstractFlowableEventListener;
 import org.flowable.common.engine.api.delegate.event.FlowableEngineEvent;
 import org.flowable.common.engine.api.delegate.event.FlowableEvent;
+import org.flowable.engine.HistoryService;
+import org.flowable.engine.impl.context.Context;
 import org.springframework.stereotype.Component;
 
 import com.sap.cloud.lm.sl.cf.core.cf.CloudControllerClientProvider;
@@ -27,9 +29,15 @@ public class ErrorProcessListener extends AbstractFlowableEventListener {
         handler.handle(event);
 
         if (event instanceof FlowableEngineEvent) {
-            ClientReleaser clientReleaser = new ClientReleaser((FlowableEngineEvent) event, clientProvider);
-            clientReleaser.releaseClient();
+            releaseClient((FlowableEngineEvent) event);
         }
+    }
+
+    private void releaseClient(FlowableEngineEvent event) {
+        HistoryService historyService = Context.getProcessEngineConfiguration()
+                .getHistoryService();
+        ClientReleaser clientReleaser = new ClientReleaser(clientProvider);
+        clientReleaser.releaseClientFor(historyService, event.getProcessInstanceId());
     }
 
     @Override
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
@@ -93,31 +93,21 @@ public class StepsUtil {
     }
 
     public static String determineCurrentUser(VariableScope scope, StepLogger stepLogger) {
-        String userId = Authentication.getAuthenticatedUserId();
-        String previousUser = (String) scope.getVariable(Constants.VAR_USER);
-        // Determine the current user
-        if (userId != null && !userId.equals(previousUser)) {
-            stepLogger.debug(Messages.AUTHENTICATED_USER_ID, userId);
-            stepLogger.debug(Messages.PREVIOUS_USER, previousUser);
-        }
+        String userId = (String) scope.getVariable(Constants.VAR_USER);
         if (userId == null) {
-            // If the authenticated user cannot be determined,
-            // use the user saved by the previous service task
-            userId = previousUser;
+            userId = getProcessInitiator(scope, stepLogger);
             if (userId == null) {
-                // If there is no previous user, this must be the first service task
-                // Use the process initiator in this case
-                userId = (String) scope.getVariable(Constants.PARAM_INITIATOR);
-                stepLogger.debug(Messages.PROCESS_INITIATOR, userId);
-                if (userId == null) {
-                    throw new SLException(Messages.CANT_DETERMINE_CURRENT_USER);
-                }
+                throw new SLException(Messages.CANT_DETERMINE_CURRENT_USER);
             }
+            scope.setVariable(Constants.VAR_USER, userId);
         }
-        // Set the current user in the context for use by later service tasks
-        scope.setVariable(Constants.VAR_USER, userId);
-
         return userId;
+    }
+
+    private static String getProcessInitiator(VariableScope scope, StepLogger stepLogger) {
+        String initiator = (String) scope.getVariable(Constants.PARAM_INITIATOR);
+        stepLogger.debug(Messages.PROCESS_INITIATOR, initiator);
+        return initiator;
     }
 
     public static MtaArchiveElements getMtaArchiveElements(VariableScope scope) {
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ClientReleaser.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ClientReleaser.java
@@ -1,9 +1,6 @@
 package com.sap.cloud.lm.sl.cf.process.util;
 
-import org.flowable.common.engine.api.delegate.event.FlowableEngineEvent;
 import org.flowable.engine.HistoryService;
-import org.flowable.engine.impl.context.Context;
-import org.flowable.variable.api.history.HistoricVariableInstance;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -15,23 +12,19 @@ public class ClientReleaser {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClientReleaser.class);
 
-    private final FlowableEngineEvent event;
     private final CloudControllerClientProvider clientProvider;
 
-    public ClientReleaser(FlowableEngineEvent event, CloudControllerClientProvider clientProvider) {
-        this.event = event;
+    public ClientReleaser(CloudControllerClientProvider clientProvider) {
         this.clientProvider = clientProvider;
     }
 
-    public void releaseClient() {
-        HistoryService historyService = Context.getProcessEngineConfiguration()
-            .getHistoryService();
-        String processInstanceId = event.getProcessInstanceId();
-
-        String user = getCurrentUser(historyService, processInstanceId);
-        String spaceName = (String) getHistoricVarInstanceValue(historyService, processInstanceId, Constants.VAR_SPACE).getValue();
-        String orgName = (String) getHistoricVarInstanceValue(historyService, processInstanceId, Constants.VAR_ORG).getValue();
-        String spaceId = (String) getHistoricVarInstanceValue(historyService, processInstanceId,
+    public void releaseClientFor(HistoryService historyService, String processInstanceId) {
+        String user = HistoricVariablesUtil.getCurrentUser(historyService, processInstanceId);
+        String spaceName = (String) HistoricVariablesUtil.getHistoricVarInstanceValue(historyService, processInstanceId,
+            Constants.VAR_SPACE).getValue();
+        String orgName = (String) HistoricVariablesUtil.getHistoricVarInstanceValue(historyService, processInstanceId,
+            Constants.VAR_ORG).getValue();
+        String spaceId = (String) HistoricVariablesUtil.getHistoricVarInstanceValue(historyService, processInstanceId,
             com.sap.cloud.lm.sl.cf.persistence.message.Constants.VARIABLE_NAME_SPACE_ID).getValue();
 
         try {
@@ -40,20 +33,5 @@ public class ClientReleaser {
         } catch (SLException e) {
             LOGGER.warn(e.getMessage());
         }
-    }
-
-    protected String getCurrentUser(HistoryService historyService, String processInstanceId) {
-        String user = (String) getHistoricVarInstanceValue(historyService, processInstanceId, Constants.VAR_USER).getValue();
-        if (user == null) {
-            user = (String) getHistoricVarInstanceValue(historyService, processInstanceId, Constants.PARAM_INITIATOR).getValue();
-        }
-        return user;
-    }
-
-    public HistoricVariableInstance getHistoricVarInstanceValue(HistoryService historyService, String processInstanceId, String parameter) {
-        return historyService.createHistoricVariableInstanceQuery()
-            .processInstanceId(processInstanceId)
-            .variableName(parameter)
-            .singleResult();
     }
 }
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/ProcessAction.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/ProcessAction.java
@@ -1,5 +1,11 @@
 package com.sap.cloud.lm.sl.cf.process.flowable;
 
+import com.sap.cloud.lm.sl.cf.core.cf.CloudControllerClientProvider;
+import com.sap.cloud.lm.sl.cf.process.Constants;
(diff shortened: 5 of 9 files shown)


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/FlowableFacade.java
```diff
@@ -35,6 +36,7 @@
 
     private static final int DEFAULT_JOB_RETRIES = 0;
     private static final int DEFAULT_ABORT_TIMEOUT_MS = 30000;
+    private static final String VAR_USER = "user";
```
valentinEmpy: FlowableFacade is a facade and should be abstract enough. It should not be aware of 'our internal variables' that we use. Read the comment below for more clarification.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/FlowableFacade.java
```diff
@@ -247,13 +249,19 @@ public void trigger(String userId, String executionId) {
             processEngine.getIdentityService()
                 .setAuthenticatedUserId(userId);
             processEngine.getRuntimeService()
-                .trigger(executionId);
+                .trigger(executionId, getVariablesWithUser(userId));
```
valentinEmpy: Several things:
1. The issue is partially solved. trigger() is not the only method which should pass this variable. Consider deleteProcessInstance() (which is called when doing abort), also moveDeadLetterJobsToExecutableJobs() (which is called on retry), check all usages and paths for this issue to occur and prevent it (including internal slp implementation). Some of these places will also lead to the need for this variable in order to use the Client and the issue will persist for the other scenarios.

2. FlowableFacade should not be aware of the presence of specific variables (in this case VAR_USER), it is a facade and should be abstract. This is a responsibility of the Actions (resume, abort, retry etc.) in this case. Could you move the additional logic in this class where it belongs - actions?

Be aware that the Actions are also in cf.core which is semantically incorrect. They are directly connected to cf.process, as they heavily manage our processes. It is better to move them to cf.process, this way you'll also be able to use Constants.VAR_USER from there instead of defining the variable above (which is code duplication).

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -93,30 +92,16 @@ static CloudControllerClient getControllerClient(DelegateExecution context, Clou
...
-                    throw new SLException(Messages.CANT_DETERMINE_CURRENT_USER);
-                }
+                throw new SLException(Messages.CANT_DETERMINE_CURRENT_USER);
             }
+            scope.setVariable(Constants.VAR_USER, userId);
         }
-        // Set the current user in the context for use by later service tasks
-        scope.setVariable(Constants.VAR_USER, userId);
-
+        stepLogger.debug(Messages.CURRENT_USER, userId);
```
valentinEmpy: Do we need to put this message every time this method is called (every single time a controller client is used, every step)? Our MAIN_LOG will heavily increase with many duplicated messages without real change in the user.
Either remove this message, or make it show only when there's a change in the user.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -93,30 +92,16 @@ static CloudControllerClient getControllerClient(DelegateExecution context, Clou
     }
 
     public static String determineCurrentUser(VariableScope scope, StepLogger stepLogger) {
-        String userId = Authentication.getAuthenticatedUserId();
```
enchobelezirev: Why the usage of the Authentication.getAuthenticatedUserId is no more needed?
The Authentication.getAuthenticatedUserId is set by each command that is being executed over the flowable engine - any method in the FlowableFacade is using it.
Something more, if there is no more need of this, consider changing the FlowableFacade not to set it.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/flowable/FlowableFacade.java
enchobelezirev: Addition to the Valio's comment - Is there a reason why passing this userId as variable and not use the Authentication service as before ? Check my below comment

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -93,30 +92,16 @@ static CloudControllerClient getControllerClient(DelegateExecution context, Clou
...
-        if (userId != null && !userId.equals(previousUser)) {
-            stepLogger.debug(Messages.AUTHENTICATED_USER_ID, userId);
-            stepLogger.debug(Messages.PREVIOUS_USER, previousUser);
-        }
+        String userId = (String) scope.getVariable(Constants.VAR_USER);
         if (userId == null) {
-            // If the authenticated user cannot be determined,
-            // use the user saved by the previous service task
-            userId = previousUser;
+            userId = (String) scope.getVariable(Constants.PARAM_INITIATOR);
```
enchobelezirev: Refactor by extracting into different methods - in this way the ifs will be replaced with method invocations.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
radito3: Even though the methods in FlowableFacade set the authenticated user id, the object returned from Authentication.getAuthenticatedUserId is not null only on the first invocation (on process start). At least that's what the results of my debugging proved. That is why I removed its usage in determineCurrentUser. The reason that I left the variable setting in FlowableFacade is so that, as you said, to let flowable use it internally.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -254,6 +254,11 @@ public void trigger(String userId, String executionId) {
         }
     }
 
+    public void setProcessVariable(String executionId, String varName, Object varValue) {
```
valentinEmpy: Make it consistent, rename it to setVariable.
Process is very specific, you're working with Executions which is abstract.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -93,33 +93,24 @@ static CloudControllerClient getControllerClient(DelegateExecution context, Clou
...
-            if (userId == null) {
-                // If there is no previous user, this must be the first service task
-                // Use the process initiator in this case
-                userId = (String) scope.getVariable(Constants.PARAM_INITIATOR);
-                stepLogger.debug(Messages.PROCESS_INITIATOR, userId);
-                if (userId == null) {
-                    throw new SLException(Messages.CANT_DETERMINE_CURRENT_USER);
-                }
-            }
+            userId = getProcessInitiator(scope, stepLogger);
```
valentinEmpy: This 'if' branching... will it ever be considered?
In what cases will the userId be null?
Previously that happened when there is no previous user and we are at the first service task.
Is this still possible?
What I am trying to ask is if this branching is necessary, in what scenario during the process will 'getVariable' for VAR_USER not be sufficient enough and need to use the INITIATOR?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -93,33 +93,24 @@ static CloudControllerClient getControllerClient(DelegateExecution context, Clou
...
+            scope.setVariable(Constants.VAR_USER, userId);
+            stepLogger.debug(Messages.UPDATED_CURRENT_USER, userId);
```
valentinEmpy: Several questions:
1. You already have a debug message inside getProcessInitiator method, and right after that you add this one. They'll be identical, except for a slightly different message, but containing the same userId. This duplication is unnecessary, am I correct?

2. If this is meant to notify on change of the current userId, then I do not think it will do just that. The userId will be updated by the corresponding action, and it will be retrieved on the first line in this method. The branching here is used only if there is no 'userId' as a fallback scenario. Previously to check if the user has changed we had previousUser to compare with. This (current) branching is not handling a normal update of the userId, therefore this message should not be here.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/AbortProcessAction.java
```diff
@@ -19,6 +20,7 @@ public AbortProcessAction(FlowableFacade flowableFacade, List<AdditionalProcessA
 
     @Override
     public void executeActualProcessAction(String userId, String superProcessInstanceId) {
+        flowableFacade.setProcessVariable(superProcessInstanceId, Constants.VAR_USER, userId);
```
valentinEmpy: Check if OptimisticLockingException is possible here.
We discussed the scenario in person.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/ResumeProcessAction.java
```diff
@@ -39,8 +40,11 @@ public void executeActualProcessAction(String userId, String superProcessInstanc
     private void triggerProcessInstance(String userId, String processId) {
         List<Execution> executionsAtReceiveTask = flowableFacade.findExecutionsAtReceiveTask(processId);
         if (!executionsAtReceiveTask.isEmpty()) {
-            executionsAtReceiveTask.stream()
-                .forEach(execution -> flowableFacade.trigger(userId, execution.getId()));
+            executionsAtReceiveTask
+                .forEach(execution -> {
+                    flowableFacade.setProcessVariable(execution.getId(), Constants.VAR_USER, userId);
```
petko-boyadzhiev: Can we please extract the two behaviour lines in a separate method and use method ref ?

enchobelezirev:
The approach with the update of the user in the currently executing action is starting to explode(in the code).
There might be another solution here. The eventing API of the Flowable could be used in order to determine when the Job is being retried or resumed. Check this part of the Flowable documentation:
https://www.flowable.org/docs/userguide/index.html#eventDispatcherEventTypes - it contains information about all the events which are currently supported by the EventDispatcher.
From them, there is one specific - ENTITY_UPDATED, which is used when some entity is being updated in the database(resumed/retried). You could try to create a listener for this event and move the logic for the update of the UserID there. Check the other listeners we have.
However, this event will be triggered when each entity is being updated - check whether this would impact the performance.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ClientReleaser.java
```diff
@@ -15,23 +15,23 @@
...
-        this.event = event;
+    public ClientReleaser(CloudControllerClientProvider clientProvider) {
         this.clientProvider = clientProvider;
     }
 
-    public void releaseClient() {
-        HistoryService historyService = Context.getProcessEngineConfiguration()
-            .getHistoryService();
-        String processInstanceId = event.getProcessInstanceId();
+    public void releaseClient(FlowableEngineEvent event) {
```
enchobelezirev: Not good as the releaseClient is taking event???
Maybe the only below method should be left and this to be deleted becase it will be consistent.

enchobelezirev: Also, try to rename it to releaseClientFor or similar.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ClientReleaser.java
```diff
@@ -15,23 +15,23 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClientReleaser.class);
 
-    private final FlowableEngineEvent event;
     private final CloudControllerClientProvider clientProvider;
+    private final HistoryService historyService = Context.getProcessEngineConfiguration()
```
enchobelezirev: Why this is moved up here? If you do not want to create it every time, maybe passing it to the class would be better.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -254,6 +254,16 @@ public void trigger(String userId, String executionId) {
...
+        processEngine.getRuntimeService()
+                .setVariable(executionId, varName, varValue);
+    }
+
+    public Object getVariable(String executionId, String varName) {
```
enchobelezirev: Personally, I do not like this approach because the FlowableFacade is intended to hide the complex usage of processEngine with all of its services. Thus, such dummy methods here are not well welcomed in the facade.
However, there were plans to remove the facade and each call to a corresponding method from it, to be extracted to the given place. If now, someone sees these two methods, there will be assumption that each variable need to be set/retrieved using these two methods - which is not OK.

What you can do is to provide the RuntimeService where it is needed and use it there(just as we are doing with the HistoryService).

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ClientReleaser.java
```diff
@@ -42,7 +42,7 @@ public void releaseClient() {
         }
     }
 
-    protected String getCurrentUser(HistoryService historyService, String processInstanceId) {
+    public String getCurrentUser(HistoryService historyService, String processInstanceId) {
```
nictas: This method does not belong to this class. The fact that it was `private` and we had to make it `public` when it doesn't fit the rest of the interface is an indication of that. Move it to somewhere else - it will probably fit best in a utility class.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ClientReleaser.java
```diff
@@ -15,19 +14,20 @@
...
-    public void releaseClient() {
+    public void releaseClientFor(String processInstanceId) {
         HistoryService historyService = Context.getProcessEngineConfiguration()
             .getHistoryService();
-        String processInstanceId = event.getProcessInstanceId();
 
+        releaseClientFor(historyService, processInstanceId);
+    }
+
+    public void releaseClientFor(HistoryService historyService, String processInstanceId) {
```
nictas: Do we really need this to be public? Yes, it's good for tests, but it messes up the interface. Usually, we overload methods when it makes sense for users to customize one of the parameters. Here, however, it shouldn't make a difference whether the user supplies the `HistoryService` or uses the default one. They should be the exact same object.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -93,33 +93,23 @@ static CloudControllerClient getControllerClient(DelegateExecution context, Clou
...
-        scope.setVariable(Constants.VAR_USER, userId);
-
         return userId;
     }
 
+    private static String getProcessInitiator(VariableScope scope, StepLogger stepLogger) {
+        String initiator = (String) scope.getVariable(Constants.PARAM_INITIATOR);
+        stepLogger.debug(Messages.PROCESS_INITIATOR, initiator);
+        if (initiator == null) {
+            throw new SLException(Messages.CANT_DETERMINE_CURRENT_USER);
```
nictas: It should be a responsibility of `determineCurrentUser` to throw this exception. `getProcessInitiator` now "knows" that it's the last option of finding the current user, which means that is also "knows" how `determineCurrentUser` works. That's not good, because `determineCurrentUser` is higher in terms of abstraction level. Furthermore, methods should treat other methods as blackboxes and not rely on knowledge of their internal implementation, regardless on where they stand in the abstraction hierarchy.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/ErrorProcessListener.java
```diff
@@ -27,8 +27,8 @@ public void onEvent(FlowableEvent event) {
         handler.handle(event);
 
         if (event instanceof FlowableEngineEvent) {
-            ClientReleaser clientReleaser = new ClientReleaser((FlowableEngineEvent) event, clientProvider);
-            clientReleaser.releaseClient();
+            ClientReleaser clientReleaser = new ClientReleaser(clientProvider);
```
nictas: Extract the body of this if in a separate method. That way, you'll reduce the explosion of braces (), which now exists because of the casting, and you'll reduce the number of things this method does (which is always a good thing :)).

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/ResumeProcessAction.java
```diff
@@ -39,8 +40,9 @@ public void executeActualProcessAction(String userId, String superProcessInstanc
     private void triggerProcessInstance(String userId, String processId) {
         List<Execution> executionsAtReceiveTask = flowableFacade.findExecutionsAtReceiveTask(processId);
         if (!executionsAtReceiveTask.isEmpty()) {
-            executionsAtReceiveTask.stream()
-                .forEach(execution -> flowableFacade.trigger(userId, execution.getId()));
+            for (Execution execution : executionsAtReceiveTask) {
+                flowableFacade.trigger(userId, execution.getId());
```
nictas: Nice! Don't show this to Petko, though. :D

petko-boyadzhiev: image

petko-boyadzhiev: Venkat is turning in his bed...

nictas: On the plus side, no one from the team will curse you when looking at a stack trace coming from this loop. Can't say the same about Petko, though... @petko-boyadzhiev If some day your ears are burning, try to remember whether you've used .forEach() that day. :P

petko-boyadzhiev: The north will remember. The winter streams are going to pluck you loops from the ground and blow them to the past...

nictas: https://www.youtube.com/watch?v=UY3oZH6Piz4&feature=youtu.be&t=126

PR #564, opened 2019-06-27
Logging the build guid when new build is created

petko-boyadzhiev:
Prerequesite for fixing LMCROSSITXSADEPLOY-1590

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/AbstractProcessExecutionListener.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/AbstractProcessExecutionListener.java
@@ -1,13 +1,5 @@
 package com.sap.cloud.lm.sl.cf.process.listeners;
 
-import javax.inject.Inject;
-
-import org.flowable.engine.delegate.DelegateExecution;
-import org.flowable.engine.delegate.ExecutionListener;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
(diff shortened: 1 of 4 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
```diff
@@ -91,10 +91,12 @@ public StepPhase stageApp(DelegateExecution context, CloudControllerClient clien
...
 
         stepLogger.info(Messages.STAGING_APP, app.getName());
-
-        context.setVariable(Constants.VAR_BUILD_GUID, client.createBuild(packageGuid)
-            .getMetadata()
-            .getGuid());
+        CloudBuild build = client.createBuild(packageGuid);
+        UUID buildGuid = build.getMetadata()
+                              .getGuid();
+        stepLogger.getProcessLogger()
```
enchobelezirev: What about using this method: StepLogger::infoWithoutProgressMessage which is only logging into the step and the process log?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
```diff
@@ -91,10 +91,12 @@ public StepPhase stageApp(DelegateExecution context, CloudControllerClient clien
...
+                  .info(MessageFormat.format(Messages.BUILD_CREATED, buildGuid));
```
enchobelezirev: Is there any particular reason that this is an info?
Would the debug also apply in this case?
If you are worried that the debug would be printed on the client side - dont be. The client only prints the INFO progress messages.

PR #567, opened 2019-06-28
Enable creation of service-keys when service-name parameter is defined

IvankaTashkova:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v2/ServiceKeysCloudModelBuilder.java
```diff
@@ -50,11 +50,16 @@ protected CloudServiceKey getServiceKey(Resource resource, Map<String, Object> s
         if (parameters == null) {
             parameters = Collections.emptyMap();
         }
+        String serviceName = resource.getParameters()
+            .get(SupportedParameters.SERVICE_NAME) == null ? resource.getName()
```
petko-boyadzhiev: extract resource.getParameters()
            .get(SupportedParameters.SERVICE_NAME) to a var and reuse it

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v2/ServiceKeysCloudModelBuilder.java
```diff
@@ -50,11 +50,16 @@ protected CloudServiceKey getServiceKey(Resource resource, Map<String, Object> s
         if (parameters == null) {
             parameters = Collections.emptyMap();
         }
+        String serviceName = resource.getParameters()
```
petko-boyadzhiev: I think this can be replaced by the NameUtils call

PR #577, opened 2019-07-09
Refer NGPBUG components

IvankaTashkova:


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/SyncFlowableStep.java
```diff
@@ -113,7 +118,18 @@ protected void onStepError(DelegateExecution context, Exception e) throws Except
 
     protected void processException(Exception e, String detailedMessage) throws Exception {
         e = handleControllerException(e);
-        throw getExceptionConstructor(e).apply(e, detailedMessage + ": " + e.getMessage());
+        String tail = computeTail(e);
+        throw getExceptionConstructor(e).apply(e, detailedMessage + ": " + e.getMessage() + " " + tail);
```
boyan-velinov: Invoke the method directly in apply() without a new variable.
Rename the method to be related to "support incident hint"

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/SyncFlowableStep.java
```diff
@@ -113,7 +118,18 @@ protected void onStepError(DelegateExecution context, Exception e) throws Except
...
+    }
+
+    protected String computeTail(Exception e) {
+        if(e instanceof CloudServiceBrokerException) {
```
boyan-velinov: Maybe we should avoid adding anything for ContentException.
Imagine we have invalid mtad.yaml, e.g. some required field is missing. It is clear the it is modelling error, and incident could not help much.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateServiceBrokerStep.java
```diff
@@ -155,8 +157,12 @@ protected void updateServiceBroker(DelegateExecution context, CloudServiceBroker
                         getStepLogger().warn(Messages.UPDATE_OF_SERVICE_BROKERS_FAILED_403, serviceBroker.getName());
                         return;
                     }
+                    getStepLogger().error(Messages.ERROR_CREATING_SERVICE_BROKERS, e, ExceptionMessageTailMapper
```
boyan-velinov: We should avoid call computeTail in steps.
Instead we have to override the method getStepErrorMessage with extended error message. 

Also I am not completely sure if this is true when registering space scoped service brokers, who are owned by the MTA, to  to mapped to specific support ticket component.
For instance, I declare my "boyan-broker" and its registering fails for some reason. What would be the error message? I believe the error message should not point to any service broker component at all.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -160,20 +162,24 @@ private void deleteService(CloudControllerClient client, String serviceName) {
...
-    private CloudOperationException evaluateCloudOperationException(CloudOperationException e, String serviceName) {
+    private CloudOperationException evaluateCloudOperationException(CloudOperationException e, String serviceName, String label) {
         if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
-            getStepLogger().warn(e, Messages.COULD_NOT_DELETE_SERVICE, serviceName);
+            getStepLogger().warn(MessageFormat.format(Messages.COULD_NOT_DELETE_SERVICE, serviceName), e,
+                ExceptionMessageTailMapper.computeExceptionMessageTail(configuration, CloudComponents.ServiceBroker, label));
             return null;
         }
         if (e.getStatusCode() == HttpStatus.BAD_GATEWAY) {
+            getStepLogger().error(MessageFormat.format(Messages.COULD_NOT_DELETE_SERVICE, serviceName), e,
```
boyan-velinov: The same here. We should avoid call getStepLogger().error(..) in steps. Instead, throw CloudServiceBrokerException with extended error message.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/StepLogger.java
```diff
@@ -92,6 +92,11 @@ public void error(String message) {
         sendProgressMessage(message, ProgressMessageType.ERROR);
     }
 
+    public void error(String message, Exception e, String tail) {
```
boyan-velinov: Think about renaming tail to something more suitable.
E.g. supportMessageHint

If the intention is to use for more generic purpose and can be extend further, it can be named with "messageDetails" or "messageHints"

PR #578, opened 2019-07-10
Ignore schema from routes parameter

boyan-velinov:
#### Description: 
We should behave the same way as CF CLI when one has defined
routes containing http(s) schema. Schema is not used at all in CF,
so we should ignore it before sending it to cf java client, that
does not expect it.

### DIFF
--- a/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/parser/UriParametersParser.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/parser/UriParametersParser.java
@@ -16,6 +16,7 @@ import org.apache.commons.collections4.CollectionUtils;
 
 import com.sap.cloud.lm.sl.cf.core.model.SupportedParameters;
 import com.sap.cloud.lm.sl.cf.core.util.ApplicationURI;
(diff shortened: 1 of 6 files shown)


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/parser/UriParametersParser.java
```diff
@@ -133,6 +133,8 @@ private void addHostBasedUris(Set<String> uris, String domain, List<String> host
             .stream()
             .map(routesMap -> (String) routesMap.get(SupportedParameters.ROUTE))
             .filter(Objects::nonNull)
+            .map(ApplicationURI::new)
+            .map(ApplicationURI::toString)
```
ikasarov: Perhaps, instead of invoking new and toSting, we can just call a static version of the stripScheme method, since this is essentially the only thing we're using the ApplicationURI object for.

ikasarov: Also, this can just be moved to line 145 as an else statement - if we don't want to modify the route - we still strip the scheme at least (the modify route logic will do it regardless)

PR #582, opened 2019-07-15
Remove streams quiet closing & refactoring

radito3:
Remove Unused Util classes.
And some minor refactoring.


com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ObjectStoreFileStorage.java
```diff
@@ -101,20 +99,16 @@ public int deleteFilesModifiedBefore(Date modificationTime) throws FileStorageEx
...
-        InputStream fileContentStream = null;
-        try {
-            Blob blob = blobStore.getBlob(container, fileEntry.getId());
-            if (blob == null) {
-                throw new FileStorageException(
-                    MessageFormat.format(Messages.FILE_WITH_ID_AND_SPACE_DOES_NOT_EXIST, fileEntry.getId(), fileEntry.getSpace()));
-            }
-            Payload payload = blob.getPayload();
-            fileContentStream = payload.openStream();
+        Blob blob = blobStore.getBlob(container, fileEntry.getId());
```
valentinEmpy: This here throws ContainerNotFoundException which is a runtime exception, previously was wrapped in FileStorageException (a checked exception). In our code, there are specific places where we catch and handle FileStorageException, rethrowing with a more meaningful message. With this change, the wrapping no longer occurs, hence the inability to enrich the message later.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerProvider.java
```diff
@@ -23,7 +23,7 @@
     private static final String DEFAULT_LOG_DIR = "logs";
     private static final String LOG_FILE_EXTENSION = ".log";
 
-    private Map<String, ProcessLogger> loggersCache = new ConcurrentHashMap<>();
+    private ConcurrentMap<String, ProcessLogger> loggersCache = new ConcurrentHashMap<>();
```
valentinEmpy: Why is this explicitness needed? Is there any reason to go from something more abstract to something more concrete?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -102,13 +101,10 @@ public Response uploadMtaFile(HttpServletRequest request, SecurityContext securi
...
 
-            InputStream in = null;
-            try {
-                in = item.openStream();
+            try (InputStream in = item.openStream()) {
                 FileEntry entry = fileService.addFile(spaceGuid, item.getName(), getConfiguration().getFileUploadProcessor(), in);
                 uploadedFiles.add(entry);
-            } finally {
-                IOUtils.closeQuietly(in);
+            } catch (IOException ignored) {
```
valentinEmpy: Leaving an exception with a name 'ignored' only makes Lint ignore the warning that this block is empty, but it does not make it more understandable why it actually is left empty. Stick to name 'e', and add a comment explaining why this block is empty. This is also the standard recommendation from sonar.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerProvider.java
radito3: ConcurrentMap overrides some methods in Map (apart from the implemented ones in ConcurrentHashMap). That's why, since it's a private variable and doesn't need to be exported, I've made this concretization. To use the benefits of a concurrent map.

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/api/impl/FilesApiServiceImpl.java
radito3: If an exception occurred in the previous state of the code, it was ignored. That's why this is left like that. I will change the name to 'e' and add //Intentionally ignored.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ProcessLoggerProvider.java
valentinEmpy: I understood the statement that it is a private variable and does not need to be exported somewhere else. But to the other statement... the Map *declared* here, has already been *initialized* as a ConcurrentHashMap, which already implements ConcurrentMap. Using this Map will make use of all the synchronization benefits. How does changing the declaration will use any more benefits? This is just the declaration type of the variable.

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/api/impl/FilesApiServiceImpl.java
valentinEmpy: I understand that. Try to make it more descriptive, for example "// Close quietly", makes more sense to me when I see it than "// Intentionally ignored". The second is already apparent.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ObjectStoreFileStorage.java
```diff
@@ -65,19 +64,18 @@ public void addFile(FileEntry fileEntry, File file) throws FileStorageException
...
-            .stream()
-            .map(StorageMetadata::getName)
-            .collect(Collectors.toSet());
-
         return fileEntries.stream()
-            .filter(fileEntry -> {
-                String id = fileEntry.getId();
-                return !existingFiles.contains(id);
-            })
+            .filter(this::isFileNotPresent)
```
nictas: This will now make the same request to the blob store for each file entry. This will (probably) degrade the performance by a lot.

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ObjectStoreFileStorage.java
```diff
@@ -146,16 +148,12 @@ private void putBlobWithRetries(Blob blob, int retries) {
...
             .stream()
             .filter(filter)
             .map(StorageMetadata::getName)
-            .collect(Collectors.toSet());
-
-        if (!entriesToDelete.isEmpty()) {
-            blobStore.removeBlobs(container, entriesToDelete);
-        }
-        return entriesToDelete.size();
+            .peek(entry -> blobStore.removeBlob(container, entry))
```
nictas: I'm not sure whether the `removeBlob` implementation is as fast as calling `removeBlobs` with a list of entries. Furthermore, previously it was clear that this method was doing more than it should've. Now it's less so. If you want to make it more readable, just extract the stream as it was before in a separate method.

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -102,13 +101,10 @@ public Response uploadMtaFile(HttpServletRequest request, SecurityContext securi
...
 
-            InputStream in = null;
-            try {
-                in = item.openStream();
+            try (InputStream in = item.openStream()) {
                 FileEntry entry = fileService.addFile(spaceGuid, item.getName(), getConfiguration().getFileUploadProcessor(), in);
                 uploadedFiles.add(entry);
-            } finally {
-                IOUtils.closeQuietly(in);
+            } catch (IOException e) { // Handle IO exceptions quietly
```
nictas: This is now different, though. The previous code's intent was to handle IOExceptions during the closing of the `InputStream`. Now, it also handles/ignores IOExceptions when opening the stream as well. This may result in problems where we report that something was uploaded, but it actually wasn't.

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/monitoring/FssMonitor.java
```diff
@@ -17,8 +17,8 @@
 @Component
 public class FssMonitor {
 
-    Map<File, Long> usedSpaceMap = new Hashtable<>(1);
```
nictas: There's an important difference between `Hashtable` and `HashMap`. `Hashtable` is synchronized. The method `calculateUsedSpace` may get called by different threads, so use a synchronized map implementation in order to avoid problems.

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/api/impl/FilesApiServiceImpl.java
radito3: Should the error propagate?

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ObjectStoreFileStorage.java
radito3: I looked into the implementations of BlobStore to check this. I'm not sure what concrete implementation we use for the injected BlobStore but as far as I see they all implement the removeBlobs method like this 
for (string name : blobNames) {
removeBlob(name)
}

com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/ObjectStoreFileStorage.java
radito3: I will revert this change.

PR #585, opened 2019-07-16
Make AppExecutionState thread safe

radito3:
Update AppExecutionState from an enum to a class to fix thread safety
issue with message


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollExecuteAppStatusExecution.java
```diff
@@ -28,24 +28,29 @@
...
-    enum AppExecutionStatus {
-        EXECUTING(""),
-        SUCCEEDED(""),
-        FAILED("");
+    enum AppExecutionState {
+        EXECUTING, SUCCEEDED, FAILED
+    }
 
-        private String message;
+    private static class AppExecutionStatus {
```
nictas: Rename to `AppExecutionDetailedState`. `State` and `Status` are synonyms and it's not a good idea to use synonyms to indicate two different things - it's confusing.

PR #591, opened 2019-07-24
enhance stage logic on 404 error

VaskoBozhurski:
#### Description: 

this change enhances PollStageAppStatusExecution to check if the
application exist in case of build not found errors.

#### Issue: 
Bug: LMCROSSITXSADEPLOY-1590


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationStager.java
```diff
@@ -35,10 +36,26 @@ public StagingState getStagingState(DelegateExecution context) {
...
+            if (HttpStatus.NOT_FOUND.equals(e.getStatusCode())) {
+                checkIfApplicationExists(context);
+            }
+            throw e;
+        }
+    }
+
+    private void checkIfApplicationExists(DelegateExecution context) {
+        CloudApplicationExtended app = StepsUtil.getApp(context);
+        client.getApplication(app.getName());
```
nictas: Add a comment here.

PR #595, opened 2019-07-25
Minor refactoring

radito3:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/v2/ApplicationCloudModelBuilder.java
```diff
@@ -200,14 +203,14 @@ protected String getInvalidServiceBindingConfigTypeErrorMessage(String moduleNam
     }
 
     protected List<String> getApplicationServices(Module module, Predicate<ResourceAndResourceType> filterRule) {
-        List<String> services = new ArrayList<>();
+        Set<String> services = new HashSet<>();
         for (RequiredDependency dependency : module.getRequiredDependencies()) {
             ResourceAndResourceType pair = getApplicationService(dependency.getName());
```
enchobelezirev: Should this be named pair?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/parser/UriParametersParser.java
```diff
@@ -72,15 +69,12 @@ public UriParametersParser(String defaultHost, String defaultDomain, String host
...
 
     public List<String> getApplicationDomains(List<Map<String, Object>> parametersList) {
-        List<String> domains;
         List<String> routes = getApplicationRoutes(parametersList);
         if (!routes.isEmpty()) {
-            domains = getDomainsFromRoutes(routes);
+            return getDomainsFromRoutes(routes);
         } else {
-            domains = getDomainValues(parametersList);
+            return getDomainValues(parametersList);
```
enchobelezirev: Remove the else and directly return the getDomainValues(parametersList);

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/parser/UriParametersParser.java
```diff
@@ -126,24 +120,17 @@ private void addHostBasedUris(Set<String> uris, String domain, List<String> host
...
+        if (routesMaps == null) {
+            return Collections.emptyList();
         }
 
-        return allNonNullRoutes.stream()
-            .map(UriUtil::stripScheme)
+        return routesMaps.stream()
+            .map(routesMap -> (String) routesMap.get(SupportedParameters.ROUTE))
+            .filter(Objects::nonNull)
+            .map(route -> modifyRoute ? modifyUri(route, parametersList) : UriUtil.stripScheme(route))
```
enchobelezirev: Extract this to separate method.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/security/serialization/SecureSerializer.java
```diff
@@ -39,15 +39,14 @@ private boolean isSensitive(String value) {
...
-        List<Element> elementsToModify = new ArrayList<>();
-        element.getMembers()
-            .forEach(nestedElement -> {
-                if (isSensitive(nestedElement)) {
-                    elementsToModify.add(0, nestedElement);
-                } else {
-                    modifySensitiveElements(nestedElement);
-                }
-            });
+        Deque<Element> elementsToModify = new LinkedList<>();
```
enchobelezirev: Why the Deque is needed? Are you removing from both ends?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStartAppStatusExecution.java
```diff
@@ -139,38 +139,36 @@ private void showInstancesStatus(ExecutionWrapper execution, String appName, Lis
             for (InstanceInfo instance : instances) {
                 final String state = instance.getState()
                     .toString();
-                final Integer stateCount = stateCounts.get(state);
-                stateCounts.put(state, (stateCount == null) ? 1 : (stateCount + 1));
+                stateCounts.compute(state, (k, stateCount) -> stateCount == null ? 1 : stateCount + 1);
```
enchobelezirev: Extract this to method, because it is not visible.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStartAppStatusExecution.java
```diff
@@ -139,38 +139,36 @@ private void showInstancesStatus(ExecutionWrapper execution, String appName, Lis
...
-            if (instance.getState()
-                .equals(state)) {
-                count++;
-            }
-        }
-        return count;
+        return (int) instances.stream()
+            .filter(instance -> instance.getState()
+                .equals(state))
+            .count();
```
enchobelezirev: Another possible thing is to collect a list and then get the size(). Or make the method return long instead ?

com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/State.java
```diff
@@ -12,11 +12,15 @@
 public enum State {
```
valentinEmpy: Can we simplify this enum further?
Its elements use an XmlEnumValue annotation, using the same value as the enum's name.
This is in the doc of XmlEnumValue:
"In the absence of this annotation, Enum.name() is used as the XML representation. "
Another point is the 'value' variable inside the enum, which also uses the same. Can these two be completely removed to simplify the enum?
Check the usages in the code to verify this.

PR #604, opened 2019-07-31
Split method into parts

enchobelezirev:
Change-Id: I8b07012aa36cbb574e9a55921d34fcd36cc39fd9

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceOperationUtil.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceOperationUtil.java
@@ -15,14 +15,19 @@ public class ServiceOperationUtil {
     }
 
     public static Map<String, Object> getServiceKeyCredentials(CloudControllerClient client, String serviceName, String serviceKeyName) {
-        List<CloudServiceKey> existingServiceKeys = client.getServiceKeys(serviceName);
-        CloudServiceKey foundServiceKey = existingServiceKeys.stream()
-                                                             .filter(existingServiceKey -> existingServiceKey.getName()
-                                                                                                             .equals(serviceKeyName))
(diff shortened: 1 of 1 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceOperationUtil.java
```diff
@@ -15,14 +15,20 @@ private ServiceOperationUtil() {
...
-                                                                                                serviceKeyName,
-                                                                                                serviceName));
+        CloudServiceKey foundServiceKey = getServiceKey(client, serviceName, serviceKeyName);
         return foundServiceKey.getCredentials();
     }
+
+    public static CloudServiceKey getServiceKey(CloudControllerClient client, String serviceName, String serviceKeyName) {
+        List<CloudServiceKey> existingServiceKeys = getServiceKeys(client, serviceName);
+        return existingServiceKeys.stream()
+            .filter(existingServiceKey -> existingServiceKey.getName()
```
nictas: Use the new formatter.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceOperationUtil.java
```diff
@@ -15,14 +15,20 @@ private ServiceOperationUtil() {
...
+                .equals(serviceKeyName))
+            .findFirst()
+            .orElseThrow(() -> new SLException(Messages.ERROR_RETRIEVING_REQUIRED_SERVICE_KEY_ELEMENT, serviceKeyName, serviceName));
+    }
+
+    public static List<CloudServiceKey> getServiceKeys(CloudControllerClient client, String serviceName) {
```
nictas: Why do we need such a method? Why not call `client.getServiceKeys` directly?

PR #606, opened 2019-08-02
Use NameUtil where it is necesary

enchobelezirev:
#### Description: 
User NameUtil::getServices where it is possible

#### Issue:
LMCROSSITXSADEPLOY-1465


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/NameUtil.java
```diff
@@ -122,8 +122,11 @@ public static String getApplicationName(Module module) {
     }
 
     public static String getServiceName(Resource resource) {
+        return getServiceNameOrDefault(resource, null);
+    }
+    public static String getServiceNameOrDefault(Resource resource, String defaultServiceName) {
         return (String) resource.getParameters()
-                                .get(SupportedParameters.SERVICE_NAME);
+            .getOrDefault(SupportedParameters.SERVICE_NAME, defaultServiceName);
```
nictas: There should be no need for this default. It should be set every time in `CollectSystemParametersStep`.

PR #616, opened 2019-08-15
Add mta archive signature verification functionality

IvanBorislavovDimitrov:
Verify archive signature. Check whether the archive is signed by Symantec.
Check CN of the certificate -> No default (can be set through env var).
https://www.tbs-certificates.co.uk/FAQ/en/Symantec_Class3_SHA256_code.html -> Symantec certificate
Option to activate verification -> --verify-archive-signature (deploy, bg-deploy)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -48,32 +58,72 @@ protected String getStepErrorMessage(DelegateExecution context) {
...
+        getStepLogger().debug(Messages.BUILDING_ARCHIVE_FROM_PARTS);
+        new ExecutionRetrier().executeWithRetry(createArchiveFromParts(context, archivePartEntries));
+    }
+
+    private List<FileEntry> getArchivePartEntries(DelegateExecution context, String[] appArchivePartsId) {
+        List<FileEntry> archivePartEntries = new ArrayList<>();
+        for (String appArchivePartId : appArchivePartsId) {
+            archivePartEntries.add(findFile(context, appArchivePartId));
+        }
+        return archivePartEntries;
```
radito3: This could be substituted with Arrays.stream().map().collect()

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -48,32 +58,72 @@ protected String getStepErrorMessage(DelegateExecution context) {
...
     }
 
-    private void validateVersionRule(DelegateExecution context) {
-        String versionRuleString = (String) context.getVariable(Constants.PARAM_VERSION_RULE);
+    private void validateStartTimeout(DelegateExecution context) {
+        Object parameter = context.getVariable(Constants.PARAM_START_TIMEOUT);
+        if (parameter == null) {
+            return;
+        }
+        int startTimeout = (Integer) parameter;
```
radito3: Why not cast directly to the primitive type?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/certficate_checker/DefaultMtaCertificateChecker.java
```diff
@@ -0,0 +1,56 @@
...
+    private Verifier verifier;
+
+    @Override
+    public void checkCertificates(URL url, List<X509Certificate> certificates, String certificateCN) {
+        verifier.verify(certificates, certificateCN, url);
+    }
+
+    @Override
+    public List<X509Certificate> readProvidedCertificates(String certificatesFilename) {
+        InputStream symantecInputStream = getClass().getResourceAsStream(certificatesFilename);
```
radito3: Can we put this in a try-with-resources block?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+    }
+
+    private void verifyJarEntries(List<X509Certificate> targetCerts, List<JarEntry> jarEntries, String certificateCN) {
+        for (JarEntry jarEntry : jarEntries) {
+            Certificate[] entryCertificates = jarEntry.getCertificates();
+            if (entryCertificates == null || entryCertificates.length == 0) {
+                checkJarEntryName(jarEntry);
+                continue;
```
radito3: Invert the condition to avoid this continue

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+        return chainIndex;
+    }
+
+    private List<X509Certificate> getX509CertificateChain(Certificate[] certificates, int startIndex, int certChainSize) {
+        List<X509Certificate> x509Certificates = new ArrayList<>();
+        for (int i = 0; i < certChainSize; i++) {
+            x509Certificates.add((X509Certificate) certificates[startIndex + i]);
+        }
+        return x509Certificates;
```
radito3: This could be substituted for IntStream.range(0, certChainSize).mapToObj().collect()

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
IvanBorislavovDimitrov: Done. Substituted.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
IvanBorislavovDimitrov: Done.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/certficate_checker/DefaultMtaCertificateChecker.java
IvanBorislavovDimitrov: Yes, we can. Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: If I invert the condition and remove the continue one or both of the actions will be executed, I want to check the jar entry name or verify jar entry is signed (One of them). I use it instead of else.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done. Replaced

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfiguration.java
```diff
@@ -40,8 +40,41 @@
 @Lazy(false)
 public class ApplicationConfiguration {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(ApplicationConfiguration.class);
-
+    // Default values:
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: To rearrange the code so that the constants are at the top.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -73,6 +73,8 @@
 
 public class StepsUtil {
 
+    public static final String DEPLOY_ID_PREFIX = "deploy-";
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Yes, moved the constant to be at the top.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ArchiveMerger.java
```diff
@@ -52,8 +48,11 @@ public void createArchiveFromParts(List<FileEntry> archivePartEntries) {
             stepLogger.info(Messages.ERROR_MERGING_ARCHIVE);
             throw new SLException(e, Messages.ERROR_MERGING_ARCHIVE_PARTS, e.getMessage());
         } finally {
-            deleteMergedFile(archiveMerger);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: This file will be deleted in ValidateDeployParametersStep. I moved it there because I need to use the archive (saved on the container) to execute the verifications.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/certficate_checker/DefaultMtaCertificateChecker.java
```diff
@@ -0,0 +1,56 @@
...
+        try {
+            CertificateFactory certificateFactory = CertificateFactory.getInstance(Constants.CERTIFICATE_TYPE_X_509);
+            Collection<? extends Certificate> certificates = certificateFactory.generateCertificates(symantecInputStream);
+            return (List<X509Certificate>) certificates;
+        } catch (CertificateException e) {
+            throw new SLException(e);
+        } finally {
+            closeInputStream(symantecInputStream);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: I agree it's better if the stream is closed quietly. Refactored using try-with-resources.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        return jarEntries;
+    }
+
+    private void readJarEntry(JarFile jarFile, List<JarEntry> jarEntries, Enumeration<JarEntry> entries) throws IOException {
+        JarEntry jarEntry = entries.nextElement();
+        validateJarEntryPath(jarFile, jarEntry);
+        if (jarEntry.isDirectory()) {
+            return;
+        }
+        jarEntries.add(jarEntry);
```
nictas: Return the `jarEntry` and add it to the list in the `getJarEntries` method. That way you'll reduce the number of arguments given to this method by 1.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        readStream(jarFile, jarEntry);
+    }
+
+    private void validateJarEntryPath(JarFile jarFile, JarEntry jarEntry) throws IOException {
```
nictas: Check whether this does the same as `FileUtils.validatePath`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        Path currentRelativePath = Paths.get("");
+        File jarEntryFile = new File(jarEntry.getName());
+        if (!jarEntryFile.getCanonicalPath()
+                         .startsWith(currentRelativePath.toAbsolutePath() + File.separator)) {
+            throw new IllegalAccessError(MessageFormat.format(Messages.ENTRY_IS_TRYING_TO_LEAVE_TARGET_DIR, jarFile.getName()));
+        }
+    }
+
+    private void readStream(JarFile jarFile, JarEntry jarEntry) throws IOException {
```
nictas: Rename to `verifySignature`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+            Certificate[] entryCertificates = jarEntry.getCertificates();
+            if (entryCertificates == null || entryCertificates.length == 0) {
+                checkJarEntryName(jarEntry);
+                continue;
+            }
+            verifyJarEntryIsSigned(targetCerts, entryCertificates, certificateCN);
+        }
+    }
+
+    private void checkJarEntryName(JarEntry jarEntry) {
```
nictas: Rename to `verifyIsMetaInformation` for consistency.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        while (!(entryCertificateChain = getAChain(entryCertificates, startIndex)).isEmpty()) {
+            if (validCertificateChain(targetCerts, entryCertificateChain, certificateCN)) {
+                return true;
+            }
+            startIndex += entryCertificateChain.size();
+        }
+        return false;
+    }
+
+    List<X509Certificate> getAChain(Certificate[] entryCertificates, int startIndex) {
```
nictas: Typo - `getAsChain`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+        int nextChainIndex = findNextChainIndex(entryCertificates, startIndex);
+        int certificateChainSize = (nextChainIndex - startIndex) + 1;
+        return getX509CertificateChain(entryCertificates, startIndex, certificateChainSize);
+    }
+
+    private int findNextChainIndex(Certificate[] certs, int startIndex) {
+        int chainIndex;
+        for (chainIndex = startIndex; chainIndex < certs.length - 1; chainIndex++) {
+            if (!Objects.equals(((X509Certificate) certs[chainIndex + 1]).getSubjectDN(),
```
nictas: Can we always assume that the certificate is an instance of `X509Certificate`? Shouldn't we validate the type first?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+    List<X509Certificate> getAChain(Certificate[] entryCertificates, int startIndex) {
+        if (startIndex > entryCertificates.length - 1) {
+            return Collections.emptyList();
+        }
+        int nextChainIndex = findNextChainIndex(entryCertificates, startIndex);
+        int certificateChainSize = (nextChainIndex - startIndex) + 1;
+        return getX509CertificateChain(entryCertificates, startIndex, certificateChainSize);
+    }
+
+    private int findNextChainIndex(Certificate[] certs, int startIndex) {
```
nictas: Use consistent naming - rename `certs` to `certificates`.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfiguration.java
```diff
@@ -40,8 +40,41 @@
 @Lazy(false)
 public class ApplicationConfiguration {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(ApplicationConfiguration.class);
-
+    // Default values:
+    public static final List<Platform> DEFAULT_PLATFORMS = Collections.emptyList();
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfiguration.java
```diff
@@ -83,43 +116,10 @@
     static final String CFG_SPACE_DEVELOPER_CACHE_TIME_IN_SECONDS = "SPACE_DEVELOPER_CACHE_TIME_IN_SECONDS";
     static final String CFG_CONTROLLER_CLIENT_CONNECTION_POOL_SIZE = "CONTROLLER_CLIENT_CONNECTION_POOL_SIZE";
     static final String CFG_CONTROLLER_CLIENT_THREAD_POOL_SIZE = "CONTROLLER_CLIENT_THREAD_POOL_SIZE";
+    static final String CFG_CERTIFICATE_CN = "CERTIFICATE_CN";
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(ApplicationConfiguration.class);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Okay. Moved at the top.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfiguration.java
```diff
@@ -40,8 +40,41 @@
...
+    public static final Integer DEFAULT_AUDIT_LOG_CLIENT_KEEP_ALIVE = 60;
+    public static final Integer DEFAULT_FLOWABLE_JOB_EXECUTOR_CORE_THREADS = 8;
+    public static final Integer DEFAULT_FLOWABLE_JOB_EXECUTOR_MAX_THREADS = 32;
+    public static final Integer DEFAULT_FLOWABLE_JOB_EXECUTOR_QUEUE_CAPACITY = 16;
+    public static final Integer DEFAULT_FSS_CACHE_UPDATE_TIMEOUT_MINUTES = 30;
+    public static final Integer DEFAULT_SPACE_DEVELOPER_CACHE_TIME_IN_SECONDS = 20;
+    public static final int DEFAULT_CONTROLLER_CLIENT_CONNECTION_POOL_SIZE = 75;
+    public static final int DEFAULT_CONTROLLER_CLIENT_THREAD_POOL_SIZE = 75;
+    public static final String CF_INSTANCE_INDEX = "CF_INSTANCE_INDEX";
+    public static final String DEFAULT_CERTIFICATE_CN = "SAP SE";
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Removed.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfiguration.java
```diff
@@ -40,8 +40,41 @@
...
+    public static final Integer DEFAULT_AUDIT_LOG_CLIENT_QUEUE_CAPACITY = 8;
+    public static final Integer DEFAULT_AUDIT_LOG_CLIENT_KEEP_ALIVE = 60;
+    public static final Integer DEFAULT_FLOWABLE_JOB_EXECUTOR_CORE_THREADS = 8;
+    public static final Integer DEFAULT_FLOWABLE_JOB_EXECUTOR_MAX_THREADS = 32;
+    public static final Integer DEFAULT_FLOWABLE_JOB_EXECUTOR_QUEUE_CAPACITY = 16;
+    public static final Integer DEFAULT_FSS_CACHE_UPDATE_TIMEOUT_MINUTES = 30;
+    public static final Integer DEFAULT_SPACE_DEVELOPER_CACHE_TIME_IN_SECONDS = 20;
+    public static final int DEFAULT_CONTROLLER_CLIENT_CONNECTION_POOL_SIZE = 75;
+    public static final int DEFAULT_CONTROLLER_CLIENT_THREAD_POOL_SIZE = 75;
+    public static final String CF_INSTANCE_INDEX = "CF_INSTANCE_INDEX";
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Actually yes, moved.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfigurationTest.java
```diff
@@ -77,52 +78,428 @@ public void testGetControllerUrlWithInvalidValue() {
...
+                                       .isEmpty());
+        Assertions.assertFalse(platform.getModuleTypes()
+                                       .isEmpty());
+        Assertions.assertTrue(platform.getParameters()
+                                      .isEmpty());
+    }
+
+    @Test
+    public void testGetPlatformNoPlatformInEnvironment() {
+        IllegalStateException illegalStateException = assertThrows(IllegalStateException.class, () -> configuration.getPlatform());
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Okay, done.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfigurationTest.java
```diff
@@ -17,28 +19,27 @@
...
 
-    private static final String VCAP_APPLICATION_JSON_WITHOUT_SPACE_ID = "{\"instance_id\":\"d827d1f2-d4c1-462a-ab9d-5bf828706d63\","
-        + "\"application_name\":\"deploy-service\"," + "\"started_at_timestamp\":\"1437740774167\","
-        + "\"state_timestamp\":\"1437651427557\"," + "\"application_uris\":[\"sofd60245639a:12993\"],"
-        + "\"uris\":[\"sofd60245639a:12993\"]}";
-    private static final String VCAP_APPLICATION_JSON = "{\"instance_id\":\"d827d1f2-d4c1-462a-ab9d-5bf828706d63\","
-        + "\"application_name\":\"deploy-service\"," + "\"space_id\":\"954229f5-4945-43eb-8acb-a8f07cc5a7f8\","
-        + "\"started_at_timestamp\":\"1437740774167\"," + "\"state_timestamp\":\"1437651427557\","
-        + "\"application_uris\":[\"sofd60245639a:12993\"]," + "\"uris\":[\"sofd60245639a:12993\"]}";
+    private static final String VCAP_APPLICATION_JSON_WITHOUT_SPACE_ID_FILE = "vcap-application-without-space.json";
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/ApplicationConfigurationTest.java
```diff
@@ -17,28 +19,27 @@
...
 import com.sap.cloud.lm.sl.cf.core.auditlogging.AuditLoggingFacade;
 import com.sap.cloud.lm.sl.cf.core.configuration.Environment;
+import com.sap.cloud.lm.sl.cf.core.health.model.HealthCheckConfiguration;
 import com.sap.cloud.lm.sl.cf.core.message.Messages;
 import com.sap.cloud.lm.sl.common.util.JsonUtil;
 import com.sap.cloud.lm.sl.common.util.MapUtil;
+import com.sap.cloud.lm.sl.common.util.TestUtil;
+import com.sap.cloud.lm.sl.mta.model.Platform;
 
 public class ApplicationConfigurationTest {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Thanks.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
```diff
@@ -50,6 +50,8 @@
     public static final String PARAM_SKIP_OWNERSHIP_VALIDATION = "skipOwnershipValidation";
     public static final String PARAM_MODULES_FOR_DEPLOYMENT = "modulesForDeployment";
     public static final String PARAM_RESOURCES_FOR_DEPLOYMENT = "resourcesForDeployment";
+    public static final String PARAM_VERIFY_ARCHIVE_SIGNATURE = "verifyArchiveSignature";
+    public static final String PARAM_CERTIFICATE_CN = "certificateCN";
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done. Removed.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/metadata/BlueGreenDeployMetadata.java
```diff
@@ -126,6 +126,16 @@
...
                                     .build());
+        PARAMS.add(ParameterMetadata.builder()
+                                    .id(Constants.PARAM_VERIFY_ARCHIVE_SIGNATURE)
+                                    .type(ParameterType.BOOLEAN)
+                                    .defaultValue(false)
+                                    .build());
+        PARAMS.add(ParameterMetadata.builder()
+                                    .id(Constants.PARAM_CERTIFICATE_CN)
+                                    .type(ParameterType.STRING)
+                                    .defaultValue(null)
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Removed, also removed from DeployMetadata

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -905,6 +905,14 @@ public static boolean shouldDeleteServices(VariableScope scope) {
         return getBoolean(scope, Constants.PARAM_DELETE_SERVICES, false);
     }
 
+    public static boolean shouldVerifyArchiveSignature(VariableScope scope) {
+        return getBoolean(scope, Constants.PARAM_VERIFY_ARCHIVE_SIGNATURE, false);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Okay, the default value is removed

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -48,32 +58,70 @@ protected String getStepErrorMessage(DelegateExecution context) {
...
+            return fileEntry;
+        } catch (FileStorageException e) {
+            throw new SLException(e, "Failed to retrieve file with id \"{0}\"", fileId);
         }
     }
 
     private void validateExtDescriptorFileId(DelegateExecution context) {
-        String parameter = (String) context.getVariable(Constants.PARAM_EXT_DESCRIPTOR_FILE_ID);
-        if (parameter == null) {
+        String extensionDescriptorFileID = (String) context.getVariable(Constants.PARAM_EXT_DESCRIPTOR_FILE_ID);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Okay, done -> renamed to Id

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -92,48 +140,63 @@ private void validateDescriptorSize(FileEntry file) {
...
+                                  versionRuleString,
+                                  Constants.PARAM_VERSION_RULE,
+                                  Arrays.asList(VersionRule.values()));
         }
+    }
 
-        List<FileEntry> archivePartEntries = new ArrayList<>();
-        for (String appArchivePartId : appArchivePartsId) {
-            archivePartEntries.add(findFile(context, appArchivePartId));
+    private void verifyArchive(DelegateExecution context) {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -19,13 +25,17 @@
...
+import com.sap.cloud.lm.sl.cf.process.util.certficate_checker.MtaCertificateChecker;
 import com.sap.cloud.lm.sl.common.SLException;
 import com.sap.cloud.lm.sl.mta.model.VersionRule;
 
 @Component("validateDeployParametersStep")
 @Scope(BeanDefinition.SCOPE_PROTOTYPE)
 public class ValidateDeployParametersStep extends SyncFlowableStep {
 
+    @Inject
+    private MtaCertificateChecker mtaCertificateChecker;
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -92,48 +140,63 @@ private void validateDescriptorSize(FileEntry file) {
...
         try {
-            String space = StepsUtil.getSpaceId(context);
-            FileEntry fileEntry = fileService.getFile(space, fileId);
-            if (fileEntry == null) {
-                throw new SLException(Messages.ERROR_NO_FILE_ASSOCIATED_WITH_THE_SPECIFIED_FILE_ID_0_IN_SPACE_1, fileId, space);
-            }
-            return fileEntry;
-        } catch (FileStorageException e) {
-            throw new SLException(e, "Failed to retrieve file with id \"{0}\"", fileId);
+            return new URL("file:" + archiveFileName);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Yes, it would

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -92,48 +140,63 @@ private void validateDescriptorSize(FileEntry file) {
...
+        } catch (MalformedURLException e) {
+            throw new SLException(e);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Okay, Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -92,48 +140,63 @@ private void validateDescriptorSize(FileEntry file) {
...
-        if (startTimeout < 0) {
-            throw new SLException(Messages.ERROR_PARAMETER_1_MUST_NOT_BE_NEGATIVE, startTimeout, Constants.PARAM_START_TIMEOUT);
+        tryDeleteArchiveFile(archiveFileName);
+    }
+
+    private void tryDeleteArchiveFile(String archiveFileName) {
+        try {
+            Files.deleteIfExists(Paths.get(archiveFileName));
+        } catch (IOException e) {
+            logger.warn("Merged file not deleted");
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/certficate_checker/MtaCertificateChecker.java
```diff
@@ -0,0 +1,12 @@
+package com.sap.cloud.lm.sl.cf.process.util.certficate_checker;
+
+import java.net.URL;
+import java.security.cert.X509Certificate;
+import java.util.List;
+
+public interface MtaCertificateChecker {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Okay, done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        try {
+            JarFile jarFile = retrieveJarFile(jarURL);
+            List<JarEntry> jarEntries = getJarEntries(jarFile);
+            verifyJarEntries(targetCertificates, jarEntries, certificateCN);
+        } catch (IOException e) {
+            throw new ResourceAccessException(Messages.CERTIFICATE_VERIFICATION_HAS_FAILED, e);
+        }
+    }
+
+    private JarFile retrieveJarFile(URL jarURL) throws IOException {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        jarURL = new URL("jar:" + jarURL.toString() + "!/");
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Actually, it doesn't do the same but it has more restrictive behaviour. It will throw an exception if the path is absolute so it should ok to use it.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: This method was removed

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+        jarEntries.add(jarEntry);
+        readStream(jarFile, jarEntry);
+    }
+
+    private void validateJarEntryPath(JarFile jarFile, JarEntry jarEntry) throws IOException {
+        Path currentRelativePath = Paths.get("");
+        File jarEntryFile = new File(jarEntry.getName());
+        if (!jarEntryFile.getCanonicalPath()
+                         .startsWith(currentRelativePath.toAbsolutePath() + File.separator)) {
+            throw new IllegalAccessError(MessageFormat.format(Messages.ENTRY_IS_TRYING_TO_LEAVE_TARGET_DIR, jarFile.getName()));
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: This method was removed.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Every certificate type extends the certificate abstract class through X509Certificate so I think it's OK if we skip the validation

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,193 @@
...
+                continue;
+            }
+            verifyJarEntryIsSigned(targetCerts, entryCertificates, certificateCN);
+        }
+    }
+
+    private void checkJarEntryName(JarEntry jarEntry) {
+        if (!jarEntry.getName()
+                     .startsWith(META_INF)) {
+            throw new SecurityException(Messages.ARCHIVE_CONTAINS_UNSIGNED_FILES);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Converted to SLException

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/message/Messages.java
```diff
@@ -61,6 +61,11 @@
     public static final String CANNOT_DETERMINE_MODULE_NAME = "Cannot determine module name";
     public static final String CANNOT_DETERMINE_CURRENT_APPLICATION = "Cannot determine current application";
     public static final String INVALID_FILE_ENTRY_NAME = "Invalid file entry name!";
+    public static final String COULD_NOT_VERIFY_ARCHIVE_SIGNATURE = "Could not verify archive signature: {0}";
+    public static final String ARCHIVE_CONTAINS_UNSIGNED_FILES = "The archive contains unsigned class files.";
+    public static final String ARCHIVE_IS_NOT_SIGNED_BY_TRUSTED_SIGNER = "The archive is not signed by a trusted signer";
```
nictas: Change to `The achive is not signed by the trusted certificate authority \"{0}\"`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,49 @@
...
+    @Inject
+    private Verifier verifier;
+
+    public void checkCertificates(URL url, List<X509Certificate> certificates, String certificateCN) {
+        verifier.verify(certificates, certificateCN, url);
+    }
+
+    public List<X509Certificate> readProvidedCertificates(String certificatesFilename) {
+        Collection<? extends Certificate> certificates = null;
+        try (InputStream symantecInputStream = getClass().getResourceAsStream(certificatesFilename)) {
```
nictas: Rename `symantecInputStream` to `certificateInputStream`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(JarSignatureVerifier.class);
+
+    @Inject
+    private Verifier verifier;
+
+    public void checkCertificates(URL url, List<X509Certificate> certificates, String certificateCN) {
+        verifier.verify(certificates, certificateCN, url);
+    }
+
+    public List<X509Certificate> readProvidedCertificates(String certificatesFilename) {
```
nictas: Rename to `readCertificates`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,49 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import com.sap.cloud.lm.sl.cf.process.Constants;
+import com.sap.cloud.lm.sl.cf.process.util.verifier.Verifier;
+import com.sap.cloud.lm.sl.common.SLException;
+
+@Component
+public class JarSignatureVerifier {
```
nictas: Rename to `JarSignatureOperations`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/Verifier.java
```diff
@@ -0,0 +1,10 @@
+package com.sap.cloud.lm.sl.cf.process.util.verifier;
+
+import java.net.URL;
+import java.security.cert.X509Certificate;
+import java.util.List;
+
+public interface Verifier {
```
nictas: This interface should also be unnecessary - especially when considering that one of the arguments of the `verify` method is called `jarURL`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.ResourceAccessException;
+
+import com.google.common.base.Splitter;
+import com.sap.cloud.lm.sl.cf.core.util.FileUtils;
+import com.sap.cloud.lm.sl.cf.process.message.Messages;
+import com.sap.cloud.lm.sl.common.SLException;
+
+@Component
+public class JarVerifier implements Verifier {
```
nictas: Rename to `JarSignatureVerifier`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    private JarFile openJarFile(URL jarUrl) throws IOException {
+        jarUrl = toJarUrl(jarUrl);
+        JarURLConnection connection = (JarURLConnection) jarUrl.openConnection();
+        return connection.getJarFile();
+    }
+
+    private URL toJarUrl(URL jarUrl) throws MalformedURLException {
```
nictas: Rename the argument to `jarFileUrl`, because now it looks like you're converting an already JAR URL to JAR URL. :)

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+    private static final int BUFFER_SIZE = 8 * 1024;
+
+    @Override
+    public void verify(List<X509Certificate> targetCertificates, String certificateCN, URL jarURL) {
+        try {
+            JarFile jarFile = openJarFile(jarURL);
+            List<JarEntry> jarEntries = getJarEntries(jarFile);
+            verifyJarEntries(targetCertificates, jarEntries, certificateCN);
+        } catch (IOException e) {
+            throw new ResourceAccessException(MessageFormat.format(Messages.COULD_NOT_VERIFY_ARCHIVE_SIGNATURE, e.getMessage()), e);
```
nictas: `ResourceAccessException`s are supposed to be thrown for HTTP related operations.  Just add `IOException`s to the catch block below.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private URL toJarUrl(URL jarUrl) throws MalformedURLException {
+        return new URL("jar:" + jarUrl.toString() + "!/");
+    }
+
+    private List<JarEntry> getJarEntries(JarFile jarFile) throws IOException {
+        List<JarEntry> jarEntries = new ArrayList<>();
+        Enumeration<JarEntry> entries = jarFile.entries();
+        while (entries.hasMoreElements()) {
+            JarEntry jarEntry = readJarEntry(jarFile, entries);
```
nictas: Call `entries.nextElement()` here and pass it to the `readJarEntry` method. That way the `readJarEntry` method would be more focused on the actual reading and not on the collection from which the JAR entry comes from.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+                jarEntries.add(jarEntry);
+            }
+        }
+        return jarEntries;
+    }
+
+    private JarEntry readJarEntry(JarFile jarFile, Enumeration<JarEntry> entries) throws IOException {
+        JarEntry jarEntry = entries.nextElement();
+        FileUtils.validatePath(jarEntry.getName());
+        if (jarEntry.isDirectory()) {
```
nictas: This check is repeated here and in the method above. Find a way to skip it in one of the places.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    private List<X509Certificate> convertToX509Certificates(Certificate[] entryCertificates) {
+        return Arrays.stream(entryCertificates)
+                     .map(certificate -> (X509Certificate) certificate)
+                     .collect(Collectors.toList());
+    }
+
+    private boolean validCertificateChain(List<X509Certificate> targetCertificates, List<X509Certificate> entryCertificateChain,
```
nictas: Rename to `isValidCertificateChain` and `entryCertificateChain` to `candidateCertificateChain`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private void verifyJarEntryIsSigned(List<X509Certificate> targetCertificates, Certificate[] entryCertificates, String certificateCN) {
+        List<X509Certificate> entryCertificatesList = convertToX509Certificates(entryCertificates);
+        if (!validCertificateChain(targetCertificates, entryCertificatesList, certificateCN)) {
+            throw new SecurityException(Messages.ARCHIVE_IS_NOT_SIGNED_BY_TRUSTED_SIGNER);
+        }
+    }
+
+    private List<X509Certificate> convertToX509Certificates(Certificate[] entryCertificates) {
```
nictas: Rename `entryCertificates` to `certificates`. This method does not need to know that these certificates belong to a JAR entry. Also, rename to method to `toX509Certificates` for consistency with `toJarUrl`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+        if (!jarEntry.getName()
+                     .startsWith(META_INF)) {
+            throw new SLException(Messages.ARCHIVE_CONTAINS_UNSIGNED_FILES);
+        }
+    }
+
+    private void verifyJarEntryIsSigned(List<X509Certificate> targetCertificates, Certificate[] entryCertificates, String certificateCN) {
+        List<X509Certificate> entryCertificatesList = convertToX509Certificates(entryCertificates);
+        if (!validCertificateChain(targetCertificates, entryCertificatesList, certificateCN)) {
+            throw new SecurityException(Messages.ARCHIVE_IS_NOT_SIGNED_BY_TRUSTED_SIGNER);
```
nictas: The Javadoc of SecurityException does not match this usage.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+        return certificateCNMatches && targetCertificatesMatchEntryCertificate;
+    }
+
+    private void checkValidityOfCertificate(X509Certificate x509Certificate) {
+        try {
+            x509Certificate.checkValidity();
+        } catch (CertificateExpiredException | CertificateNotYetValidException e) {
+            throw new SecurityException(e);
```
nictas: The Javadoc of SecurityException does not match this usage.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    private boolean doTargetCertificatesMatchEntryCertificate(List<X509Certificate> targetCertificates, X509Certificate entryCertificate) {
+        return targetCertificates.stream()
+                                 .anyMatch(certificate -> Objects.equals(certificate, entryCertificate));
+    }
+
+    private boolean checkCertificateCN(X509Certificate entryCertificate, String certificateCN) {
```
nictas: Rename to `certificateCNMatches` and `entryCertificate` to `certificate`. Methods should know as little context as possible. In this case this method should not know that the certificate comes from a JAR entry.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
```diff
@@ -0,0 +1,162 @@
...
+        if (certificateCN == null) {
+            return true;
+        }
+        Map<String, String> params = Splitter.on(", ")
```
nictas: The whitespace is optional. We should handle that case.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/message/Messages.java
```diff
@@ -61,6 +61,11 @@
     public static final String CANNOT_DETERMINE_MODULE_NAME = "Cannot determine module name";
     public static final String CANNOT_DETERMINE_CURRENT_APPLICATION = "Cannot determine current application";
     public static final String INVALID_FILE_ENTRY_NAME = "Invalid file entry name!";
+    public static final String COULD_NOT_VERIFY_ARCHIVE_SIGNATURE = "Could not verify archive signature: {0}";
+    public static final String ARCHIVE_CONTAINS_UNSIGNED_FILES = "The archive contains unsigned class files.";
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: "class" is removed

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/message/Messages.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/Verifier.java
IvanBorislavovDimitrov: The interface is removed

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: It could be skipped here

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Used SLException instead of SecurityException

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/verifier/JarVerifier.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,175 @@
...
+        for (JarEntry jarEntry : jarEntries) {
+            Certificate[] entryCertificates = jarEntry.getCertificates();
+            if (entryCertificates == null || entryCertificates.length == 0) {
+                verifyIsMetaInformation(jarEntry, filesWithoutCertificates);
+                continue;
+            }
+            verifyJarEntryIsSigned(targetCertificates, entryCertificates, certificateCN);
+        }
+        if (!filesWithoutCertificates.isEmpty()) {
+            throw new SLException(MessageFormat.format(Messages.ARCHIVE_CONTAINS_UNSIGNED_FILES, filesWithoutCertificates));
```
nictas: Print the list of files only if the archive is signed partially. Use something of the sort:
```
List<JarEntry> nonMetaInformationJarEntries = jarEntries.stream()
                                                        .filter(jarEntry -> !isMetaInformation(jarEntry))

List<JarEntry> nonSignedJarEntries = jarEntries.stream()
                                               .filter(jarEntry -> { // Extract the body of this filter in a separate method.
                                                   Certificate[] certificates = jarEntry.getCertificates();
                                                   return certificates == null || certificates.length == 0;
                                                })

if (!nonSignedJarEntries.isEmpty()) {
    if (nonMetaInformationJarEntries.size() == nonSignedJarEntries.size() {
        throw new SLException("The archive is not signed");
    }
    throw new SLException("The archive contains unsigned files: {0}", String.join("\n", nonSignedJarEntries));
}
```

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -92,48 +140,59 @@ private void validateDescriptorSize(FileEntry file) {
...
 
-        StepsUtil.setAsJsonBinaries(context, Constants.VAR_FILE_ENTRIES, archivePartEntries);
-        getStepLogger().debug(Messages.BUILDING_ARCHIVE_FROM_PARTS);
-        new ExecutionRetrier().executeWithRetry(() -> new ArchiveMerger(fileService,
-                                                                        getStepLogger(),
-                                                                        context,
-                                                                        logger).createArchiveFromParts(archivePartEntries));
+    private void verifyArchiveSignature(String archiveFileName) {
+        String certificateCN = configuration.getCertificateCN();
+        getStepLogger().debug(Messages.NO_CUSTOM_CERTIFICATE_CN_NAME_USING_DEFAULT_0, certificateCN);
```
nictas: Change this message to "Will look for certificate CN \"{0}\".". The "using default" part is not relevant anymore, considering that the certificateCN should no longer be configurable through the CLI options.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -48,32 +58,70 @@ protected String getStepErrorMessage(DelegateExecution context) {
...
+    private List<FileEntry> getArchivePartEntries(DelegateExecution context, String[] appArchivePartsId) {
+        return Arrays.stream(appArchivePartsId)
+                     .map(appArchivePartId -> findFile(context, appArchivePartId))
+                     .collect(Collectors.toList());
+    }
+
+    private Runnable createArchiveFromParts(DelegateExecution context, List<FileEntry> archivePartEntries) {
+        return () -> {
+            String archiveFileName = new ArchiveMerger(fileService, getStepLogger(), context).createArchiveFromParts(archivePartEntries);
+            context.setVariable(Constants.VAR_ARCHIVE_FILE_NAME, archiveFileName);
```
nictas: Setting a Flowable variable in order to read it in the same step and not use it anywhere else is a code smell. Think of some other way to accomplish this.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/ValidateDeployParametersStep.java
```diff
@@ -92,48 +140,59 @@ private void validateDescriptorSize(FileEntry file) {
...
-                                                                        logger).createArchiveFromParts(archivePartEntries));
+    private void verifyArchiveSignature(String archiveFileName) {
+        String certificateCN = configuration.getCertificateCN();
+        getStepLogger().debug(Messages.NO_CUSTOM_CERTIFICATE_CN_NAME_USING_DEFAULT_0, certificateCN);
+        List<X509Certificate> certificates = jarSignatureOperations.readCertificates(Constants.SYMANTEC_CERTIFICATE_FILE);
+        jarSignatureOperations.checkCertificates(getArchiveFileNameURL(archiveFileName), certificates, certificateCN);
     }
 
-    private FileEntry findFile(DelegateExecution context, String fileId) {
+    private URL getArchiveFileNameURL(String archiveFileName) {
```
nictas: This is not actually the file's name but rather its path.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,175 @@
...
+
+    private static final String META_INF = "META-INF";
+    private static final String CN = "CN";
+    private static final int BUFFER_SIZE = 8 * 1024;
+
+    public void verify(List<X509Certificate> targetCertificates, String certificateCN, URL jarURL) {
+        try {
+            JarFile jarFile = openJarFile(jarURL);
+            List<JarEntry> jarEntries = getJarEntries(jarFile);
+            verifyJarEntries(targetCertificates, jarEntries, certificateCN);
```
nictas: The order of the arguments of this method is not intuitive. `targetCertificates` and `certificateCN` are both used as verification parameters (i.e. how things **should** be), while `jarEntries` are the things that must be verified. Change the signature so that the `jarEntries` come first.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,175 @@
...
+        return connection.getJarFile();
+    }
+
+    private URL toJarUrl(URL jarUrl) throws MalformedURLException {
+        return new URL("jar:" + jarUrl.toString() + "!/");
+    }
+
+    private List<JarEntry> getJarEntries(JarFile jarFile) throws IOException {
+        List<JarEntry> jarEntries = new ArrayList<>();
+        Enumeration<JarEntry> entries = jarFile.entries();
```
nictas: Would `EnumerationUtils.toList()` be helpful here (followed by `.stream()`)?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,175 @@
...
+        FileUtils.validatePath(jarEntry.getName());
+        verifySignature(jarFile, jarEntry);
+        return jarEntry;
+    }
+
+    private void verifySignature(JarFile jarFile, JarEntry jarEntry) throws IOException {
+        byte[] buffer = new byte[BUFFER_SIZE];
+        try (InputStream inputStream = jarFile.getInputStream(jarEntry)) {
+            while (inputStream.read(buffer, 0, buffer.length) != -1) {
+                // Just read. This will throw a SecurityException if a signature/digest check fails.
```
nictas: Add another comment saying that `JarEntry.getCertificates()` will return null if the stream isn't read first.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,175 @@
...
+                certificateCNMatches = certificateCNMatches(certificate, certificateCN);
+            }
+            if (!targetCertificatesMatchEntryCertificate) {
+                targetCertificatesMatchEntryCertificate = doTargetCertificatesMatchEntryCertificate(targetCertificates, certificate);
+            }
+        }
+        if (!certificateCNMatches) {
+            throw new SLException(MessageFormat.format(Messages.NONE_OF_YOUR_CERTIFICATES_CNS_MATCH, certificateCN));
+        }
+        return targetCertificatesMatchEntryCertificate;
```
nictas: It's inconsistent to throw an exception in case the CN doesn't match, and return false in case the certificates don't.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,175 @@
...
+        Pattern subjectSeparatorPattern = Pattern.compile(", *");
+        return Splitter.on(subjectSeparatorPattern)
+                       .withKeyValueSeparator("=")
+                       .split(certificate.getSubjectDN()
+                                         .getName());
+    }
+
+    private List<String> getTrustedCertificatesNames(List<X509Certificate> targetCertificates) {
+        return targetCertificates.stream()
+                                 .map(this::getCertificateSubjectParameters)
```
nictas: This can be extracted in a separate method `getCertificateCN`. You're using it in 2 places.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -0,0 +1,209 @@
...
+        for (X509Certificate certificate : candidateCertificateChain) {
+            checkValidityOfCertificate(certificate);
+            if (!certificateCNMatches) {
+                certificateCNMatches = certificateCNMatches(certificate, certificateCN);
+            }
+            if (!targetCertificatesMatchEntryCertificate) {
+                targetCertificatesMatchEntryCertificate = doTargetCertificatesMatchEntryCertificate(targetCertificates, certificate);
+            }
+        }
+        validateCertificateChain(certificateCNMatches, targetCertificatesMatchEntryCertificate, targetCertificates, certificateCN);
```
nictas: I preferred it when this method was inlined. Yes, the method was long, but the reason why we extract methods is to improve readability. This only works if each method handles a separate concept/operation that is understandable on its own. If you need to read one method in order to understand what another does, then there's something wrong with those methods.

Have you tested the performance, BTW? If this isn't slow, then the whole for loop could be split into two separate loops. Yes, it would be 2 times slower, but it would be much easier to read and undertand.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
nictas: I meant to have the following method:
```
    private String getCertificateCN(X509Certificate certificate) {
        Map<String, String> parameters = getCertificateSubjectParameters(parameters);
        return parameters.get(CN);
    }
```
This would reduce the duplication even further. :)

PR #620, opened 2019-08-20
Some refactoring

radito3:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/CloudControllerClientProvider.java
```diff
@@ -109,42 +100,17 @@ public CloudControllerClient getClientFromCache(String userName, String org, Str
...
     }
 
-    public CloudControllerClient getClientFromCache(String userName, String spaceId) {
+    private CloudControllerClient getClientFromCache(String userName, String spaceId) {
         // Get a client from the cache or create a new one if needed
         String key = getKey(userName, spaceId);
         return clients.computeIfAbsent(key, k -> clientFactory.createClient(getValidToken(userName), spaceId));
     }
 
-    /**
```
nictas: Where did this go?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/CloudControllerClientProvider.java
```diff
@@ -83,20 +83,11 @@ private OAuth2AccessToken getValidToken(String userName) {
...
-     * @param space the space associated with the client
-     * @return a CF client for the specified access token, organization, and space
-     */
-    public CloudControllerClient getClientFromCache(String userName, String org, String space) {
+    private CloudControllerClient getClientFromCache(String userName, String org, String space) {
         return getClientFromCache(userName, org, space, null);
     }
 
-    public CloudControllerClient getClientFromCache(String userName, String org, String space, String processId) {
+    private CloudControllerClient getClientFromCache(String userName, String org, String space, String processId) {
```
nictas: Why are these now private?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/CloudControllerClientProvider.java
```diff
@@ -83,20 +83,11 @@ private OAuth2AccessToken getValidToken(String userName) {
         return token;
     }
 
-    /**
```
nictas: Where did this go?

com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/Message.java
```diff
@@ -19,7 +19,7 @@ public Message id(Long id) {
         return this;
     }
 
-    @ApiModelProperty(value = "")
```
nictas: This is generated code. Avoid touching it.

com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/MessageType.java
```diff
@@ -1,39 +1,18 @@
...
 import javax.xml.bind.annotation.XmlEnum;
-import javax.xml.bind.annotation.XmlEnumValue;
 import javax.xml.bind.annotation.XmlType;
 
 @XmlType(name = "")
 @XmlEnum(MessageType.class)
 public enum MessageType {
 
-    @XmlEnumValue("INFO")
-    INFO(String.valueOf("INFO")), @XmlEnumValue("ERROR")
```
nictas: This is generated code. Avoid touching it. Furthermore, are you sure this will produce the same XML as before?

radito3: Yes, the docs for xml enum and enum value state that "if no xmlEnumValue annotation is present on the enum values in a class annotated with xmlEnum, the Enum.name() method will be used for the xml mapping".
And I see here that we are using the same strings as the names of the enum values. That is why I removed it

com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/Message.java
radito3: api model property has a default value for "value" which is this empty string.
That's why I removed it

PR #626, opened 2019-08-26
Change RecentLogsRetriever to call controller

radito3:
#### Description: 
Update RecentLogsRetriever to call controller client on getRecentLogs.
Add filtering to logs to fix issue with logs containing duplicate messages & messages being shuffled.
Delete Protobuf files & dependency.


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/clients/RecentLogsRetriever.java
```diff
@@ -1,164 +1,75 @@
...
-        super(restTemplateFactory);
-    }
-
     RecentLogsRetriever withErrorHandlerFactory(Supplier<CustomControllerClientErrorHandler> errorHandlerFactory) {
         this.errorHandlerFactory = errorHandlerFactory;
         return this;
     }
 
-    public List<ApplicationLog> getRecentLogsSafely(CloudControllerClient client, String appName) {
+    public List<ApplicationLog> getRecentLogsSafely(CloudControllerClient client, String appName, VariableScope context) {
```
nictas: This class should not know about Flowable. Don't use `VariableScope` here.

radito3: Where should the filtering be then?

nictas:
Merged as part of https://github.com/cloudfoundry-incubator/multiapps-controller/pull/697.

PR #646, opened 2019-09-19
Add rate limiting

radito3:
Add Resilience4J dependency


com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/RateLimiterFilter.java
```diff
@@ -0,0 +1,44 @@
...
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+            throws ServletException, IOException {
+        String ipAddress = request.getRemoteAddr();
+        RateLimiter rateLimiter = rateLimiterProvider.getRateLimiter(ipAddress);
+
+        boolean canProcessRequest = rateLimiter.acquirePermission();
+        AtomicRateLimiterMetrics metrics = (AtomicRateLimiterMetrics) rateLimiter.getMetrics();
+
+        if (canProcessRequest) {
+            response.setHeader("X-Rate-Limit-Remaining", Integer.toString(metrics.getAvailablePermissions()));
```
valentinEmpy: Can we be consistent with CF and return the same information they also do?
CloudFoundry returns the following information about rate limiting:
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 56
X-RateLimit-Reset: 1372700873

Take a look @ [Rate Limit info](https://docs.cloudfoundry.org/running/rate-limit-cloud-controller-api.html)

valentinEmpy: Also extract to constants

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/RateLimiterFilter.java
```diff
@@ -0,0 +1,44 @@
...
+            filterChain.doFilter(request, response);
+            return;
+        }
+        response.setHeader("X-Rate-Limit-Retry-After-Milliseconds",
```
valentinEmpy: Make it consistent with CF and remove this. Add the missing ones. Also extract to constants

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/RateLimiterFilter.java
```diff
@@ -0,0 +1,44 @@
...
+                Long.toString(TimeUnit.NANOSECONDS.toMillis(metrics.getNanosToWait())));
+        response.sendError(429, "Too many requests");
```
valentinEmpy: Make it exactly the same as CF "Too Many Requests". Extract to constant

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/RateLimiterFilter.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    protected boolean shouldNotFilter(HttpServletRequest request) {
+        return request.getHeader("Authorization") != null;
```
valentinEmpy: HttpHeaders.AUTHORIZATION

com.sap.cloud.lm.sl.cf.web/src/main/webapp/WEB-INF/security-context.xml
```diff
@@ -25,12 +25,15 @@
...
+
         <sec:access-denied-handler ref="accessDeniedHandler" />
-        <sec:custom-filter ref="csrfHeadersFilter" after="CSRF_FILTER" />
+
         <sec:custom-filter ref="correlationIdFilter" position="FIRST" />
+        <sec:custom-filter ref="resourceServerFilter" before="PRE_AUTH_FILTER" />
+        <sec:custom-filter ref="authFilter" after="BASIC_AUTH_FILTER" />
+        <sec:custom-filter ref="csrfHeadersFilter" after="CSRF_FILTER" />
         <sec:custom-filter ref="requestSizeFilter" after="PRE_AUTH_FILTER" />
+        <sec:custom-filter ref="rateLimitingFilter" position="LAST"/>
```
valentinEmpy: Will not the "authFilter" fail if there is no user authenticated? 
Meaning this filter will never be reached for non authenticated users and only for authenticated ones, also it will not do anything because since they are authenticated the check for Authorization header will always return true. None of the code will be executed.
Can you check this?
We were supposed to block non-authenticated users.

com.sap.cloud.lm.sl.cf.web/src/main/webapp/WEB-INF/security-context.xml
```diff
@@ -94,4 +97,6 @@
         <constructor-arg ref="dataSource" />
     </bean>
 
+    <bean id="rateLimitingFilter" class="com.sap.cloud.lm.sl.cf.web.security.RateLimiterFilter" />
```
valentinEmpy: Can we use Named annotation?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/RateLimiterFilter.java
```diff
@@ -0,0 +1,44 @@
...
+    @Inject
+    private RateLimiterProvider rateLimiterProvider;
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+            throws ServletException, IOException {
+        String ipAddress = request.getRemoteAddr();
+        RateLimiter rateLimiter = rateLimiterProvider.getRateLimiter(ipAddress);
+
+        boolean canProcessRequest = rateLimiter.acquirePermission();
```
enchobelezirev: What is this doing: acquirePermission?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/RateLimiterFilter.java
enchobelezirev: Can we use the HttpStatus enum?

com.sap.cloud.lm.sl.cf.web/pom.xml
```diff
@@ -165,6 +165,11 @@
             <groupId>org.flowable</groupId>
             <artifactId>flowable-spring-common</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.github.resilience4j</groupId>
+            <artifactId>resilience4j-ratelimiter</artifactId>
+            <version>0.17.0</version>
```
enchobelezirev: Why are we not using the latest 1.0.0 version?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/RateLimiterFilter.java
radito3: Tries to get a permit for the request, if no one is available returns false

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/RateLimiterFilter.java
```diff
@@ -0,0 +1,64 @@
...
+            return;
+        }
+        filterChain.doFilter(request, response);
+    }
+
+    private static long getUtcTimeForNextReset(RateLimiterConfig config, AtomicRateLimiterMetrics metrics) {
+        long cyclePeriodInNanos = config.getLimitRefreshPeriod()
+                                        .toNanos();
+        long currentCycle = metrics.getCycle();
+        long nanosToWaitForReset = (currentCycle + 1) * cyclePeriodInNanos;
```
enchobelezirev: Why incrementing the current cycle?
If I understood correctly, the whole calculation would do this:
1. Get the currentCycle (let's say 5th cycle)
2. nanosToWaitForReset = 6 * 1hour as nanos
3. now + nanosToWaitForReset = 6 cycles ahead
This means that value returned from this method will be the release after six cycles, from now.
Have I understood something wrong?

enchobelezirev: Is there some kind of a method which is saying how many nano seconds have been passed since the start of the current cycle?
There is a method in the AtomicRateLimiterMetrics:
https://github.com/resilience4j/resilience4j/blob/master/resilience4j-ratelimiter/src/main/java/io/github/resilience4j/ratelimiter/internal/AtomicRateLimiter.java#L413
Isn't that going to return the value which need to be waited before the next cycle?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/helpers/RateLimiterProvider.java
```diff
@@ -0,0 +1,28 @@
...
+        return rateLimiters.computeIfAbsent(ipAddress, k -> createRateLimiter());
+    }
+
+    private RateLimiter createRateLimiter() {
+        RateLimiterConfig config = RateLimiterConfig.custom()
+                                                    .limitForPeriod(5000)
+                                                    .limitRefreshPeriod(Duration.ofHours(1))
+                                                    .timeoutDuration(Duration.ZERO)
+                                                    .build();
+        return RateLimiter.of("MTA Rate Limiter", config);
```
enchobelezirev: Would it be nicer if we create the AtomicRateLimiter explicitly?
What would happen if they change the implementation of the "of" method to return SemaphoreBasedRateLimiter in future - the cast to the AtomicRateLimiterMetrics would be broken?

PR #658, opened 2019-09-27
Centralize authorization checks

nictas:


com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/security/UriAuthorizationFilter.java
```diff
@@ -0,0 +1,11 @@
+package com.sap.cloud.lm.sl.cf.web.security;
+
+import java.io.IOException;
+
+public interface UriAuthorizationFilter {
+
+    String getUriRegex();
+
+    void validateUserAuthorization(HttpCommunication communication) throws IOException;
```
enchobelezirev: Not sure about the naming - maybe validateUserIsAuthorized or just validateAuthorization?
I am not sure because the interface is defined really generically and specifying the word "User" seems not so generically any more to me ?

nictas: In the updated commit, I reused the naming from `AuthorizationChecker` - `ensureUserIsAuthorized`.

PR #673, opened 2019-10-17
Do not allow to overwrite read-only parameters

IvanBorislavovDimitrov:
LMCROSSITXSADEPLOY-1685

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
@@ -1,0 +1,61 @@
+package com.sap.cloud.lm.sl.cf.process.util;
+
+import static com.sap.cloud.lm.sl.cf.core.model.SupportedParameters.READ_ONLY_MODULE_PARAMETERS;
+import static com.sap.cloud.lm.sl.cf.core.model.SupportedParameters.READ_ONLY_RESOURCE_PARAMETERS;
+import static com.sap.cloud.lm.sl.cf.core.model.SupportedParameters.READ_ONLY_SYSTEM_PARAMETERS;
+
+import java.text.MessageFormat;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.inject.Named;
+
+import com.google.common.collect.Sets;
+import com.sap.cloud.lm.sl.cf.process.message.Messages;
+import com.sap.cloud.lm.sl.common.SLException;
+import com.sap.cloud.lm.sl.mta.model.DeploymentDescriptor;
+import com.sap.cloud.lm.sl.mta.model.NamedParametersContainer;
+
+@Named
+public class ReadOnlyParametersChecker {
+
+    public void check(DeploymentDescriptor descriptor) {
+        Map<String, Set<String>> detectedReadOnlyParameters = new LinkedHashMap<>();
+        checkForCommonParameters(new GlobalParameters(descriptor), READ_ONLY_SYSTEM_PARAMETERS, detectedReadOnlyParameters);
+        checkCollectionForCommonParameters(descriptor.getModules(), READ_ONLY_MODULE_PARAMETERS, detectedReadOnlyParameters);
+        checkCollectionForCommonParameters(descriptor.getResources(), READ_ONLY_RESOURCE_PARAMETERS, detectedReadOnlyParameters);
+        if (!detectedReadOnlyParameters.isEmpty()) {
+            throw new SLException(getFormattedOutput(detectedReadOnlyParameters));
+        }
+    }
+
+    private void checkForCommonParameters(NamedParametersContainer namedParametersContainer, Set<String> readOnlyParameters,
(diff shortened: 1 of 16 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.google.common.collect.Sets;
+import com.sap.cloud.lm.sl.cf.core.model.SupportedParameters;
+import com.sap.cloud.lm.sl.cf.process.message.Messages;
+import com.sap.cloud.lm.sl.common.SLException;
+import com.sap.cloud.lm.sl.mta.model.DeploymentDescriptor;
+import com.sap.cloud.lm.sl.mta.model.ParametersContainer;
+
+@Named
+public class ReadOnlyParametersChecker {
```
boyan-velinov: Let's say we have 2 modules and 2 resources, all with invalid parameters. With the current logic, the mta deploy will fail 4 times and every time the customer should adjust something new.
It would be better if they receives faster feedback.

Other idea for improvement is that parameter is defined which module/resource has define it. If you like that idea, it will exclude the other one, I guess. Because we have to know currently processed module or resource.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
```diff
@@ -0,0 +1,62 @@
...
+        Set<String> commonParameters = Sets.intersection(namedElementWithParametersContainer.getParameters()
+                                                                                            .keySet(),
+                                                         readOnlyParameters);
+        if (!commonParameters.isEmpty()) {
+            commonReadOnlyParameters.put(namedElementWithParametersContainer.getName(), commonParameters);
+        }
+    }
+
+    private void
+            checkCollectionForCommonParameters(List<? extends NamedElementWithParametersContainer> namedElementsWithParametersContainers,
```
boyan-velinov: This seems as a strange formatting?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
```diff
@@ -0,0 +1,62 @@
...
+import com.sap.cloud.lm.sl.cf.process.message.Messages;
+import com.sap.cloud.lm.sl.common.SLException;
+import com.sap.cloud.lm.sl.mta.model.DeploymentDescriptor;
+import com.sap.cloud.lm.sl.mta.model.NamedElementWithParametersContainer;
+
+@Named
+public class ReadOnlyParametersChecker {
+
+    public void check(DeploymentDescriptor descriptor) {
+        Map<String, Set<String>> commonReadOnlyParameters = new LinkedHashMap<>();
```
boyan-velinov: Rename the variable because it is not common any more

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersCheckerTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import com.sap.cloud.lm.sl.cf.process.message.Messages;
+import com.sap.cloud.lm.sl.common.SLException;
+import com.sap.cloud.lm.sl.mta.handlers.v3.DescriptorParser;
+import com.sap.cloud.lm.sl.mta.model.DeploymentDescriptor;
+
+public class ReadOnlyParametersCheckerTest {
```
boyan-velinov: I think it is ok to validate only for one read-only parameter per group.
Now it is harder for read.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
IvanBorislavovDimitrov: Formatted

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersCheckerTest.java
IvanBorislavovDimitrov: Done, only one parameter per group is tested

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
IvanBorislavovDimitrov: Done

PR #710, opened 2019-12-04
Sonar fixes

radito3:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/clients/RecentLogsRetriever.java
```diff
@@ -24,7 +24,12 @@
 
     public List<ApplicationLog> getRecentLogs(CloudControllerClient client, String appName, LogsOffset offset) {
         List<ApplicationLog> appLogs = client.getRecentLogs(appName);
-        return offset == null ? appLogs : ListUtils.select(appLogs, appLog -> isLogNew(appLog, offset));
+        if (offset == null) {
+            return appLogs;
+        }
+        return appLogs.stream()
```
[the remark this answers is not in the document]
radito3: Back when I first refactored it to use `ListUtils.select` I didn't know the performance and scalability difference between it and using the stream api. Now I understand that the `select` method scales linearly while streams scale better the larger the input data. And because application logs are usually quite a few, I decided to revert it to the way it was.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UpdateSubscribersStep.java
```diff
@@ -180,13 +179,7 @@ private CloudOrganization createDummyOrg(String orgName) {
...
 
     private List<CloudApplication> removeDuplicates(List<CloudApplication> applications) {
-        Map<UUID, CloudApplication> applicationsMap = new LinkedHashMap<>();
-        for (CloudApplication application : applications) {
-            applicationsMap.put(application.getMetadata()
-                                           .getGuid(),
-                                application);
-        }
-        return new ArrayList<>(applicationsMap.values());
+        return new ArrayList<>(new LinkedHashSet<>(applications));
```
ikasarov: This won't work - the above code assume we have different CloudApplication instances but with the same guid in the metadata. Fix this and add unit test if this wasn't caught.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UpdateSubscribersStep.java
```diff
@@ -288,11 +281,7 @@ private boolean updateCurrentEnvironment(Map<String, String> currentEnvironment,
     }
 
     private String getFirstComponent(String propertyName) {
-        int index = propertyName.indexOf(ValidatorUtil.DEFAULT_SEPARATOR);
-        if (index != -1) {
-            return propertyName.substring(0, index);
-        }
-        return propertyName;
+        return StringUtils.substringBefore(propertyName, ValidatorUtil.DEFAULT_SEPARATOR);
```
ikasarov: This again changes behaviour - the new code will return empty string instead of the whole propertyName. Add test.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -142,21 +138,13 @@ private String getJarEntriesNames(List<JarEntry> jarEntries) {
...
-        }
-        if (!certificateCNMatches) {
+        candidateCertificateChain.forEach(this::checkValidityOfCertificate);
+
+        List<String> certificateCNs = getCertificatesNames(candidateCertificateChain);
+        if (certificateCN != null && !certificateCNs.contains(certificateCN)) {
             throw new SLException(Messages.WILL_LOOK_FOR_CERTIFICATE_CN, certificateCN);
         }
-        if (!targetCertificatesMatchEntryCertificate) {
+        if (!targetCertificates.containsAll(candidateCertificateChain)) {
```
ikasarov: Old code checked for at least 1 match, this one checks for all matches. Add a test that breaks with this case + fix.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/JarSignatureVerifier.java
```diff
@@ -170,35 +158,33 @@ private void checkValidityOfCertificate(X509Certificate x509Certificate) {
...
+        private String country;
+
+        CertificateSubject(String subjectName) {
+            Matcher matcher = X509_CERT_SUBJECT_NAME.matcher(subjectName);
+            if (matcher.matches()) {
+                commonName = matcher.group(1);
+                organizationalUnit = matcher.group(2);
+                organization = matcher.group(3);
+                country = matcher.group(4);
+            }
```
ikasarov: Much better way to get the CN, great.

com.sap.cloud.lm.sl.cf.web/src/test/java/com/sap/cloud/lm/sl/cf/web/api/impl/MtaApiServiceImplTest.java
```diff
@@ -32,27 +34,27 @@
...
+    private List<Mta> mtas;
 
     private static final String USER_NAME = "someUser";
-    private static final String SPACE_GUID = "896e6be9-8217-4a1c-b938-09b30966157a";
+    private static final String SPACE_GUID = UUID.randomUUID()
+                                                 .toString();
 
-    @Before
-    public void initialize() throws Exception {
+    @BeforeEach
```
ikasarov: is the @BeforeEach change necessary?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/detect/DeployedComponentsDetector.java
```diff
@@ -32,54 +38,52 @@ public DeployedComponents detectAllDeployedComponents(Collection<CloudApplicatio
...
+                       .addAll(module.getServices());
         }
 
-        return createComponents(modulesMap, servicesMap, standaloneApps);
+        return new DeployedComponents(createDeployedMtas(modulesMap, servicesMap), standaloneApps);
+    }
+
+    private DeployedMtaModule createDeployedModule(CloudApplication app, ApplicationMtaMetadata appMetadata) {
+        String moduleName = ObjectUtils.defaultIfNull(appMetadata.getModuleName(), app.getName());
+        List<String> providedDependencies = ObjectUtils.defaultIfNull(appMetadata.getProvidedDependencyNames(), new ArrayList<>());
```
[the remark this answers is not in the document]
radito3: No, this does the same thing as before, only the ternary operator is in the `defaultIfNull` method, not in our code. I thought it makes things clearer.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/detect/DeployedComponentsDetector.java
```diff
@@ -89,43 +93,12 @@ private DeployedMta createDeployedMta(Entry<DeployedMtaMetadata, List<DeployedMt
...
-            services.addAll(mta.getServices());
-            modules.addAll(mta.getModules());
-        }
-        return new DeployedMta(new DeployedMtaMetadata(mtaId), modules, services);
+    private DeployedMta mergeMtas(DeployedMta mta1, DeployedMta mta2) {
+        List<DeployedMtaModule> modules = ListUtils.union(mta1.getModules(), mta2.getModules());
+        Set<String> services = SetUtils.union(mta1.getServices(), mta2.getServices());
+        String id = mta1.getMetadata()
+                        .getId();
+        return new DeployedMta(new DeployedMtaMetadata(id), modules, services);
```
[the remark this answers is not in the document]
radito3: What about this refactoring?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/persistence/dto/OperationDto.java
```diff
@@ -81,7 +81,7 @@ private OperationDto(String processId, String processType, Date startedAt, Date
 
     @Override
     public String getPrimaryKey() {
-        return processId;
+        return getProcessId();
```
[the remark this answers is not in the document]
radito3: The two changes are for different reasons.
This was thanks to a sonar issue, as it stated that this method and `getProcessId` did the same thing. So I thought that the "primary key" should get that "primary" property from the concrete object (as the `getPrimaryKey` method is from the interface)
The change in DeployedComponents was because there was no need to use a getter on a field from the same object as long as the getter doesn't do any extra processing.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UpdateSubscribersStep.java
radito3: Can more that one application (with same guid) ever exist in the same space? 
Wasn't the point of the guid of the app that it was unique?
Are more that one instances of the same app treated as separate `CloudApplication` objects?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UpdateSubscribersStep.java
radito3: No, it won't. This is the implementation of the substringBefore method 
`if (isEmpty(str) || separator == null) {
            return str;
        }
        if (separator.isEmpty()) {
            return EMPTY;
        }
        final int pos = str.indexOf(separator);
        if (pos == INDEX_NOT_FOUND) {
            return str;
        }
        return str.substring(0, pos);
`

com.sap.cloud.lm.sl.cf.web/src/test/java/com/sap/cloud/lm/sl/cf/web/api/impl/MtaApiServiceImplTest.java
radito3: This is just migrating from JUnit 4 to 5. It's not a change in behaviour.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/clients/ServiceUpdater.java
```diff
@@ -123,7 +123,7 @@ private String getCloudControllerUrl(CloudControllerClient client) {
     }
 
     private String getUpdateServiceUrl(String serviceUrl, String serviceGuid, String acceptsIncomplete) {
-        return serviceUrl + "/" + serviceGuid + acceptsIncomplete;
+        return serviceUrl + '/' + serviceGuid + acceptsIncomplete;
```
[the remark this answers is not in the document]
radito3: No, I just thought that it was better to use single quotations on a character, since it's not a string.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/SSLUtil.java
```diff
@@ -47,26 +42,4 @@ public static void disableSSLValidation() {
         }
     }
 
-    public static void useCaCertificateValidation(String certificatePath) {
```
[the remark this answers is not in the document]
radito3: Why? This is just removing dead code that isn't used anywhere.

com.sap.cloud.lm.sl.cf.web/src/test/java/com/sap/cloud/lm/sl/cf/web/api/impl/MtaApiServiceImplTest.java
ikasarov: oh, yes, you're right

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/detect/DeployedComponentsDetector.java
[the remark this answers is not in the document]
ikasarov: it's crucial that this merge case works the same way as before and this is a big change. Validate that it's sufficiently covered in the unit tests and maybe add an additional case or two

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/clients/ServiceUpdater.java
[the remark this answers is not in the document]
ikasarov: It doesn't really make much of a difference readability-wise but it forces a String conversion at runtime (https://docs.oracle.com/javase/specs/jls/se13/html/jls-15.html#jls-15.18.1)

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/UpdateSubscribersStep.java
ikasarov: discussed on other channel - please verify and revert

com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/Module.java
```diff
@@ -37,10 +37,9 @@
     @JsonProperty("updatedOn")
     Date getUpdatedOn();
 
-    // FIXME: This name should be changed in the API as it contains a typo.
     @ApiModelProperty
-    @JsonProperty("providedDendencyNames")
-    List<String> getProvidedDendencyNames();
+    @JsonProperty("providedDependencyNames")
```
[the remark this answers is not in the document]
radito3: Moved to https://github.com/cloudfoundry-incubator/multiapps-controller/pull/743

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/SSLUtil.java
```diff
@@ -32,7 +32,7 @@ public void checkServerTrusted(X509Certificate[] xcs, String string) {
 
         @Override
         public X509Certificate[] getAcceptedIssuers() {
-            return null;
```
nictas: Are you sure null and an empty array are equivalent in this context?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/analytics/collectors/AnalyticsCollector.java
```diff
@@ -70,7 +70,7 @@ public AnalyticsData collectAnalyticsData(DelegateExecution context) {
 
     }
 
-    public AbstractCommonProcessAttributesCollector<?> getProcessType(ProcessType processType) {
+    public AbstractCommonProcessAttributesCollector getProcessType(ProcessType processType) {
```
nictas: Isn't there a compiler warning now about usage of raw types?

com.sap.cloud.lm.sl.cf.shutdown.client/src/main/java/com/sap/cloud/lm/sl/cf/shutdown/client/configuration/EnvironmentBasedShutdownConfiguration.java
```diff
@@ -14,6 +14,8 @@
     private static final String CFG_USERNAME = "SHUTDOWN_USERNAME";
     private static final String CFG_PASSWORD = "SHUTDOWN_PASSWORD";
 
+    private static final String URL_IS_NOT_SPECIFIED_IN_APP_ENV = "{0} is not specified in the application's environment.";
```
nictas: This constant is not always used for URLs.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/util/SSLUtil.java
radito3: Yes, in the implementation of the SSLContext upon initiating the engine, when the trust manager is being chosen, there is a check for null and, if not, iterating through the array, etc. But in our case, both null and an empty array result in a dummy X509 trust manager. So, the functionality isn't being changed.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/analytics/collectors/AnalyticsCollector.java
radito3: I don't see one in my ide

com.sap.cloud.lm.sl.cf.shutdown.client/src/main/java/com/sap/cloud/lm/sl/cf/shutdown/client/configuration/EnvironmentBasedShutdownConfiguration.java
radito3: Changed to ARG_0_...

PR #718, opened 2020-01-07
Build tree to get all leaf nodes to abort process

boyan-velinov:
#### Description: 
When process is aborted, AbortProcessAction should wait
until all running subprocess have deadletter jobs.
So far AbortProcessAction do not wait and delete the process
hierarchy which lead to deadlock in DB.

Jira: LMCROSSITXSADEPLOY-1761

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
@@ -34,7 +34,8 @@ public class FlowableFacade {
     private static final Logger LOGGER = LoggerFactory.getLogger(FlowableFacade.class);
 
     private static final int DEFAULT_JOB_RETRIES = 0;
-    private static final int DEFAULT_ABORT_TIMEOUT_MS = 30000;
+    private static final int DEFAULT_ABORT_TIMEOUT_MS = 30 * 1000;
+    private static final int DEFAULT_ABORT_WAIT_TIMEOUT_MS = 2 * 60 * 1000;
 
     private final ProcessEngine processEngine;
 
@@ -111,6 +112,11 @@ public class FlowableFacade {
 
     public boolean hasDeadLetterJobs(String processId) {
         return !getDeadLetterJobs(processId).isEmpty();
+    }
+
+    private boolean allExecutionsHaveDeadLetterJobs(List<Execution> executions) {
+        return executions.stream()
+                         .allMatch(e -> !getDeadLetterJobsForExecution(e).isEmpty());
     }
 
     private List<Job> getDeadLetterJobs(String processId) {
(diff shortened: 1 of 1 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -113,6 +113,11 @@ public boolean hasDeadLetterJobs(String processId) {
         return !getDeadLetterJobs(processId).isEmpty();
     }
 
+    private boolean haveAllLeafsHaveDeadLetterJobs(List<Execution> executions) {
```
enchobelezirev: For me an you leaf is an understandable word but not for the rest of the team.
Maybe renaming it to child processes or subProcessLeafs would be better.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -268,6 +280,25 @@ public boolean isProcessInstanceAtReceiveTask(String processInstanceId) {
                                    .collect(Collectors.toList());
     }
 
+    private List<Execution> getAllLeafExecutions(String processInstanceId) {
+        List<Execution> allExecutions = getAllProcessExecutions(processInstanceId);
+        return allExecutions.stream()
```
enchobelezirev: This is a lot complexity: O(n^2).
Could we revert the things like this:
1. Find the execution which is related to the processInstanceId
2. Find each executions which have parentId = processInstaceId || superExecutionId = processInstanceId
3. Do 1 and 2, recursively for each execution foud in the step 2 
Would this work?
This way we would lower the complexity to O(nlogn)

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -230,6 +235,13 @@ public void deleteProcessInstance(String processInstanceId, String deleteReason)
                 // different if the process has parallel executions.
                 processEngine.getRuntimeService()
                              .setVariable(processInstanceId, Constants.PROCESS_ABORTED, Boolean.TRUE);
+                while (true) {
+                    List<Execution> allLeafExecutions = getAllLeafExecutions(processInstanceId);
+
+                    if (haveAllLeafsHaveDeadLetterJobs(allLeafExecutions)) {
+                        break;
+                    }
```
enchobelezirev: What if there is a bug in the code? We would stay in the loop forever.
Maybe giving it some time would be enough in order the executions to finish. For instance, 10 mins MAX or more?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
boyan-velinov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
boyan-velinov: Thanks for the idea!
Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
boyan-velinov: About the complexity I do agree.
However, I am not sure if it worths to have recursion here in order to optimize a little bit when the code will not look nicer.

PR #721, opened 2020-01-08
Fix objectstore for AWS GOV cloud

boyan-velinov:
DRAFT

Jira: LMCROSSITXSADEPLOY-1757


boyan-velinov:
Closing because it is just a draft for BLI that is not with high prio.
This could be fixed or partially fixed with commit b2799a6d22b6e515053a9c191060b78ac5e7dba9

PR #723, opened 2020-01-10
Abort processes in action required state immediately

boyan-velinov:
Jira: LMCROSSITXSADEPLOY-1761


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -229,13 +229,20 @@ public void deleteProcessInstance(String processInstanceId, String deleteReason)
         long overallAbortDeadline = System.currentTimeMillis() + DEFAULT_ABORT_WAIT_TIMEOUT_MS + DEFAULT_ABORT_TIMEOUT_MS;
         while (true) {
             try {
+                if (isProcessInstanceAtReceiveTask(processInstanceId)) {
```
enchobelezirev: Do it outside the loop.

boyan-velinov: It should remain in the outer loop, because of OptimisticLockingException mitigation

PR #739, opened 2020-01-27
Forward slash security fix

enchobelezirev:
#### Description: 
This commit provides additional path normalization before filter initialization. This is needed because when someone is trying to call the deploy-service using '//' in the request path, the deploy service matching mechanism does not detect it and thus the request is successful -> this brings security problems like accessing another user space


com.sap.cloud.lm.sl.cf.web/src/test/java/com/sap/cloud/lm/sl/cf/web/security/CompositeUriAuthorizationFilterTest.java
```diff
@@ -22,6 +22,7 @@
     private static final String FOO_REQUEST_URI = "/foo/qux";
     private static final String BAR_REQUEST_URI = "/bar/qux";
     private static final String BAZ_REQUEST_URI = "/baz/qux";
+    private static final String LAINA_URI = "/////bar////qux";
```
nictas: Lel

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/util/ServletUtil.java
```diff
@@ -15,6 +15,9 @@
 
 public final class ServletUtil {
 
+    private static final String SINGLE_FORWARD_SLASH = "/";
+    private static final String TWO_OR_MORE_DOUBLE_FRONT_SLASHES_REGEX = "\\/\\/+";
```
nictas: Can we rename this to `TWO_OR_MORE_FORWARD_SLASHES_REGEX`?

PR #743, opened 2020-02-03
Fix typo in Module model

radito3:
Backwards incompatible change.
To be merged with https://github.com/cloudfoundry-incubator/multiapps-cli-plugin/pull/92

### DIFF
--- a/com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/Module.java
+++ b/com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/Module.java
@@ -37,10 +37,9 @@ public interface Module {
     @JsonProperty("updatedOn")
     Date getUpdatedOn();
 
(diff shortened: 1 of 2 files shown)


com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/Module.java
```diff
@@ -37,10 +37,9 @@
     @JsonProperty("updatedOn")
     Date getUpdatedOn();
 
-    // FIXME: This name should be changed in the API as it contains a typo.
     @ApiModelProperty
-    @JsonProperty("providedDendencyNames")
-    List<String> getProvidedDendencyNames();
+    @JsonProperty("providedDependencyNames")
```
enchobelezirev: Is this also changed in the CLI ?

nictas:
Unfortunately, as much as I would like to merge this change, it's backwards-incompatible. It'll need to wait until the next version of the API.

PR #747, opened 2020-02-07
Fix abort deadlocks

IvanBorislavovDimitrov:
Sometimes when an operation gets aborted, deadlocks occur because flowable hasn't managed to update all dead letter jobs for all executions. When a process is deleted all of its subprocesses are also deleted as so as their dead letter jobs. This change provides a way to wait for flowable to create dead letter jobs for all aborted processes and afterward remove the processes and this way the deadlocks should not occur anymore.

LMCROSSITXSADEPLOY-1792

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
@@ -1,13 +1,34 @@
 package com.sap.cloud.lm.sl.cf.process.flowable;
 
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.ABORT_OPERATION_FAILED;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.ABORT_OPERATION_FOR_PROCESS_0_FAILED;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.CURRENT_EXECUTION_WITHOUT_CHILDREN_0;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.DEAD_LETTER_JOBS_FOR_PROCESS_0_1;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.EXECUTION_0_DOES_NOT_HAVE_DEAD_LETTER_JOBS;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.NEW_ACTIVE_EXECUTIONS_WITHOUT_CHILDREN_0_FOR_PROCESS_1;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.NO_DEAD_LETTER_JOBS_FOUND_FOR_PROCESS_WITH_ID_0;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.PARENT_EXECUTION_WILL_NOT_BE_WAITED_TO_FINISH;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.PROCESS_0_HAS_ALREADY_BEEN_SUSPENDED;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.PROCESS_0_HAS_BEEN_DELETED_SUCCESSFULLY;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.PROCESS_0_HAS_BEEN_SUSPENDED_SUCCESSFULLY;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.PROCESS_INSTANCE_0_IS_AT_RECEIVE_TASK;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.PROCESS_WITH_ID_0_NOT_FOUND_WHILE_WAITING_FOR_EXECUTION_1_TO_FINISH;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.TIMEOUT_OF_0_FOR_PROCESS_1_HAS_BEEN_REACHED;
+import static com.sap.cloud.lm.sl.cf.process.flowable.Messages.TIMER_EXECUTION_WILL_NOT_BE_WAITED_TO_FINISH;
 import static java.text.MessageFormat.format;
 
 import java.text.MessageFormat;
+import java.util.ArrayDeque;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Queue;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
@@ -17,6 +38,7 @@ import org.flowable.common.engine.api.FlowableOptimisticLockingException;
 import org.flowable.engine.ProcessEngine;
 import org.flowable.engine.history.HistoricActivityInstance;
 import org.flowable.engine.history.HistoricProcessInstance;
+import org.flowable.engine.impl.persistence.entity.ExecutionEntityImpl;
 import org.flowable.engine.runtime.Execution;
 import org.flowable.engine.runtime.ProcessInstance;
 import org.flowable.job.api.Job;
@@ -27,6 +49,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.sap.cloud.lm.sl.cf.persistence.Constants;
+import com.sap.cloud.lm.sl.common.NotFoundException;
 
 @Named
 public class FlowableFacade {
@@ -34,8 +57,7 @@ public class FlowableFacade {
     private static final Logger LOGGER = LoggerFactory.getLogger(FlowableFacade.class);
 
     private static final int DEFAULT_JOB_RETRIES = 0;
-    private static final int DEFAULT_ABORT_TIMEOUT_MS = 30 * 1000;
-    private static final int DEFAULT_ABORT_WAIT_TIMEOUT_MS = 1 * 60 * 1000;
+    private static final int DEFAULT_ABORT_WAIT_TIMEOUT_MS = 3 * 60 * 1000;
 
     private final ProcessEngine processEngine;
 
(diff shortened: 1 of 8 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -226,42 +241,128 @@ public void trigger(String executionId, Map<String, Object> variables) {
...
+    private void waitAllActiveExecutionsToFinish(String processInstanceId, List<Execution> allActiveExecutionsWithoutChildren) {
+        List<Execution> activeExecutionsWithoutChildren = allActiveExecutionsWithoutChildren;
+        long allSubprocessesFinishedDeadline = System.currentTimeMillis() + DEFAULT_ABORT_WAIT_TIMEOUT_MS;
+        while (true) {
+            List<Execution> newActiveExecutionsWithoutChildren = getNewExecutionsWithoutChildren(processInstanceId,
+                                                                                                 activeExecutionsWithoutChildren);
+            LOGGER.info(format(NEW_ACTIVE_EXECUTIONS_WITHOUT_CHILDREN_0_FOR_PROCESS_1, newActiveExecutionsWithoutChildren,
+                               processInstanceId));
+            activeExecutionsWithoutChildren = ListUtils.union(activeExecutionsWithoutChildren, newActiveExecutionsWithoutChildren);
+            activeExecutionsWithoutChildren = removeExecutionsIfTheyHaveChildren(processInstanceId, activeExecutionsWithoutChildren);
```
enchobelezirev: Do you need this method ? The above two methods getNewExecutionsWithoutChildren and getAllActiveExecutionsWithoutChildren assume that you are will not have executios which have children.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -226,42 +241,128 @@ public void trigger(String executionId, Map<String, Object> variables) {
...
+            activeExecutionsWithoutChildren = removeTimerExecutions(activeExecutionsWithoutChildren);
```
enchobelezirev: Why?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -226,42 +241,128 @@ public void trigger(String executionId, Map<String, Object> variables) {
...
+                               processInstanceId));
+            if (allExecutionsHaveDeadLetterJobs(activeExecutionsWithoutChildren)) {
+                LOGGER.info(format(ALL_EXECUTIONS_0_FOR_PROCESS_1_HAVE_BEEN_ABORTED, activeExecutionsWithoutChildren, processInstanceId));
+                return;
+            }
+            if (isPastDeadline(allSubprocessesFinishedDeadline)) {
+                LOGGER.info(format(TIMEOUT_OF_0_FOR_PROCESS_1_HAS_BEEN_REACHED, DEFAULT_ABORT_WAIT_TIMEOUT_MS, processInstanceId));
+                return;
             }
+            if (getProcessInstance(processInstanceId) == null) {
```
enchobelezirev: What is this check about?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/Messages.java
```diff
@@ -5,11 +5,28 @@
     private Messages() {
     }
 
+    // Info messages
+
+    public static final String NEW_ACTIVE_EXECUTIONS_WITHOUT_CHILDREN_0_FOR_PROCESS_1 = "New active executions without children: \"{0}\" for process: \"{1}\"";
```
enchobelezirev: Remove the ':'

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -226,42 +241,128 @@ public void trigger(String executionId, Map<String, Object> variables) {
...
+
+    private List<String> toExecutionIds(List<Execution> executions) {
+        return executions.stream()
+                         .map(Execution::getId)
+                         .collect(Collectors.toList());
+    }
+
+    private List<Execution> removeExecutionsIfTheyHaveChildren(String processInstanceId, List<Execution> currentExecutions) {
+        List<String> executionsWithoutChildrenIds = toExecutionIds(getAllExecutionsWithoutChildren(processInstanceId));
+        LOGGER.info(MessageFormat.format(CURRENT_EXECUTION_WITHOUT_CHILDREN_0, executionsWithoutChildrenIds));
```
enchobelezirev: This will spam a lot our log.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/Messages.java
```diff
@@ -5,11 +5,28 @@
...
+    public static final String ALL_EXECUTIONS_0_FOR_PROCESS_1_HAVE_BEEN_ABORTED = "All executions \"{0}\" for process: \"{1}\" have been aborted";
+    public static final String TIMEOUT_OF_0_FOR_PROCESS_1_HAS_BEEN_REACHED = "Timeout of \"{0}\" for process: \"{1}\" has been reached";
+    public static final String CURRENT_EXECUTION_WITHOUT_CHILDREN_0 = "Current execution without children: \"{0}\"";
+    public static final String NOT_LEAF_EXECUTION_0 = "Not a leaf execution: \"{0}\"";
```
enchobelezirev: Misleading and non-understandable log message.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/RetryProcessAction.java
```diff
@@ -30,20 +30,16 @@ public RetryProcessAction(FlowableFacade flowableFacade, List<AdditionalProcessA
 
     @Override
     protected void executeActualProcessAction(String user, String superProcessInstanceId) {
-        List<String> subProcessIds = getActiveExecutionIds(superProcessInstanceId);
-        ListIterator<String> subProcessesIdsIterator = subProcessIds.listIterator(subProcessIds.size());
-
+        List<String> inactiveExecutionIdsWithoutChildren = flowableFacade.getInactiveExecutionIdsWithoutChildren(superProcessInstanceId);
```
enchobelezirev: I think that the inactive executions will be removed from the execution table when some time passes. That is why all the active executions are taken. Is there any particular reason why this is changed?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/RetryProcessAction.java
```diff
@@ -30,20 +30,16 @@ public RetryProcessAction(FlowableFacade flowableFacade, List<AdditionalProcessA
...
+        LOGGER.info(MessageFormat.format("Inactive execution ids without children: \"{0}\"", inactiveExecutionIdsWithoutChildren));
         updateUserIfNecessary(user, superProcessInstanceId);
-        while (subProcessesIdsIterator.hasPrevious()) {
```
enchobelezirev: This was done in reverse order because if there are some serviceTasks failed inside a CallActivity, we need to retry first them, then the callActivity and after that the whole process. If you change this -> try to check whether you will not hit any OptimisticLockingException

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/OperationsHelper.java
```diff
@@ -107,6 +107,9 @@ public Operation addState(Operation operation) {
         if (flowableFacade.isProcessInstanceAtReceiveTask(processInstanceId)) {
             return Operation.State.ACTION_REQUIRED;
         }
+        if (flowableFacade.isProcessAborted(processInstanceId)) {
+            return Operation.State.ABORTED;
```
enchobelezirev: Hmm, what about the cli? Could this result in error on CLI side?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -226,42 +241,128 @@ public void trigger(String executionId, Map<String, Object> variables) {
...
-                LOGGER.warn(format(Messages.RETRYING_PROCESS_ABORT, processInstanceId));
+                throwExceptionInCaseOfTimeout(processInstanceId, overallAbortDeadline, e);
+            } catch (NotFoundException e) {
+                LOGGER.error(e.getMessage(), e);
+                return;
+            }
+        }
+    }
+
+    private void abortProcessSafely(String processInstanceId, String deleteReason) {
```
enchobelezirev: !!!!!Check this comment last!!!!!

Overall, the idea is the same which was firstly proposed. However, I do not like the approach here because you are making while cycle inside another while cycle -> this is a lot of complexity. What I would suggest is to to make the while cycle with condition: while(!executions.isEmpty()) ... and put in this executions list whenever you find new execution that need to be checked. If you want we could discuss this further..

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/OperationsHelper.java
IvanBorislavovDimitrov: I've tested it and I haven't seen unexpected behavior. But now, we retry the process even though it was aborted

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/RetryProcessAction.java
IvanBorislavovDimitrov: Reverted as it was before. The main idea was to take only executions.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
```diff
@@ -371,4 +516,7 @@ private String findHistoricProcessInstanceIdsAndProcessDefinitionKey(Set<String>
                             .getId();
     }
 
+    public static ExecutionEntityImpl toExecutionEntityImpl(Execution execution) {
+        return (ExecutionEntityImpl) execution;
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: ExecutionEntityImpl is the only descendant class I think it is safe enough

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/FlowableFacade.java
IvanBorislavovDimitrov: Done

PR #759, opened 2020-02-12
Do not retry when process is aborted

IvanBorislavovDimitrov:
Fixes:  LMCROSSITXSADEPLOY-1941

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/OperationsHelper.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/OperationsHelper.java
@@ -8,15 +8,12 @@ import javax.inject.Inject;
 import javax.inject.Named;
 
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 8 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/OperationsHelper.java
```diff
@@ -95,26 +94,25 @@ public Operation addState(Operation operation) {
...
+        return processHelper.findProcessInstanceById(processId)
+                            .isPresent() ? computeNonFinalState(processId) : computeFinalState(processId);
     }
 
-    private Operation.State computeNonFinalState(ProcessInstance processInstance) {
-        String processInstanceId = processInstance.getProcessInstanceId();
-        if (flowableFacade.isProcessInstanceAtReceiveTask(processInstanceId)) {
+    private Operation.State computeNonFinalState(String processInstanceId) {
+        if (processHelper.isAborted(processInstanceId)) {
+            return Operation.State.ABORTED;
```
nictas: "ABORTED" is a final state. This does not fit with the name of the method.

PR #774, opened 2020-02-24
ExecutionListeners for start and end of bg-deploy

radoslav-d:
Add ExecutionListeners for the start and end of testing phase of
blue-green deploy. Refactor ProcessLoggerProvider to create non-existing
files. JIRA: LMCROSSITXSADEPLOY-1870

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/AbstractProcessExecutionListener.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/AbstractProcessExecutionListener.java
@@ -40,21 +40,12 @@ public abstract class AbstractProcessExecutionListener implements ExecutionListe
     @Override
     public void notify(DelegateExecution context) {
         try {
-            this.stepLogger = createStepLogger(context);
             String correlationId = StepsUtil.getCorrelationId(context);
             if (correlationId == null) {
                 correlationId = context.getProcessInstanceId();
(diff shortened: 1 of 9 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/AbstractProcessExecutionListener.java
```diff
@@ -78,7 +69,7 @@ protected void finalizeLogs(DelegateExecution context) {
     }
 
     private String getCorrelationId(DelegateExecution context) {
-        return (String) context.getVariable(com.sap.cloud.lm.sl.cf.persistence.Constants.CORRELATION_ID);
+        return StepsUtil.getCorrelationId(context);
```
ikasarov: this is just a method wrapping another method with no additional value added

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/AbstractProcessExecutionListener.java
```diff
@@ -101,6 +92,12 @@ private StepLogger createStepLogger(DelegateExecution context) {
         return stepLoggerFactory.create(context, progressMessageService, processLoggerProvider, getLogger());
     }
 
+    protected boolean isRootProcess(DelegateExecution context) {
+        String correlationId = StepsUtil.getCorrelationId(context);
```
ikasarov: and you're calling it the old way here as well

PR #775, opened 2020-02-25
Filter mtas by space id as well as mta id

radito3:
When updating configuration subscriptions, mtas must be filtered by
space id as well, otherwise it would break mtas with same mta id in
different spaces.

LMCROSSITXSADEPLOY-1159

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/RemoveNewApplicationsSuffixStep.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/RemoveNewApplicationsSuffixStep.java
@@ -39,14 +39,16 @@ public class RemoveNewApplicationsSuffixStep extends SyncFlowableStep {
 
         String mtaId = (String) execution.getContext()
                                          .getVariable(Constants.PARAM_MTA_ID);
-        updateConfigurationSubscribers(appsToProcess, mtaId);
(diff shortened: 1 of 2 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/RemoveNewApplicationsSuffixStep.java
```diff
@@ -37,16 +37,16 @@ protected StepPhase executeStep(ExecutionWrapper execution) {
...
-        String mtaId = (String) execution.getContext()
-                                         .getVariable(Constants.PARAM_MTA_ID);
-        updateConfigurationSubscribers(appsToProcess, mtaId);
+        updateConfigurationSubscribers(appsToProcess, execution.getContext());
 
         return StepPhase.DONE;
     }
 
-    private void updateConfigurationSubscribers(List<String> appsToProcess, String mtaId) {
+    private void updateConfigurationSubscribers(List<String> appsToProcess, DelegateExecution context) {
```
nictas: I'd prefer passing the space GUID in addition to the MTA ID, instead of passing the context, since the context is just a map of variables. It's like passing everything to this method **and allowing it to change it**.

radito3: Done

PR #778, opened 2020-02-26
Skip rebind of services after testing phase

radito3:
LMCROSSITXSADEPLOY-1868

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
@@ -166,6 +166,7 @@ public class Constants {
     public static final String VAR_HOOK_FOR_EXECUTION = "hookForExecution";
     public static final String VAR_FILE_ENTRIES = "fileEntries";
     public static final String VAR_APPS_TO_RENAME = "appsToRename";
+    public static final String VAR_SHOULD_SKIP_SERVICE_REBINDING = "shouldSkipServiceRebinding";
 
     public static final String TOOL_TYPE = "tool_type";
     public static final String FEEDBACK_MAIL = "feedback_form";
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateAppStep.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateAppStep.java
@@ -245,6 +245,10 @@ public class CreateOrUpdateAppStep extends SyncFlowableStep {
 
         @Override
         public void handleApplicationServices() throws FileStorageException {
+            if (StepsUtil.getBoolean(execution.getContext(), Constants.VAR_SHOULD_SKIP_SERVICE_REBINDING, false)) {
+                return;
+            }
+
             List<String> services = app.getServices();
             boolean hasUnboundServices = unbindServicesIfNeeded(app, existingApp, client, services);
 
--- a/com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateStepWithExistingAppTest.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateStepWithExistingAppTest.java
@@ -3,25 +3,24 @@ package com.sap.cloud.lm.sl.cf.process.steps;
 import static org.mockito.ArgumentMatchers.eq;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.collections4.ListUtils;
 import org.cloudfoundry.client.lib.ApplicationServicesUpdateCallback;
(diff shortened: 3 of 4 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/EnteredTestingPhaseListener.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Named("enteredTestingPhaseListener")
+public class EnteredTestingPhaseListener extends AbstractProcessExecutionListener {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(EnteredTestingPhaseListener.class);
+
+    @Override
+    protected void notifyInternal(DelegateExecution context)  {
+        context.setVariable(Constants.VAR_ENTERED_TESTING_PHASE, true);
+    }
```
nictas: See `PrepareAppsRestartStep`. Can't we do something similar to what is done there?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
```diff
@@ -166,6 +166,7 @@
     public static final String VAR_HOOK_FOR_EXECUTION = "hookForExecution";
     public static final String VAR_FILE_ENTRIES = "fileEntries";
     public static final String VAR_APPS_TO_RENAME = "appsToRename";
+    public static final String VAR_PASSED_TESTING_PHASE = "passedTestingPhase";
```
nictas: The name of the variable is too specific. Why should the `CreateOrUpdateAppStep` know about the existence of a testing phase? What we want to tell the step via this variable is whether it should skip the bindings or not.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateAppStep.java
```diff
@@ -245,6 +251,14 @@ public void handleApplicationAttributes() {
 
         @Override
         public void handleApplicationServices() throws FileStorageException {
+            Boolean passedTestingPhase = (Boolean) execution.getContext()
+                                                            .getVariable(Constants.VAR_PASSED_TESTING_PHASE);
+            if (processTypeParser.getProcessType(execution.getContext())
+                                 .equals(ProcessType.BLUE_GREEN_DEPLOY) &&
+                BooleanUtils.toBooleanDefaultIfNull(passedTestingPhase, false)) {
```
nictas: Why not handle this in a getter in `StepsUtil`. There we already have a generic method that allows you to provide a default value.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateAppStep.java
```diff
@@ -245,6 +251,14 @@ public void handleApplicationAttributes() {
 
         @Override
         public void handleApplicationServices() throws FileStorageException {
+            Boolean passedTestingPhase = (Boolean) execution.getContext()
+                                                            .getVariable(Constants.VAR_PASSED_TESTING_PHASE);
+            if (processTypeParser.getProcessType(execution.getContext())
```
nictas: Does it matter whether the deployment is "blue-green" or not? If the variable is set to `true` we should skip the bindings.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateStepWithExistingAppTest.java
```diff
@@ -3,26 +3,26 @@
...
 import org.cloudfoundry.client.lib.domain.CloudApplication.State;
 import org.flowable.engine.delegate.DelegateExecution;
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.ExpectedException;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-import org.junit.runners.Parameterized.Parameters;
+import org.junit.jupiter.api.Assertions;
```
nictas: Are you absolutely sure this works? `SyncFlowableStepTest` has JUnit4 annotations, while here you're using JUnit5. Check whether this test is not lying about being successful both in the IDE and in the console (`mvn clean install`). I remember that with some versions of JUnit4 and JUnit5, the integration did not work well. The test would pass successfully, even when it should've failed.

radito3: It works

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Constants.java
radito3: Changed to "shouldSkipServiceRebinding"

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CreateOrUpdateAppStep.java
radito3: Changed to `StepsUtil.getBoolean(context, varName, defaultValue)` and removed check for blue-green deploy process id

PR #796, opened 2020-03-06
[Metadata] Move version and mta id to annotations

valentinEmpy:
LMCROSSITXSADEPLOY-1964


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/metadata/util/MtaMetadataUtil.java
```diff
@@ -44,6 +45,10 @@ public static boolean hasMtaMetadata(CloudEntity entity) {
                      .anyMatch(MTA_METADATA_LABELS::contains);
     }
 
+    public static String getHashedMtaId(String mtaId) {
+        return DigestUtils.md5DigestAsHex(mtaId.getBytes());
```
enchobelezirev: How do we guarantee hash size?

valentinEmpy: Guaranteed by the hashing algorithm itself. MD5 = 128 bits = 16 bytes = 32 hex digits

PR #799, opened 2020-03-09
Delete ownership validation

theghost5800:


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -99,129 +86,36 @@ protected String getStepErrorMessageAdditionalDescription(DelegateExecution cont
...
-                           .map(this::buildCloudServiceExtended)
-                           .collect(Collectors.toList());
-    }
-
-    private ImmutableCloudServiceExtended buildCloudServiceExtended(CloudService service) {
-        return ImmutableCloudServiceExtended.builder()
-                                            .metadata(service.getMetadata())
-                                            .name(service.getName())
-                                            .build();
+        return servicesToDelete.parallelStream()
```
VaskoBozhurski: This is not really properly parallel. It will have a maximum of 2 threads operating in parallel. You should use ForkJoinPoolUtil to ensure proper parallelism.

pom.xml
```diff
@@ -55,7 +55,7 @@
         <netty-codec-http.version>4.1.45.Final</netty-codec-http.version>
         <liquibase.version>3.4.1</liquibase.version>
         <liquibase-slf4j.version>1.2.1</liquibase-slf4j.version>
-        <cloudfoundry.client.version>1.21.1</cloudfoundry.client.version>
+        <cloudfoundry.client.version>1.22.0-SNAPSHOT</cloudfoundry.client.version>
```
VaskoBozhurski: Perhaps it would be better to first just release a new version of the cloudfoundry CLI so that we don't forget about this snapshot. :)

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceRemovalCalculator.java
```diff
@@ -0,0 +1,72 @@
...
+    @Inject
+    public ServiceRemovalCalculator(ServiceGetter serviceGetter) {
+        this.serviceGetter = serviceGetter;
+    }
+
+    public List<ServiceInstanceWithServiceKeys> calculateServicesToDelete(ExecutionWrapper execution,
+                                                                          List<CloudServiceInstance> serviceInstances) {
+        List<ServiceInstanceWithServiceKeys> serviceInstancesWithServiceKeys = getServiceInstancesWithServiceKeys(execution.getControllerClient(),
+                                                                                                                  serviceInstances);
+        if (StepsUtil.getServiceActionsToExecute(execution.getContext())
```
VaskoBozhurski: can you doublecheck if this is correct? The service action should be for 1 service, but the signature of the method is a list of services.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceRemoval.java
```diff
@@ -0,0 +1,155 @@
...
+        this.serviceRemovalCalculator = serviceRemovalCalculator;
+    }
+
+    public Map<String, ServiceOperation.Type> deleteServices(ExecutionWrapper execution, List<CloudServiceInstance> serviceInstancesData) {
+        Map<String, ServiceOperation.Type> triggeredServiceOperations = new HashMap<>();
+        CloudControllerClient client = execution.getControllerClient();
+
+        List<ServiceInstanceWithServiceKeys> servicesToDelete = determineServiceIntancesToDelete(execution, serviceInstancesData);
+        for (ServiceInstanceWithServiceKeys service : servicesToDelete) {
+            try {
```
VaskoBozhurski: maybe extract into a method and turn the for() into a stream?

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/clients/ServiceGetter.java
```diff
@@ -37,24 +36,19 @@ public ServiceGetter(@Qualifier("serviceInstanceGetter") AbstractServiceGetter s
         return serviceInstance;
     }
 
-    public List<ServiceInstanceWithServiceKeys> getServiceInstancesWithServiceKeys(CloudControllerClient client,
-                                                                                   List<CloudServiceInstance> services) {
-        return services.stream()
-                       .map(service -> buildServiceInstanceWithServiceKeys(service, client))
-                       .collect(Collectors.toList());
+    public ServiceInstanceWithServiceKeys getServiceInstancesWithServiceKeys(CloudControllerClient client, CloudServiceInstance service) {
+        return buildServiceInstanceWithServiceKeys(service, client);
```
VaskoBozhurski: If buildServiceInstanceWithServiceKeys(service, client) isn't called elsewhere you can just move the code to getServiceInstancesWithServiceKeys.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -345,13 +345,19 @@ static void setAppsToUndeploy(VariableScope scope, List<CloudApplication> apps)
...
     public static List<String> getServicesToDelete(VariableScope scope) {
-        TypeReference<List<String>> type = new TypeReference<List<String>>() {
-        };
-        return getFromJsonBinary(scope, Constants.VAR_SERVICES_TO_DELETE, type);
+        return getObject(scope, Constants.VAR_SERVICES_TO_DELETE, Collections.emptyList());
     }
 
     public static void setServicesToDelete(VariableScope scope, List<String> services) {
-        setAsJsonBinary(scope, Constants.VAR_SERVICES_TO_DELETE, services);
+        scope.setVariable(Constants.VAR_SERVICES_TO_DELETE, services);
```
VaskoBozhurski: this was a json binary before, are you sure the list will be properly serialized now with just setVariable in all cases?

com.sap.cloud.lm.sl.cf.process/src/main/resources/com/sap/cloud/lm/sl/cf/process/xs2-bg-deploy.bpmn
```diff
@@ -113,11 +113,12 @@
     </sequenceFlow>
     <serviceTask id="computeNextModulesToDeployInParallel" name="Compute Next Modules To Deploy" flowable:async="true" flowable:delegateExpression="${computeNextModulesStep}"></serviceTask>
     <exclusiveGateway id="hasParallelDeploymentCompletedGateway" name="Has Parallel Deployment Completed" default="deployAppsParallelNextIterationFlow"></exclusiveGateway>
-    <sequenceFlow id="flow151" sourceRef="computeNextModulesToDeployInParallel" targetRef="deployModulesParallel"></sequenceFlow>
```
VaskoBozhurski: Was this deleted intentionally?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
theghost5800: This change was necessary because Flowable engine throw exception that result from getServicesToDelete() method is not a Collection.

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/clients/ServiceGetter.java
```diff
@@ -22,19 +28,26 @@ public ServiceGetter(@Qualifier("serviceInstanceGetter") AbstractServiceGetter s
...
-
     public Map<String, Object> getServiceInstanceEntity(CloudControllerClient client, String serviceName, String spaceId) {
         Map<String, Object> serviceInstance = serviceInstanceGetter.getServiceInstanceEntity(client, serviceName, spaceId);
         if (serviceInstance == null || serviceInstance.isEmpty()) {
             serviceInstance = userProvidedServiceInstanceGetter.getServiceInstanceEntity(client, serviceName, spaceId);
         }
         return serviceInstance;
     }
+
+    public ServiceInstanceWithServiceKeys getServiceInstancesWithServiceKeys(CloudControllerClient client, CloudServiceInstance service) {
```
valentinEmpy: getServiceInstance(S), plural? Furthermore, read below about the ServiceInstanceWithServiceKeys class

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/ServiceInstanceWithServiceKeys.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.Collections;
+import java.util.List;
+
+import org.cloudfoundry.client.lib.domain.CloudServiceInstance;
+import org.cloudfoundry.client.lib.domain.CloudServiceKey;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface ServiceInstanceWithServiceKeys {
```
valentinEmpy: What is the benefit of having this class? As far as I can see, it is used only in delete services step, do you think it'd be simpler to just acquire the service and the service keys as separate entities and process them accordingly. What does this 'wrapping' help with? Wouldn't the code be more understandable and explicit in DeleteServicesStep if that was applied? What do you think

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/ServiceInstanceWithServiceKeys.java
```diff
@@ -0,0 +1,26 @@
...
+
+    CloudServiceInstance getServiceInstance();
+
+    @Value.Default
+    default List<CloudServiceKey> getServiceKeys() {
+        return Collections.emptyList();
```
valentinEmpy: Take a look at Immutables documentation. Quote:
"There's no need to use @Value.Default to return empty collections as collection attributes are empty by default if not initialized."

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Messages.java
```diff
@@ -180,6 +172,7 @@
     public static final String MISSING_SERVICE_OPERATION_STATE = "Stopping service operation polling of \"{0}\" due to missing operation state...";
     public static final String CANNOT_GET_CONTEXT_FOR_EVENT_0_AND_PROCESS_1 = "Could not get context for Flowable engine event (type: {0}, process ID: {1}).";
     public static final String SERVICE_0_IS_IN_STATE_1_AND_MAY_NOT_BE_OPERATIONAL = "Service \"{0}\" is in state \"{1}\" and may not be operational. Actions like update of credentials and binding may fail! Consider recreating it by specifying the --delete-services option.";
+    public static final String SERVICE_NOT_BE_DELETED_DUE_TO_SERVICE_BINDINGS_AND_SERVICE_KEYS = "Service \"{0}\" won''t be deleted due to existing service bindings and/or service keys";
```
valentinEmpy: What is the double ' for? I assume it is a type?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -99,129 +87,33 @@ protected String getStepErrorMessageAdditionalDescription(DelegateExecution cont
         return ExceptionMessageTailMapper.map(configuration, CloudComponents.SERVICE_BROKERS, offering);
     }
 
-    private List<CloudServiceExtended> getServicesData(List<String> serviceNames, ExecutionWrapper execution) {
-        CloudControllerClient client = execution.getControllerClient();
+    private CloudServiceInstance getServiceInstanceData(String serviceToDelete, CloudControllerClient client) {
```
valentinEmpy: Why does this method have the suffix 'Data'? Also, the variables assigned in the above methods also include the 'Data' suffix. Do we really need this method anyway, why not just call the client directly where this method is used?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
valentinEmpy: From the conversation, my understanding is that you'd want to use this variable in flowable bpmn diagram in order to parallelize the deletion of services, and you could not do that because you need the variable not to be in 'bytearray' format in the database. This is okay. However - is this backwards compatible? What would happen if I retry my deployment, that has this variable stored as 'bytearray' after the release of the diagram change happens - will the deployment fail?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceRemoval.java
```diff
@@ -0,0 +1,138 @@
...
+import com.sap.cloud.lm.sl.cf.core.util.ApplicationConfiguration;
+import com.sap.cloud.lm.sl.cf.process.Constants;
+import com.sap.cloud.lm.sl.cf.process.Messages;
+import com.sap.cloud.lm.sl.cf.process.steps.ExecutionWrapper;
+import com.sap.cloud.lm.sl.cf.process.steps.StepsUtil;
+import com.sap.cloud.lm.sl.cf.process.util.ExceptionMessageTailMapper.CloudComponents;
+import com.sap.cloud.lm.sl.common.SLException;
+
+@Named
+public class ServiceRemoval {
```
valentinEmpy: Can we rename this class? Reading it leaves me with the impression that it represents some 'operation' - the removal of the service. Perhaps - remover?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceRemoval.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private void prepareServicesToDelete(CloudControllerClient client, StepLogger stepLogger, ServiceInstanceWithServiceKeys service) {
+        unbindService(client, stepLogger, service.getServiceInstance());
+        deleteServiceKeys(client, stepLogger, service.getServiceKeys());
+
+    }
+
+    private void unbindService(CloudControllerClient client, StepLogger stepLogger, CloudServiceInstance serviceInstance) {
+        List<CloudServiceBinding> bindings = serviceInstance.getBindings();
+        if (bindings.isEmpty()) {
```
valentinEmpy: Why are those checks needed? You check for empty associations before you call deleteService. If there are no empty associations - would this ever be reached at all?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ServiceRemoval.java
```diff
@@ -0,0 +1,138 @@
...
+        return new CloudControllerException(e);
+
+    }
+
+    private void wrapAndThrowException(Exception e, CloudServiceInstance serviceInstance) {
+        String msg = buildNewExceptionMessage(e, serviceInstance.getService());
+        throw new SLException(e, msg);
+    }
+
+    private String buildNewExceptionMessage(Exception e, CloudService service) {
```
valentinEmpy: The method name is not very descriptive. Reading it, does not point at what exception message exactly it builds, especially the 'New' keyword. I had to check in the implementation to understand that it builds an error message wen deleting a service.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Messages.java
theghost5800: It was required before to escape single quote because if you don't escape it, MessageFormatter stop to resolve placeholders after single quote, now there is not others placeholders, so I will delete it.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServiceStep.java
```diff
@@ -0,0 +1,121 @@
...
+        return hasNoConflictWithServiceBindings(context, serviceBindings) && hasNoConflictWithServiceKeys(context, serviceKeys);
+    }
+
+    private boolean hasNoConflictWithServiceBindings(DelegateExecution context, List<CloudServiceBinding> serviceBindings) {
+        return serviceBindings.isEmpty() || StepsUtil.getServiceActionsToExecute(context)
+                                                     .contains(ServiceAction.RECREATE);
+    }
+
+    private boolean hasNoConflictWithServiceKeys(DelegateExecution context, List<CloudServiceKey> serviceKeys) {
+        return serviceKeys.isEmpty() || StepsUtil.shouldDeleteServiceKeys(context);
```
valentinEmpy: Same as the above comment. What do you think?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServiceStep.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private void deleteService(ExecutionWrapper execution, CloudServiceInstance serviceInstance, List<CloudServiceKey> serviceKeys) {
```
valentinEmpy: Why not just get rid of this method? It is a one-liner anyway.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -1,124 +1,232 @@
...
+        CloudControllerClient client = execution.getControllerClient();
+
+        List<String> servicesToDelete = new ArrayList<>(StepsUtil.getServicesToDelete(execution.getContext()));
+
+        if (servicesToDelete.isEmpty()) {
+            getStepLogger().debug(Messages.MISSING_SERVICES_TO_DELETE);
+            return StepPhase.DONE;
+        }
+
+        List<CloudServiceExtended> servicesData = getServicesData(servicesToDelete, execution);
```
valentinEmpy: As mentioned earlier - I do not understand the suffix 'Data' here. The method just retrieves Service Instances so 'getServices' seems sufficient. What purpose does the 'Data' suffix serve here?
This includes variables and methods. Below is another variable and method containing the 'Data' suffix again.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -1,124 +1,232 @@
...
+        return new CloudControllerException(e);
+
+    }
+
+    private void wrapAndThrowException(Exception e, CloudServiceInstance serviceInstance, String serviceName) {
+        String msg = buildNewExceptionMessage(e, serviceInstance, serviceName);
+        throw new SLException(e, msg);
+    }
+
+    private String buildNewExceptionMessage(Exception e, CloudServiceInstance serviceInstance, String serviceName) {
```
valentinEmpy: The old comment still stands.
The method name is not very descriptive. Reading it, does not point at what exception message exactly it builds, especially the 'New' keyword. I had to check in the implementation to understand that it builds an error message wen deleting a service.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
```diff
@@ -1,124 +1,232 @@
...
+        if (serviceInstance == null) {
+            return MessageFormat.format(Messages.ERROR_DELETING_SERVICE_SHORT, serviceName, e.getMessage());
+        }
+        CloudService service = serviceInstance.getService();
+        return MessageFormat.format(Messages.ERROR_DELETING_SERVICE, service.getName(), service.getLabel(), service.getPlan(),
+                                    e.getMessage());
+    }
+
+    private void logBindings(List<CloudServiceBinding> bindings) {
```
valentinEmpy: Why not just get rid of this method? It is a one-liner anyway.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteServicesStep.java
theghost5800: This class was returned as it is due backward compatability. This class will be removed after the release of these changes. I can rename this method and variable and add comment at the begging that this class will be removed.

valentinEmpy: Perhaps put the @Deprecated annotation as well and explain why it will be removed

PR #808, opened 2020-03-17
Refactor ServiceCreator to call cf client

radoslav-d:
Replace the programmatic filtering of services and the plain http rest
call for creating a service with a call to the CloudControllerClient.
This will remove duplicated logic and code, which already lives in
cf-java-client. 

**JIRA:** LMCROSSITXSADEPLOY-1516


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/cf/clients/ServiceCreator.java
```diff
@@ -36,36 +29,15 @@ ServiceCreator withErrorHandlerSupplier(Supplier<CustomControllerClientErrorHand
...
-        Map<String, Object> serviceRequest = new HashMap<>();
-        serviceRequest.put(SPACE_GUID, spaceId);
-        serviceRequest.put(SERVICE_NAME, service.getName());
-        serviceRequest.put(SERVICE_PLAN_GUID, cloudServicePlan.getMetadata()
-                                                              .getGuid()
-                                                              .toString());
-        serviceRequest.put(SERVICE_PARAMETERS, service.getCredentials());
-        serviceRequest.put(SERVICE_TAGS, service.getTags());
-        return serviceRequest;
+        client.createService(service);
```
enchobelezirev: Please make sure that this request covers all the Request parameters + calls the correct URL(accepts_incomplete query parameter).

enchobelezirev: If you are sure that this call is enough and is doing all the necessarry putting of request parameters, remove the class and replace every usage of ServiceCreator with `client.createService(service);`

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/cf/clients/CloudServiceOperatorTest.java
```diff
@@ -49,15 +46,13 @@ private void prepareRestTemplateFactory() {
 
     private void prepareClient() throws IOException {
         URL controllerUrl = new URL(CONTROLLER_URL);
+        Mockito.reset(client);
```
enchobelezirev: Why reset? Isn't the new Mocked object be enough here?
If no, please refactor the test in order to be.
Further reading: https://softwareengineering.stackexchange.com/questions/188299/is-this-an-appropriate-use-of-mockitos-reset-method

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/cf/clients/ServiceWithAlternativesCreatorTest.java
```diff
@@ -48,62 +38,44 @@
                 Arguments.of("service-03.json", "Service label must not be null", IllegalArgumentException.class),
                 Arguments.of("service-04.json", "Service name must not be null", IllegalArgumentException.class),
                 Arguments.of("service-05.json", "Service plan must not be null", IllegalArgumentException.class),
-                Arguments.of("service-06.json", "Could not create service instance \"foo\". Service plan \"v3.4-extra-large\" from service offering \"mongodb\" was not found.", CloudOperationException.class),
-                Arguments.of("service-07.json", "Could not create service instance \"foo\". Service plan \"v2.0-sp3\" from service offering \"hana\" was not found.", CloudOperationException.class),
+                Arguments.of("service-06.json", "401 Unauthorized", CloudOperationException.class),
```
enchobelezirev: What about the detailed message? Is it gone and how the 401 Unauthorized is replacing it?

com.sap.cloud.lm.sl.cf.core/src/test/resources/com/sap/cloud/lm/sl/cf/core/cf/clients/service-01.json
```diff
@@ -1,5 +1,5 @@
 {
-  "service": {
+  "initialService": {
```
enchobelezirev: In testing the appropiate words for inputs are "actual" and "expected".
So, maybe you culd use actualService ?

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/cf/clients/ServiceWithAlternativesCreatorTest.java
radoslav-d: It is a totally different test, in fact test case 16 and 17 are in place 06 and 07. Test cases 06 and 07 are removed because they are not applicable any more. The logic, which they tested is in CloudControllerClient implementation.

enchobelezirev: Okay

PR #815, opened 2020-03-20
Check if route is used by service before deletion

radoslav-d:
JIRA: LMCROSSITXSADEPLOY-1452


radoslav-d:
Build will fail until release of CF Java client 1.21.7

PR #816, opened 2020-03-20
Execute AbortProcessAction when abortOnError

radoslav-d:
As CTS deploy process relies on that sub-processes will not abort/delete
the root process, add new parameter NO_ROOT_ABORT_PROCESS_ABORT, which
will be TRUE as default for CTS deploy. When considering, what should
happen for a process, which has failed, check if this flag is set. Abort
root process if the flag is not set or false and delete failed root
process if set to true. JIRA: LMCROSSITXSADEPLOY-1793


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/AbortFailedProcessCommandFactory.java
```diff
@@ -66,14 +67,26 @@ private String findProcessInstanceId(CommandContext commandContext) {
...
-                return;
-            }
-            Boolean shouldAbortOnError = HistoryUtil.getVariableValue(commandContext, processInstanceId, Constants.PARAM_ABORT_ON_ERROR);
-            if (BooleanUtils.toBoolean(shouldAbortOnError)) {
+            if (shouldAbort(commandContext, processInstanceId)) {
                 abortProcess(commandContext, processInstanceId);
             }
+            String rootProcessInstanceId = getRootProcessInstanceId(commandContext, processInstanceId);
+            if (rootProcessInstanceId != null && shouldAbort(commandContext, rootProcessInstanceId)) {
+                abortProcess(commandContext, rootProcessInstanceId);
```
enchobelezirev: I would suggest the following workflow -
1. Check in the current process that there is a variable PARAM_ABORT_ON_ERROR - this is done.
2. If there is -> get the correlation id
3. abort the process with the correlation id.

Benefits:
- delete the whole process instance(by the correlationId) recursively(do not rely on that logic - it could be broken). 
- No OptimisticLockingException could rise because if we are pass here we know that there is no running executions

Drawbacks:
- if there are parallel processing and we should abort - there could be an OptimisticLockingException. Check what is done in the AbortProcessAction and decide what could be done here.

enchobelezirev: Current implementation drawbacks:
- OptimisticLocking would be risen if you have parallel exectuins because of this: `if (rootProcessInstanceId != null && shouldAbort(commandContext, rootProcessInstanceId)) {
                 abortProcess(commandContext, rootProcessInstanceId);`
- Some lefover processes will be presented because the you are not making sure that all the child executions will be deleted.
- Other elements, part of the process(jobs, executions, timers, variables, etc.), are not again guaranteed to be removed as a whole because the deletion will not happen recursively from the root process element.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/AbortFailedProcessCommandFactory.java
```diff
@@ -2,42 +2,48 @@
...
 
 import com.sap.cloud.lm.sl.cf.process.util.HistoryUtil;
-import com.sap.cloud.lm.sl.cf.web.api.model.Operation;
+import org.springframework.context.annotation.Lazy;
 
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@Named
+@Lazy
```
enchobelezirev: Why @ Lazy ? It is initialized at the startup of the deploy-service.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/AbortFailedProcessCommandFactory.java
```diff
@@ -50,43 +56,34 @@ public Object execute(CommandContext commandContext) {
...
 
         private void abortProcessIfRequested(CommandContext commandContext, String processInstanceId) {
             String correlationId = HistoryUtil.getVariableValue(commandContext, processInstanceId, Constants.VAR_CORRELATION_ID);
-            if (!processInstanceId.equals(correlationId)) {
-                return;
-            }
-            Boolean shouldAbortOnError = HistoryUtil.getVariableValue(commandContext, processInstanceId, Constants.PARAM_ABORT_ON_ERROR);
-            if (BooleanUtils.toBoolean(shouldAbortOnError)) {
-                abortProcess(commandContext, processInstanceId);
+            if (shouldAbort(commandContext, processInstanceId) || shouldAbort(commandContext, correlationId)) {
```
enchobelezirev: Why the two checks?
If you end up here, the process with the correlationId(the super parent) will have the variable (ABORT_ON_ERROR). So, isn't only the second check enough?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/AbortFailedProcessCommandFactory.java
radoslav-d: Because there are circular dependencies when the component is initialiazed eagerly

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/commands/CommandExecutor.java
```diff
@@ -0,0 +1,7 @@
+package com.sap.cloud.lm.sl.cf.process.flowable.commands;
+
+
+public interface CommandExecutor {
```
enchobelezirev, line 4: Let's rename this to FlowableCommandExecutor

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/commands/CommandExecutorFactory.java
```diff
@@ -0,0 +1,8 @@
+package com.sap.cloud.lm.sl.cf.process.flowable.commands;
+
+import org.flowable.common.engine.impl.interceptor.CommandContext;
+
+public interface CommandExecutorFactory {
```
enchobelezirev, line 5: Same, FlowableCommandExecutorFactory

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/commands/abort/AbortCommandExecutor.java
```diff
@@ -0,0 +1,30 @@
...
+import com.sap.cloud.lm.sl.cf.process.flowable.AbortProcessAction;
+import com.sap.cloud.lm.sl.cf.process.flowable.ProcessAction;
+import com.sap.cloud.lm.sl.cf.process.flowable.ProcessActionRegistry;
+import com.sap.cloud.lm.sl.cf.process.flowable.commands.CommandExecutor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.text.MessageFormat;
+
+public class AbortCommandExecutor implements CommandExecutor {
```
enchobelezirev, line 13: This could be renamed to AbortProcessFlowableCommandExecutor

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/commands/abort/AbortCommandExecutorFactory.java
```diff
@@ -0,0 +1,28 @@
...
+import com.sap.cloud.lm.sl.cf.process.flowable.commands.CommandExecutor;
+import com.sap.cloud.lm.sl.cf.process.flowable.commands.CommandExecutorFactory;
+import com.sap.cloud.lm.sl.cf.process.util.HistoryUtil;
+import org.flowable.common.engine.impl.interceptor.CommandContext;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@Named
+public class AbortCommandExecutorFactory implements CommandExecutorFactory {
```
enchobelezirev, line 14: Same, AbortProcessFlowableCommandExec...

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/commands/abort/AbortFailedProcessCommand.java
```diff
@@ -0,0 +1,55 @@
...
+    private String findProcessInstanceId(CommandContext commandContext) {
+        JobEntity job = CommandContextUtil.getJobService(commandContext)
+                                          .findJobById(jobId);
+        return job != null ? job.getProcessInstanceId() : null;
+    }
+
+    private void abortProcessIfRequested(CommandContext commandContext, String processInstanceId) {
+        String correlationId = HistoryUtil.getVariableValue(commandContext, processInstanceId, Constants.VAR_CORRELATION_ID);
+        if (shouldAbortOnError(commandContext, correlationId)) {
+            abortCommandExecutorFactory.getExecutor(commandContext, processInstanceId)
```
enchobelezirev, line 46: Extract to variable

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/commands/abort/AbortFailedProcessCommandFactory.java
```diff
@@ -0,0 +1,26 @@
...
+import org.flowable.engine.impl.jobexecutor.DefaultFailedJobCommandFactory;
+
+import org.springframework.context.annotation.Lazy;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@Named
+@Lazy
+public class AbortFailedProcessCommandFactory extends DefaultFailedJobCommandFactory {
```
enchobelezirev, line 13: This has naming problem from the beginning -> AbortProcessFailedJobCommandFactory

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/flowable/commands/abort/AbortFailedProcessCommandFactory.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private AbortCommandExecutorFactory abortCommandExecutorFactory;
+
+    @Override
+    public Command<Object> getCommand(String jobId, Throwable exception) {
+        return new AbortFailedProcessCommand(abortCommandExecutorFactory, jobId, super.getCommand(jobId, exception));
+    }
+
+    @Inject
+    public void setCommandExecutorFactory(AbortCommandExecutorFactory abortCommandExecutorFactory) {
```
enchobelezirev, line 23: Why not Constructor-based injection?

com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/configuration/FlowableConfiguration.java
```diff
@@ -83,10 +83,6 @@ public SpringProcessEngineConfiguration processEngineConfiguration(DataSource da
         return new ArrayList<>(Arrays.asList(flowableResources));
     }
```
enchobelezirev, line 84: Empty comment for changing the review status

PR #819, opened 2020-03-23
Refactor getting/setting of process variables (Part 3)

nictas:

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/variables/Variables.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/variables/Variables.java
@@ -63,9 +63,45 @@ public interface Variables {
     Variable<String> APP_ARCHIVE_ID = ImmutableVariable.<String> builder()
                                                        .name(Constants.PARAM_APP_ARCHIVE_ID)
                                                        .build();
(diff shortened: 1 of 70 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/variables/Variables.java
```diff
@@ -60,6 +60,12 @@
     Variable<String> SERVICE_TO_PROCESS_NAME = ImmutableVariable.<String> builder()
                                                                 .name(Constants.VAR_SERVICE_TO_PROCESS_NAME)
                                                                 .build();
+    Variable<String> APP_ARCHIVE_ID = ImmutableVariable.<String> builder()
+                                                       .name(Constants.PARAM_APP_ARCHIVE_ID)
+                                                       .build();
+    Variable<String> MTA_ID = ImmutableVariable.<String> builder()
```
enchobelezirev: Are there more variables which are required?
I am asking because if someone adds another required variable, what will happen if he/she forgets to retrieve it using the context.getRequiredProcessVariable? 
I wonder whether there is a way to prevent this - using builder ?

nictas: I prefer to leave this in the hands of the steps. A variable might be "required" in one step and optional in another. Because of this, "required" is not so much an attribute of the variable itself, but more of the step that uses it.

PR #831, opened 2020-04-07
Fix check for dependency satisfaction

radoslav-d:
Make ModuleDependencyChecker verify that a module is either in the
deployed modules or in the modules not for deployment. Add a check in
ComputeNextModulesStep for possible infinite loop as safe mechanism.
JIRA: LMCROSSITXSADEPLOY-2010

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleDependencyChecker.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleDependencyChecker.java
@@ -52,7 +52,7 @@ public class ModuleDependencyChecker {
 
         return module.getDeployedAfter()
                      .isEmpty()
-            || modulesAlreadyDeployed.containsAll(module.getDeployedAfter()) || areAllDependenciesAlreadyPresent(module.getDeployedAfter());
+            || areDependenciesProcessed(module) || areAllDependenciesAlreadyPresent(module.getDeployedAfter());
     }
 
     private boolean areAllDependenciesAlreadyPresent(List<String> deployedAfter) {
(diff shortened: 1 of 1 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleDependencyChecker.java
```diff
@@ -82,4 +83,20 @@ private boolean areAllDependenciesAlreadyPresent(List<String> deployedAfter) {
     public Set<String> getAlreadyDeployedModules() {
         return modulesAlreadyDeployed;
     }
+
+    public Set<String> getNotProcessedModules() {
+        return modulesForDeployment.stream()
+                                   .filter(isProcessed().negate())
```
radito3, line 89: Rather than using the `negate()` method, change `isProcessed()` to accept a parameter and return a boolean then negate that function

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleDependencyChecker.java
```diff
@@ -82,4 +83,20 @@ private boolean areAllDependenciesAlreadyPresent(List<String> deployedAfter) {
...
+                                   .collect(Collectors.toSet());
+    }
+
+    private boolean areDependenciesProcessed(Module module) {
+        return module.getDeployedAfter()
+                     .stream()
+                     .allMatch(isProcessed());
```
radito3, line 96: That way you can use a method reference here

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/Messages.java
```diff
@@ -61,6 +61,7 @@
     public static final String APPLICATION_UNBOUND_IN_PARALLEL = "Application {0} was bound to service {1} which was unbound in parallel";
     public static final String ABORT_OPERATION_TIMED_OUT = "Abort operation timed out";
     public static final String FLOWABLE_JOB_RETRY_FAILED = "Flowable job retry failed";
+    public static final String DEPENDENCIES_CANNOT_BE_SATISFIED = "Dependencies for modules {0} cannot be satisfied";
```
radito3, line 64: "Dependencies can't be satisfied" is too generic. If you are going to print an error message make it as descriptive as possible

PR #836, opened 2020-04-09
Clean leftovers from configuration registry

boyan-velinov:


com.sap.cloud.lm.sl.cf.core/src/main/resources/com/sap/cloud/lm/sl/cf/core/db/changelog/db-changelog-update_configuration_registry_clean_up_leftovers.xml
```diff
@@ -0,0 +1,33 @@
...
+			    SELECT
+			        id
+			    FROM (
+			        SELECT
+			            id,
+			            ROW_NUMBER() OVER w AS rnum
+			        FROM configuration_registry 
+			        WINDOW w AS (
+			            PARTITION BY provider_nid, provider_id, space_id
+			            ORDER BY id DESC
```
boyan-velinov, line 27: I am thinking about improve the ordering:
-> ORDER BY provider_version DESC, id DESC
This will handle the situation when redeployment is done when sequence is changed.

nictas: I'm not a fan of this SQL statement, because I'd have to read a lot of online information to understand it. Maybe we should convert this change to Java code and use the `ConfigurationEntryService`?

boyan-velinov: as agreed offline, the sql will be removed since many of these duplicated configuration entries will be removed after re-deploy.

PR #837, opened 2020-04-10
Improve REST API responses for CC-related errors

nictas:
JIRA: LMCROSSITXSADEPLOY-2037


com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/resources/CFExceptionMapper.java
```diff
@@ -11,29 +12,31 @@
...
 
 @ControllerAdvice
 public class CFExceptionMapper {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CFExceptionMapper.class);
 
-    @ExceptionHandler({ SLException.class, IllegalArgumentException.class })
-    public ResponseEntity<String> handleException(Throwable t) {
+    @ExceptionHandler
+    public ResponseEntity<String> handleException(Exception e) {
```
enchobelezirev, line 22: Why not Throwable?

nictas: Spring exception handlers do not support catching `Throwable`s. The only reason this worked in the past is because of the annotation `@ExceptionHandler({ SLException.class, IllegalArgumentException.class })`. Basically, it caught only `SLException`s and `IllegalArgumentException`s and upcasted them to `Throwable`. However, I wanted this mapper to handle all kinds of exceptions without having to bloat that annotation with additional types. I also wanted to show the exception message regardless of the exception's type. That's why I removed the specific exception types from the annotation, and had to change the signature of the method to match what Spring supports.

PR #840, opened 2020-04-18
Feature flag parameter for disabling env based detection

ddonchev:
#### Description: 
To enable MTA deploy orchestrators working in monstrous spaces to disable env-based detection and avoid hitting the CC rate limit. 
#### Issue: NGPBUG-112089

PS: I know that adding a boolean parameter to a method is not exactly clean code. This code is anyway meant to be deleted in a couple of months, when env-based detection is phased out.


ddonchev:
10х for the fixes!

PR #841, opened 2020-04-21
Fix async service creation

boyan-velinov:
This fix regression provided by commit:
206c84cec7b994e62b7f479fa96da305f72b8146
In the past response from controller was parsed to check
if the service was created and set MethoExecution to FINISH or EXECUTING.
With refered commit we always set MethodExecution to FINISH which is wrong
and in that case PollServiceCreateOrUpdateOperationsExecution is never invoked.
With the current change we will set MethodExecution to EXECUTING alwasy which means that
polling will be executed always.

Jira: LMCROSSITXSADEPLOY-2047


radoslav-d:
Maybe we should consider applying this change to the method for creating user provided service in `CreateServiceStep`

PR #845, opened 2020-04-24
fix SLP issues

VaskoBozhurski:
#### Description: 

-Fix issue that the SLP wrapper exception used was not part of the
ExceptionMapper which resulted in all error cases (bad content, not
found, etc) to always be mapped to 500 (internal server error)
-Fix issue with SLP URLs ending in known file types (for instance .log)
resulting in Spring throwing an error because it could not convert to
said type.

#### Issue: 
LMCROSSITXSADEPLOY-2056


com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/resources/CFExceptionMapper.java
```diff
@@ -35,6 +36,11 @@
         if (e instanceof ConflictException) {
             status = HttpStatus.CONFLICT;
         }
+        if (e instanceof ResponseStatusException) {
```
nictas, line 39: Add a test case for this in the unit test.

VaskoBozhurski: Done.

PR #850, opened 2020-05-02
POC: Masking env variable values

ddonchev:
@nictas - do you think this approach makes any sense?

#### Description: 
Does Not work:
- for the env of detected (to be updated) apps; 
- printed env when app digest is updated (upload app step)
- updating subscriptions

ugly change: 
- cloud controller client api has the notion of sensitive env variables passed as a parameter to the method updating them
- even uglier change  may be required  - persisting the list of sensitive variables as metadata


sunny:
Hi @ddonchev, I think you meant to mention another `sunny`. Have a nice day! 👋

PR #856, opened 2020-05-14
Implement "before application start" hooks

IvanBorislavovDimitrov:
Add support for before start hooks:
application.before-start.idle
application.before-start.live
application.before-start

Jira item: LMCROSSITXSADEPLOY-2058


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/RestartAppStep.java
```diff
@@ -81,6 +87,19 @@ private StartingInfo startApp(CloudControllerClient client, CloudApplication app
         return client.startApplication(app.getName());
     }
 
+    @Override
+    public HookPhase getHookPhaseBeforeStep(ProcessContext context) {
+        ProcessType processType = processTypeParser.getProcessType(context.getExecution());
+        if (ProcessType.BLUE_GREEN_DEPLOY.getName()
+                                         .equals(processType.getName())) {
+            if (context.getVariable(Variables.PHASE) == Phase.RESUME) {
```
enchobelezirev, line 95: Actually, this is ATER_RESUME :)

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/RestartAppStepTest.java
```diff
@@ -4,26 +4,41 @@
```
enchobelezirev, line 4: Merge this commit with the previous one.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleHooksAggregator.java
```diff
@@ -0,0 +1,145 @@
...
+        return resolvedHookPhases.contains(hookTypeForCurrentStepPhase);
+    }
+
+    private List<HookPhase> mapToHookPhases(List<String> hookPhases) {
+        return hookPhases.stream()
+                         .map(HookPhase::fromString)
+                         .collect(Collectors.toList());
+    }
+
+    private List<Hook> getHooksForExecution(Map<String, List<String>> alreadyExecutedHooks, List<Hook> moduleHooksToBeExecuted,
```
radito3, line 80: This method can be done with one `stream().filter().collect()` chain.
<pre> hook -> !alreadyExecutedHooks.containsKey(hook.getName()) || !hasNonExecutedPhases(alreadyExecutedHooks, hook, hookPhaseForCurrentStepPhase)</pre>
That way, there is no need for the following 2 methods

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/RestartAppStepTest.java
```diff
@@ -4,26 +4,41 @@
...
 public class RestartAppStepTest extends SyncFlowableStepTest<RestartAppStep> {
 
     private static final String APP_NAME = "foo";
 
-    @Before
+    @Mock
+    private ProcessTypeParser processTypeParser;
+
+    public RestartAppStepTest() {
+        MockitoAnnotations.initMocks(this);
```
radito3, line 30: SyncFlowableStepTest already does this. Remove it

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/TimeoutAsyncFlowableStepWithHooksTest.java
```diff
@@ -0,0 +1,104 @@
...
+    private Hook createHook(String name) {
+        return Hook.createV3()
+                   .setName(name);
+    }
+
+    private TimeoutAsyncFlowableStepWithHooks createStep() {
+        return new TimeoutAsyncFlowableStepWithHooksMock();
+    }
+
+    private class TimeoutAsyncFlowableStepWithHooksMock extends TimeoutAsyncFlowableStepWithHooks {
```
radito3, line 76: Sonar would most probably complain that this isn't a static inner class

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepWithHooks.java
```diff
@@ -0,0 +1,23 @@
...
+
+    default HookPhase getHookPhaseBeforeStep(ProcessContext context) {
+        return HookPhase.NONE;
+    }
+
+    default HookPhase getHookPhaseAfterStep(ProcessContext context) {
+        return HookPhase.NONE;
+    }
+
+    default HooksHelper.Builder getDefaultHooksHelperBuilder(ProcessContext context) {
```
radito3, line 16: I didn't see this method ever being overridden.
Change it to a static method

com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/model/Phase.java
```diff
@@ -1,5 +1,5 @@
 package com.sap.cloud.lm.sl.cf.core.model;
 
 public enum Phase {
-    UNDEPLOY
+    UNDEPLOY, RESUME
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: To determine whether the deployment is after the resume phase so the correct phase could be returned.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HooksHelper.java
```diff
@@ -0,0 +1,170 @@
...
+import com.sap.cloud.lm.sl.cf.core.model.HookPhase;
+import com.sap.cloud.lm.sl.cf.process.steps.ProcessContext;
+import com.sap.cloud.lm.sl.cf.process.steps.StepPhase;
+import com.sap.cloud.lm.sl.cf.process.steps.StepsUtil;
+import com.sap.cloud.lm.sl.cf.process.variables.Variables;
+import com.sap.cloud.lm.sl.mta.model.DeploymentDescriptor;
+import com.sap.cloud.lm.sl.mta.model.Hook;
+import com.sap.cloud.lm.sl.mta.model.Module;
+
+public class HooksHelper {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: yes

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepWithHooks.java
```diff
@@ -0,0 +1,23 @@
...
+        return new HooksHelper.Builder().withContext(context)
+                                        .withHooksBeforeStep(() -> getHookPhaseBeforeStep(context))
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StepsUtil.java
```diff
@@ -220,11 +219,11 @@ public static boolean getIsServiceUpdatedExportedVariable(VariableScope scope, S
         return getBoolean(scope, Constants.VAR_IS_SERVICE_UPDATED_VAR_PREFIX + serviceName, false);
     }
 
-    static void setExecutedHooksForModule(VariableScope scope, String moduleName, Map<String, List<String>> moduleHooks) {
+    public static void setExecutedHooksForModule(VariableScope scope, String moduleName, Map<String, List<String>> moduleHooks) {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: I don't think so because the name of the executed hooks is dynamically calculated from the module's name.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleHooksAggregator.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private List<Hook> collectHooksWithPhase(Module moduleToDeploy, HookPhase hookTypeForCurrentStepPhase) {
+        return getModuleHooks(moduleToDeploy).stream()
+                                             .filter(hook -> shouldExecuteHook(hook.getPhases(), hookTypeForCurrentStepPhase))
+                                             .collect(Collectors.toList());
+    }
+
+    private List<Hook> getModuleHooks(Module moduleToDeploy) {
+        return moduleToDeploy.getMajorSchemaVersion() < MAJOR_SCHEMA_VERSION_THREE ? Collections.emptyList() : moduleToDeploy.getHooks();
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: If the Major Schema Version is less than 3 an exception will be thrown while trying to invoke module.getHooks().

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleHooksAggregator.java
```diff
@@ -0,0 +1,145 @@
...
+                                                                                              hook.getPhases());
+        if (alreadyExecutedHooks.containsKey(hook.getName())) {
+            alreadyExecutedHooks.get(hook.getName())
+                                .addAll(hookPhasesBasedOnCurrentHookPhase);
+            return;
+        }
+        alreadyExecutedHooks.put(hook.getName(), hookPhasesBasedOnCurrentHookPhase);
+    }
+
+    private List<String> getHookPhasesBasedOnCurrentHookPhase(HookPhase currentHookPhaseForExecution, List<String> hookPhases) {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: As you have noticed its only one element unless there are repeating hooks. But this also covers the case that the hook will be executed only one time even though it has repeating phases. In my opinion, it's safer this way.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleHooksAggregator.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private List<HookPhase> getExecutedHookPhasesForHook(Map<String, List<String>> alreadyExecutedHooks, String hookName) {
+        List<String> executedHookPhasesForHook = alreadyExecutedHooks.get(hookName);
+        return mapToHookPhases(executedHookPhasesForHook);
+    }
+
+    private void updateExecutedHooksForModule(Map<String, List<String>> alreadyExecutedHooks, HookPhase currentHookPhaseForExecution,
+                                              List<Hook> hooksForExecution) {
+        Map<String, List<String>> result = new HashMap<>(alreadyExecutedHooks);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: I extracted this code, but I think because it's an immutable map when its fetched from the process' context.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/RestartAppStep.java
```diff
@@ -81,6 +99,15 @@ private StartingInfo startApp(CloudControllerClient client, CloudApplication app
         return client.startApplication(app.getName());
     }
 
+    @Override
+    public HookPhase getHookPhaseBeforeStep(ProcessContext context) {
+        ProcessType processType = processTypeParser.getProcessType(context.getExecution());
+        if (ProcessType.BLUE_GREEN_DEPLOY.equals(processType) && (context.getVariable(Variables.PHASE) != Phase.AFTER_RESUME)) {
```
radito3, line 105: There is no need for the extra parenthesis around the second condition

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HooksHelper.java
```diff
@@ -0,0 +1,163 @@
...
+        HandlerFactory handlerFactory = StepsUtil.getHandlerFactory(context.getExecution());
+        return findModuleByNameFromDeploymentDescriptor(handlerFactory, deploymentDescriptor, cloudApplication.getModuleName());
+    }
+
+    private Module determineModuleFromAppName(DeploymentDescriptor deploymentDescriptor, CloudApplicationExtended cloudApplication) {
+        String moduleName = getModuleName(cloudApplication);
+        if (moduleName == null) {
+            return null;
+        }
+        return deploymentDescriptor.getModules()
```
radito3, line 76: This does the same as DescriptorHandler::findModule

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HooksHelper.java
```diff
@@ -0,0 +1,163 @@
...
+
+        CloudApplicationExtended cloudApplication = context.getVariable(Variables.APP_TO_PROCESS);
+        if (cloudApplication.getModuleName() == null) {
+            // This case handles the deletion of old applications when the process is blue-green deployment. Here the application is taken
+            // from the
+            // CloudController and thus we do not have moduleName in it.
+            return determineModuleFromAppName(deploymentDescriptor, cloudApplication);
+        }
+
+        HandlerFactory handlerFactory = StepsUtil.getHandlerFactory(context.getExecution());
```
radito3, line 67: We can use this in the `determineModuleFromAppName` method as well, so move it above the second if condition.
Also, since that `determine` method does the same as the `findModuleByNameFromDeploymentDescriptor` method only with one more processing of the module name, we can refactor this method so as to "determine" the module name that needs checking in the if statement (whether it's directly from the cloud application object, or after mta metadata / env parsing) and then passing it to the `findModule` method.
Some renaming may be worthwhile and maybe some extracting into methods

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HooksProcessor.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+        if (!hooksHelper.isInPreExecuteStepPhase(currentStepPhase)) {
+            return false;
+        }
+        List<Hook> hooksForExecution = hooksHelper.getHooksForStepPhase(moduleToDeploy, currentStepPhase);
+        hooksHelper.setHooksForExecution(hooksForExecution);
+        return !hooksForExecution.isEmpty();
+    }
+
+    private boolean executeAfterStepHooks(HooksHelper hooksHelper, StepPhase currentStepPhase, Module moduleToDeploy) {
```
radito3, line 52: This method and the one before it differ only in one `if` condition. Is it possible to refactor them to remove the duplication?

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteApplicationRoutesStepTest.java
```diff
@@ -8,9 +8,15 @@
...
 
 import org.cloudfoundry.client.lib.domain.CloudApplication;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import com.sap.cloud.lm.sl.cf.process.util.HooksProcessor;
 
 public class DeleteApplicationRoutesStepTest extends UndeployAppStepTest {
 
+    private final HooksProcessor hooksProcessor = Mockito.mock(HooksProcessor.class);
```
radito3, line 18: Why not just the `@Mock` annotation?

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/steps/RestartAppStepTest.java
```diff
@@ -4,26 +4,37 @@
...
+import com.sap.cloud.lm.sl.cf.process.util.HooksProcessor;
+import com.sap.cloud.lm.sl.cf.process.util.ProcessTypeParser;
 import com.sap.cloud.lm.sl.cf.process.variables.Variables;
+import com.sap.cloud.lm.sl.cf.web.api.model.ProcessType;
 import com.sap.cloud.lm.sl.common.util.JsonUtil;
 
 public class RestartAppStepTest extends SyncFlowableStepTest<RestartAppStep> {
 
     private static final String APP_NAME = "foo";
+    private final HooksProcessor hooksProcessor = Mockito.mock(HooksProcessor.class);
```
radito3, line 25: Use `@Mock` instead of `Mockito::mock` in this and all following instances

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HooksHelper.java
```diff
@@ -0,0 +1,161 @@
...
+            // This will be the case only when the process is undeploy.
+            return null;
+        }
+        return determineModuleFromDeploymentDescriptor(context.getVariable(Variables.APP_TO_PROCESS), deploymentDescriptor);
+    }
+
+    private Module determineModuleFromDeploymentDescriptor(CloudApplicationExtended cloudApplication,
+                                                           DeploymentDescriptor deploymentDescriptor) {
+        HandlerFactory handlerFactory = StepsUtil.getHandlerFactory(context.getExecution());
+        if (cloudApplication.getModuleName() == null) {
```
radito3, line 64: I was thinking more along the lines of
<pre>
Module determineModuleFromDescriptor(...){
   HandlerFactory handlerFactory = getHandlerFactory();
   String moduleName = computeModuleName(...);
   return findModuleByName(...);
}

String computeModuleName(...) {
   if (cloudApp.getModuleName() == null {
      return parseModuleName(); //use getDeployedMtaApplication().getModuleName() here
   }
   return cloudApp.getModuleName();
}
</pre>

But your solution is good as well

IvanBorislavovDimitrov: Looks better. Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/DeleteApplicationRoutesStep.java
```diff
@@ -25,19 +27,24 @@
...
-    }
+    private MtaMetadataParser mtaMetadataParser;
+    @Inject
+    private EnvMtaMetadataParser envMtaMetadataParser;
 
     @Override
-    protected StepPhase undeployApplication(CloudControllerClient client, CloudApplication cloudApplicationToUndeploy) {
+    protected StepPhase undeployApplication(CloudControllerClient client, CloudApplication cloudApplicationToUndeploy,
+                                            ProcessContext context) {
+        HooksProcessor hooksProcessor = getHooksProcessor(context);
```
enchobelezirev, line 42: This refactoring introduces a lot of copy-pasted code. Meaning that if I want to enable the hooks in another step, I need to copy-paste this code which does all the magic. My point is that we need to make external logic which should handles this before and after step hooks calculation. Also this logic should use the newly introduced interfaces in order to operate. Meaning that when some step is started, if it is "specified" that it supports hooks, the hook building logic should understand that this step is either of type "AfterStepHookPhaseProvider" and/or "BeforeStepHookPhaseProvider" and calculate the hooks based on the current step phase.
We could discuss it - the design and the whole concept.
Try thinking of a pattern which is executing some logic by calling some delegates in order to do its job.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HooksHelper.java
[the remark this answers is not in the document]
enchobelezirev: I would try to make it @ Immutable

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleHooksAggregator.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public List<Hook> aggregateHooks(HookPhase currentHookPhaseForExecution) {
+        substituteBeforeStartHooksPhases();
+        Map<String, List<String>> alreadyExecutedHooksForModule = getAlreadyExecutedHooks();
+        List<Hook> hooksCalculatedForExecution = determineHooksForExecution(alreadyExecutedHooksForModule, currentHookPhaseForExecution);
+        updateExecutedHooksForModule(alreadyExecutedHooksForModule, currentHookPhaseForExecution, hooksCalculatedForExecution);
+        return hooksCalculatedForExecution;
+    }
+
+    private void substituteBeforeStartHooksPhases() {
```
enchobelezirev, line 38: I am concerned about this. First, why is this here? Second, this brings notation for the before start hook phases which is not correct from the Object point of view. It's purpose is to get some hooks, calculate some hooks for execution and return them.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleHooksAggregator.java
```diff
@@ -0,0 +1,125 @@
...
+        return hookPhases.stream()
+                         .map(HookPhase::fromString)
+                         .collect(Collectors.toList());
+    }
+
+    private List<Hook> getHooksForExecution(Map<String, List<String>> alreadyExecutedHooks, List<Hook> moduleHooksToBeExecuted,
+                                            HookPhase hookPhaseForCurrentStepPhase) {
+        return moduleHooksToBeExecuted.stream()
+                                      .filter(hook -> !alreadyExecutedHooks.containsKey(hook.getName())
+                                          || !hasAllPhasesExecuted(alreadyExecutedHooks, hook, hookPhaseForCurrentStepPhase))
```
enchobelezirev, line 84: I would extract this condition in a separate method

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/listeners/SetResumePhaseListener.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.flowable.engine.delegate.DelegateExecution;
+import org.flowable.engine.delegate.ExecutionListener;
+
+import com.sap.cloud.lm.sl.cf.core.model.Phase;
+import com.sap.cloud.lm.sl.cf.process.variables.VariableHandling;
+import com.sap.cloud.lm.sl.cf.process.variables.Variables;
+
+@Named("setResumePhase")
+public class SetResumePhaseListener implements ExecutionListener {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/AfterHooksPhaseProvider.java
```diff
@@ -0,0 +1,8 @@
+package com.sap.cloud.lm.sl.cf.process.steps;
+
+import com.sap.cloud.lm.sl.cf.core.model.HookPhase;
+
+public interface AfterHooksPhaseProvider {
+
+    HookPhase getHookPhaseAfterStep(ProcessContext context);
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: get methods will conflict with each other if a step implements both interfaces
Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/BeforeHooksPhaseProvider.java
```diff
@@ -0,0 +1,8 @@
+package com.sap.cloud.lm.sl.cf.process.steps;
+
+import com.sap.cloud.lm.sl.cf.core.model.HookPhase;
+
+public interface BeforeHooksPhaseProvider {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleHooksAggregator.java
IvanBorislavovDimitrov: Removed. RestartAppStep will return the required phases

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ModuleHooksAggregator.java
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/AfterHooksPhaseProvider.java
```diff
@@ -0,0 +1,8 @@
+package com.sap.cloud.lm.sl.cf.process.steps;
+
+import com.sap.cloud.lm.sl.cf.core.model.HookPhase;
+
+public interface AfterHooksPhaseProvider {
```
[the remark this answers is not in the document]
IvanBorislavovDimitrov: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/RestartAppStep.java
```diff
@@ -17,18 +17,30 @@
...
+public class RestartAppStep extends TimeoutAsyncFlowableStepWithHooks implements BeforeStepHookPhaseProvider {
 
     @Inject
     protected RecentLogsRetriever recentLogsRetriever;
+    @Inject
+    private ProcessTypeParser processTypeParser;
+    @Inject
+    private MtaMetadataParser mtaMetadataParser;
+    @Inject
+    private EnvMtaMetadataParser envMtaMetadataParser;
```
enchobelezirev, line 40: Do we need these lines of code here?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StopAppStep.java
```diff
@@ -20,13 +23,17 @@
...
 @Scope(BeanDefinition.SCOPE_PROTOTYPE)
-public class StopAppStep extends SyncFlowableStepWithHooks {
+public class StopAppStep extends SyncFlowableStepWithHooks implements BeforeStepHookPhaseProvider, AfterStepHookPhaseProvider {
 
     @Inject
     private ProcessTypeParser processTypeParser;
+    @Inject
+    private MtaMetadataParser mtaMetadataParser;
+    @Inject
+    private EnvMtaMetadataParser envMtaMetadataParser;
```
enchobelezirev, line 33: Again, do we need these lines of code here?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/StopApplicationUndeploymentStep.java
```diff
@@ -1,29 +1,38 @@
...
 
 @Named("stopApplicationUndeploymentStep")
 @Scope(BeanDefinition.SCOPE_PROTOTYPE)
-public class StopApplicationUndeploymentStep extends UndeployAppStep {
+public class StopApplicationUndeploymentStep extends UndeployAppStep implements BeforeStepHookPhaseProvider, AfterStepHookPhaseProvider {
+
+    @Inject
+    private EnvMtaMetadataParser envMtaMetadataParser;
+    @Inject
+    private MtaMetadataParser mtaMetadataParser;
```
enchobelezirev, line 28: Same

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HooksProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+
+import javax.inject.Named;
+
+import com.sap.cloud.lm.sl.cf.process.steps.StepPhase;
+import com.sap.cloud.lm.sl.mta.model.Hook;
+import com.sap.cloud.lm.sl.mta.model.Module;
+
+@Named
+public class HooksProcessor {
```
enchobelezirev, line 13: I would say that this is HooksExecutor class as there are only execute* methods in it.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/HooksHelper.java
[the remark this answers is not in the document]
enchobelezirev: I do not like this Helper pattern because it is "сбирщайн" of different logics which are combined in a class and each of each logic has noting in common with the other ones. Maybe it would be great if we could extract these separate logics into different classes and check what will be result?

PR #868, opened 2020-06-04
Refactor parameters and their metadata

nictas:
JIRA: LMCROSSITXSADEPLOY-2069


com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/ParameterMetadata.java
```diff
@@ -11,23 +11,17 @@
 @JsonDeserialize(as = ImmutableParameterMetadata.class)
 public interface ParameterMetadata {
 
-    @Nullable
```
enchobelezirev, line 14: They are required now?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/metadata/UndeployMetadata.java
```diff
@@ -12,60 +9,54 @@
 
 public class UndeployMetadata {
 
-    private static final Set<ImmutableParameterMetadata> PARAMS = new HashSet<>();
```
enchobelezirev, line 15: Ivan R. would not be happy at all about this... :D

com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/BooleanParameterConverter.java
```diff
@@ -0,0 +1,21 @@
...
+
+    @Override
+    public Object convert(Object value) {
+        String parameterValue = String.valueOf(value);
+        if ("true".equals(parameterValue)) {
+            return true;
+        } else if ("false".equals(parameterValue)) {
+            return false;
+        } else {
+            throw new SLException(MessageFormat.format("Parameter is not boolean {0}", value));
```
enchobelezirev: Do we need this exception? If not, we could make Boolean.valueOf?
How stupid would be to think that "opusum" = false?
Something else, we could implement a logic which is for validation of those parameters and then the conversion to happen?

com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/ParameterMetadata.java
```diff
@@ -23,4 +24,7 @@ default boolean getRequired() {
 
     ParameterType getType();
 
+    @Nullable
+    ParameterConverter getCustomConverter();
```
enchobelezirev, line 28: Why custom? Isn't that the default converter for this parameter or?

nictas: Nah, the converter that matches the type of the parameter is picked by the `ParameterConversion` class, if this one is `null`.

com.sap.cloud.lm.sl.cf.api/src/main/java/com/sap/cloud/lm/sl/cf/web/api/model/ParameterMetadata.java
nictas: Yes, it wouldn't make sense to have a parameter without an ID, so it should not be allowed at all.

PR #874, opened 2020-06-09
Fix detection of old blue-green processes

theghost5800:
Due to new abort logic, when both application colors exists (blue/green)
application color detector fails to find historic data for previous
blue-green process

JIRA:LMCROSSITXSADEPLOY-2097

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationColorDetector.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationColorDetector.java
@@ -1,25 +1,32 @@
 package com.sap.cloud.lm.sl.cf.process.util;
 
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.inject.Named;
 
-import org.apache.commons.collections4.CollectionUtils;
 import org.flowable.variable.api.history.HistoricVariableInstance;
 
 import com.sap.cloud.lm.sl.cf.core.Messages;
 import com.sap.cloud.lm.sl.cf.core.model.ApplicationColor;
 import com.sap.cloud.lm.sl.cf.core.model.DeployedMta;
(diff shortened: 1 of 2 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationColorDetector.java
```diff
@@ -100,13 +100,59 @@ private ApplicationColor getOlderApplicationColor(DeployedMta deployedMta) {
...
-    private Phase getPhaseFromHistoricProcess(String processInstanceId) {
-        HistoricVariableInstance phaseVariableInstance = flowableFacade.getHistoricVariableInstance(processInstanceId, Variables.PHASE.getName());
-        if (phaseVariableInstance == null) {
-            return null;
-        }
+    private Optional<Operation> getLastOperationsWithFinalState(String correlationId) {
+        List<Operation> operations = getOperations(correlationId);
 
-        return Phase.valueOf((String) phaseVariableInstance.getValue());
+        Map<String, List<HistoricOperationEvent>> operationsWithHistoricEvents = operations.stream()
```
radito3, line 106: I think this aggregation to a map first and then extracting from it is unneded.
This function can be simplified like so:
<pre>
List<> operations = getOperations();
return operations.stream().map(this::getFinalState).filter(op -> op.state != null).max(Comparator.comparing(endedAt);
</pre>
or you can even substitute the `max()` operation with `findFirst()` because `getOperations()` is already sorted
more on the `getFinalState()` changes in a later comment

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationColorDetector.java
```diff
@@ -100,13 +100,59 @@ private ApplicationColor getOlderApplicationColor(DeployedMta deployedMta) {
...
+                               .list();
+    }
+
+    private List<HistoricOperationEvent> getOperationHistoricEvents(Operation operation) {
+        return historicOperationEventService.createQuery()
+                                            .processId(operation.getProcessId())
+                                            .list();
+    }
+
+    private Operation getFinalState(Operation operation, Map<String, List<HistoricOperationEvent>> operationsWithHistoricEvents) {
```
radito3, line 138: This can be changed like so:
<pre>
if (operation.getState() == null) {
   return operation;
}
List<> opHistoricEvents = getOperationHistoricEvents(operation);
</pre>
and the rest can be the same

PR #875, opened 2020-06-10
Fix application color detector

theghost5800:


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/persistence/query/Query.java
```diff
@@ -3,14 +3,21 @@
...
 
     T limitOnSelect(int limit);
 
     T offsetOnSelect(int offset);
 
-    R singleResult() throws NoResultException;
+    /**
+     * @return the result
+     * @throws NoResultException Thrown when there is no result. Do not use singleResult() method in case you can expect no result
+     * @throws NonUniqueResultException Thrown when there is multiple results. If you would like to get single result only, you can use
```
nictas, line 17: Change to `Thrown when there are multiple results. If you want to get the first of these, use {@code limitOnSelect(1).list()} instead.`

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ApplicationColorDetector.java
```diff
@@ -58,12 +57,11 @@ public ApplicationColor detectLiveApplicationColor(DeployedMta deployedMta, Stri
                       .getState() != Operation.State.ABORTED) {
             return olderApplicationColor;
         }
-        String xs2BlueGreenDeployHistoricProcessInstanceId = flowableFacade.findHistoricProcessInstanceIdByProcessDefinitionKey(operations.get(0)
-                                                                                                                                          .getProcessId(),
-                                                                                                                                Constants.BLUE_GREEN_DEPLOY_SERVICE_ID);
+        String processIdOfPreviousProcess = operations.get(0)
```
nictas, line 60: Rename to `previousProcessId`.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ProcessConflictPreventer.java
```diff
@@ -66,13 +67,15 @@ private void validateNoConflictingOperationsExist(String mtaId, String spaceId)
                                .list();
     }
 
-    public void releaseLock(String processInstanceId) {
```
nictas, line 69: `endedAt` can be hardcoded to `ZonedDateTime.now`. I don't foresee anyone wanting to pass some different value.

theghost5800: If someone decide to use releaseLock method for different purpose and don't want to set final state, he can pass null and you won't have final state but you will have endedAt field already set.

PR #877, opened 2020-06-12
Fix possible file leak

IvanBorislavovDimitrov:
Issue: LMCROSSITXSADEPLOY-2100


com.sap.cloud.lm.sl.cf.persistence/src/main/java/com/sap/cloud/lm/sl/cf/persistence/services/FileUploader.java
```diff
@@ -69,6 +69,9 @@ public static FileInfo uploadFile(InputStream is) throws FileStorageException {
         } catch (IOException e) {
```
nictas, line 69: Why not just change this catch block to catch `Exception`s instead of `IOException`s?

PR #883, opened 2020-06-17
Change retrival of default domain to not fail

radito3:
#### Issue: LMCROSSITXSADEPLOY-2007

### DIFF
--- a/com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/NameUtilTest.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/NameUtilTest.java
@@ -17,38 +17,6 @@ import com.sap.cloud.lm.sl.cf.core.util.NameUtil.NameRequirements;
 
 public class NameUtilTest {
 
-    @Test
-    public void testIsValidXsAppNameWhenNamesAreValid() {
-        testIsValidName(Arrays.asList("Fo0", "foo.bar", "foo_bar", "foo-bar", "//foo\\"), NameRequirements.XS_APP_NAME_PATTERN, true);
-    }
-
-    @Test
-    public void testIsValidXsAppNameWhenNamesAreInvalid() {
-        String longName = StringUtils.repeat("f", NameRequirements.XS_APP_NAME_MAX_LENGTH + 1);
-
-        testIsValidName(Arrays.asList("sap_system", "sap_system_1", "foo&&bar", "foo bar", longName), NameRequirements.XS_APP_NAME_PATTERN,
-                        false);
-    }
-
-    @Test
-    public void testIsValidContainerNameWhenNamesAreValid() {
-        testIsValidName(Arrays.asList("FOO", "FOO_BAR", "1_FOO", "FOO_1_BAR"), NameRequirements.CONTAINER_NAME_PATTERN, true);
-    }
-
-    @Test
-    public void testIsValidContainerNameWhenNamesAreInvalid() {
-        String longName = StringUtils.repeat("F", NameRequirements.CONTAINER_NAME_MAX_LENGTH + 1);
-
-        testIsValidName(Arrays.asList("foo", "FOO BAR", "1-FOO", "FOO_&_BAR", "_FOO", " ", longName),
-                        NameRequirements.CONTAINER_NAME_PATTERN, false);
-    }
-
-    private void testIsValidName(List<String> names, String namePattern, boolean expectedResult) {
-        for (String name : names) {
-            assertEquals(expectedResult, NameUtil.isValidName(name, namePattern));
-        }
-    }
-
     public static Stream<Arguments> testGetNameWithProperLength() {
         return Stream.of(
         // @formatter:off
@@ -70,7 +38,7 @@ public class NameUtilTest {
         String containerName = NameUtil.computeValidContainerName("initial", "initial",
                                                                   "com.sap.cloud.lm.sl.xs2.a.very.very.long.service.name.with.illegal.container.name.characters");
         assertEquals("INITIAL_INITIAL_COM_SAP_CLOUD_LM_SL_XS2_A_VERY_VERY_LONG3AC0B612", containerName);
-        assertTrue(NameUtil.isValidName(containerName, NameRequirements.CONTAINER_NAME_PATTERN));
+        assertTrue(containerName.matches(NameRequirements.CONTAINER_NAME_PATTERN));
     }
 
     public static Stream<Arguments> testCreateValidXsAppName() {
@@ -88,7 +56,7 @@ public class NameUtilTest {
     public void testCreateValidXsAppName(String serviceName, String expectedName) {
         String xsAppName1 = NameUtil.computeValidXsAppName(serviceName);
         assertEquals(expectedName, xsAppName1);
-        assertTrue(NameUtil.isValidName(xsAppName1, NameRequirements.XS_APP_NAME_PATTERN));
+        assertTrue(xsAppName1.matches(NameRequirements.XS_APP_NAME_PATTERN));
     }
 
     public static Stream<Arguments> testComputeNamespacedNameWithLength() {
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/BuildCloudDeployModelStep.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/BuildCloudDeployModelStep.java
@@ -19,6 +19,7 @@ import org.flowable.engine.delegate.DelegateExecution;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Scope;
 
+import com.sap.cloud.lm.sl.common.SLException;
 import com.sap.cloud.lm.sl.cf.client.lib.domain.CloudServiceInstanceExtended;
 import com.sap.cloud.lm.sl.cf.core.cf.HandlerFactory;
 import com.sap.cloud.lm.sl.cf.core.cf.util.CloudModelBuilderContentCalculator;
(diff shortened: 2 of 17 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/BuildCloudDeployModelStep.java
```diff
@@ -240,22 +242,27 @@ protected ServiceKeysCloudModelBuilder getServiceKeysCloudModelBuilder(ProcessCo
...
         for (Module module : modules) {
             if (!moduleToDeployHelper.isApplication(module)) {
                 continue;
             }
             ParametersChainBuilder parametersChainBuilder = new ParametersChainBuilder(context.getVariable(Variables.COMPLETE_DEPLOYMENT_DESCRIPTOR));
-            List<String> appDomains = applicationCloudModelBuilder.getApplicationDomains(parametersChainBuilder.buildModuleChain(module.getName()),
-                                                                                         module);
+            List<Map<String, Object>> paramsList = parametersChainBuilder.buildModuleChain(module.getName());
+
+            boolean noRoute = (Boolean) PropertiesUtil.getPropertyValue(paramsList, SupportedParameters.NO_ROUTE, false);
```
theghost5800, line 253: This could be throw class cast exception if the user set non boolean value for "no-route" parameter. getProprtyValue method should be refactored to pass what class type you expect and throw exception if doesn't match to the required class type. If this refactor requires a lot of change, you can skip it but track BI in our jira or look for existing one and increase the prio.

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/util/NameUtilTest.java
```diff
@@ -17,38 +17,6 @@
 
 public class NameUtilTest {
 
-    @Test
-    public void testIsValidXsAppNameWhenNamesAreValid() {
```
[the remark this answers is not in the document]
radito3: Because the method "isNameValid" did nothing custom, it just delegated its arguments to String::matches. That's why it's deleted and so are its tests.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CollectSystemParametersStep.java
```diff
@@ -68,12 +68,14 @@ protected String getStepErrorMessage(ProcessContext context) {
         return Messages.ERROR_COLLECTING_SYSTEM_PARAMETERS;
     }
 
-    private String getDefaultDomain(CloudControllerClient client) {
-        CloudDomain defaultDomain = client.getDefaultDomain();
-        if (defaultDomain != null) {
-            return defaultDomain.getName();
+    private String getDefaultDomain(CloudControllerClient client, ProcessContext context) {
+        try {
+            return client.getDefaultDomain()
```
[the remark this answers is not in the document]
radito3: Our cf java client can't return a null object on `getDefaultDomain`

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CollectSystemParametersStep.java
```diff
@@ -68,12 +68,14 @@ protected String getStepErrorMessage(ProcessContext context) {
...
+                         .getName();
+        } catch (CloudOperationException exc) {
+            context.setVariable(Variables.MISSING_DEFAULT_DOMAIN, true);
+            return "apps.internal";
```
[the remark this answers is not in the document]
radito3: It doesn't work with null. The reference parsing for `${default-domain}` in the mta descriptor will fail if this is null

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
```diff
@@ -34,27 +35,26 @@ public void check(DeploymentDescriptor descriptor) {
...
             commonReadOnlyParameters.put(namedParametersContainer.getName(), commonParameters);
         }
     }
 
     private void checkCollectionForCommonParameters(List<? extends NamedParametersContainer> namedElementsWithParametersContainers,
                                                     Set<String> readOnlyParameters, Map<String, Set<String>> commonReadOnlyParameters) {
-        namedElementsWithParametersContainers.forEach(namedElementWithParametersContainers -> checkForCommonParameters(namedElementWithParametersContainers,
-                                                                                                                       readOnlyParameters,
-                                                                                                                       commonReadOnlyParameters));
+        for (NamedParametersContainer namedElement : namedElementsWithParametersContainers) {
```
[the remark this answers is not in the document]
radito3: In my opinion the lambda expression was too long a line and it was unreadable

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/BuildCloudDeployModelStep.java
```diff
@@ -240,22 +242,27 @@ protected ServiceKeysCloudModelBuilder getServiceKeysCloudModelBuilder(ProcessCo
...
-
         Set<String> domains = new TreeSet<>();
         for (Module module : modules) {
             if (!moduleToDeployHelper.isApplication(module)) {
                 continue;
             }
             ParametersChainBuilder parametersChainBuilder = new ParametersChainBuilder(context.getVariable(Variables.COMPLETE_DEPLOYMENT_DESCRIPTOR));
-            List<String> appDomains = applicationCloudModelBuilder.getApplicationDomains(parametersChainBuilder.buildModuleChain(module.getName()),
-                                                                                         module);
+            List<Map<String, Object>> paramsList = parametersChainBuilder.buildModuleChain(module.getName());
```
boyan-velinov, line 251: Rename to parametersList

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CollectSystemParametersStep.java
```diff
@@ -68,12 +70,14 @@ protected String getStepErrorMessage(ProcessContext context) {
...
 
-    private String getDefaultDomain(CloudControllerClient client) {
-        CloudDomain defaultDomain = client.getDefaultDomain();
-        if (defaultDomain != null) {
-            return defaultDomain.getName();
+    private String getDefaultDomain(CloudControllerClient client, ProcessContext context) {
+        try {
+            return client.getDefaultDomain()
+                         .getName();
+        } catch (CloudOperationException exc) {
```
boyan-velinov, line 77: What if CloudOPerationException is thrown for other reaseon?
I think, we will consider that default domain is missing where there was a temporary issue with controller, for example.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
```diff
@@ -34,27 +35,26 @@ public void check(DeploymentDescriptor descriptor) {
 
     private void checkForCommonParameters(NamedParametersContainer namedParametersContainer, Set<String> readOnlyParameters,
                                           Map<String, Set<String>> commonReadOnlyParameters) {
-        Set<String> commonParameters = Sets.intersection(namedParametersContainer.getParameters()
-                                                                                 .keySet(),
-                                                         readOnlyParameters);
+        Set<String> commonParameters = SetUtils.intersection(namedParametersContainer.getParameters()
```
boyan-velinov, line 38: Why this is replaced and formatting is changed?

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CollectSystemParametersStep.java
radito3: For that to happen, the exception must be thrown 4 times. This is our retryable client, so if a timeout, or a network issue occurs, it will retry automatically. But for a missing default domain, it will always fail with this exception.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
radito3: Guava should only be kept as a transitive dependency, there is no need to use it when the Apache commons libraries have simpler implementations for some utility methods, like this one. Otherwise, for the formatting it's a mistake on my part.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/BuildCloudDeployModelStep.java
radito3: Done

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/util/ReadOnlyParametersChecker.java
boyan-velinov: Good!
I missed that Sets is coming from guava

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/CollectSystemParametersStep.java
boyan-velinov: I cannot fully agree with that because client is defined as CloudControllerClient, not as some special RetryableClient or anything else that have some logic.
Also CloudControllerClient does not declare throwing of such exception.
So if the retry failed all 4 times, what will happen. I guess we will continue silently with the DEFAULT_DOMAIN_PLACEHOLDER, right?

PR #887, opened 2020-06-25
Db transaction timeout

theghost5800:


com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/resources/CFExceptionMapper.java
```diff
@@ -41,6 +44,9 @@
             status = rse.getStatus();
             message = rse.getReason();
         }
+        if (e instanceof SQLException || e instanceof PersistenceException) {
+            message = Messages.PROBLEM_WITH_PERSISTENCE_LAYER;
```
boyan-velinov, line 48: What is the purpose of this message?
I mean whether the customer will see. If so, I think it is not good to expose our internal failures outside.

Other idea is to use log marker, see CleanUpJob

PR #888, opened 2020-06-30
Fix cleanup of processes in Flowable tables

theghost5800:
Change logic for searching active operations in AbortProcessListener and
add new cleaner which will delete all left processes in act_ru_execution
table by TTL

JIRA:LMCROSSITXSADEPLOY-2120

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/jobs/AbortedOperationsCleaner.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/jobs/AbortedOperationsCleaner.java
@@ -8,7 +8,6 @@ import java.util.List;
 
 import javax.inject.Inject;
 import javax.inject.Named;
-import javax.persistence.NoResultException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -17,7 +16,6 @@ import org.springframework.core.annotation.Order;
 import com.sap.cloud.lm.sl.cf.core.model.HistoricOperationEvent;
 import com.sap.cloud.lm.sl.cf.core.model.HistoricOperationEvent.EventType;
 import com.sap.cloud.lm.sl.cf.core.persistence.service.HistoricOperationEventService;
-import com.sap.cloud.lm.sl.cf.core.persistence.service.OperationService;
 import com.sap.cloud.lm.sl.cf.process.Messages;
 import com.sap.cloud.lm.sl.cf.process.flowable.FlowableFacade;
 import com.sap.cloud.lm.sl.cf.web.api.model.Operation;
@@ -30,14 +28,11 @@ public class AbortedOperationsCleaner implements Cleaner {
 
     private HistoricOperationEventService historicOperationEventService;
     private FlowableFacade flowableFacade;
-    private OperationService operationService;
 
     @Inject
-    public AbortedOperationsCleaner(HistoricOperationEventService historicOperationEventService, FlowableFacade flowableFacade,
-                                    OperationService operationService) {
+    public AbortedOperationsCleaner(HistoricOperationEventService historicOperationEventService, FlowableFacade flowableFacade) {
         this.historicOperationEventService = historicOperationEventService;
         this.flowableFacade = flowableFacade;
-        this.operationService = operationService;
     }
 
     @Override
@@ -55,19 +50,8 @@ public class AbortedOperationsCleaner implements Cleaner {
                          .forEach(this::deleteProcessInstance);
     }
 
-    private boolean isInActiveState(String operationId) {
-        Operation operation = getOperation(operationId);
-        return operation != null && operation.getState() == null;
-    }
-
-    private Operation getOperation(String operationId) {
-        try {
-            return operationService.createQuery()
-                                   .processId(operationId)
-                                   .singleResult();
-        } catch (NoResultException e) {
-            return null;
-        }
+    private boolean isInActiveState(String processId) {
+        return flowableFacade.getProcessInstance(processId) != null;
     }
 
     private void deleteProcessInstance(String processInstanceId) {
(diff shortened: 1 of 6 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/jobs/FlowableOperationsCleaner.java
```diff
@@ -0,0 +1,51 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.core.annotation.Order;
+
+import com.sap.cloud.lm.sl.cf.process.Messages;
+import com.sap.cloud.lm.sl.cf.process.flowable.FlowableFacade;
+import com.sap.cloud.lm.sl.cf.web.api.model.Operation;
+
+@Named
+@Order(30)
+public class FlowableOperationsCleaner implements Cleaner {
```
nictas, line 21: Flowable uses "process" as a term and not "operation". Operations exist only in our codebase and we're not using the `OperationService` here, so the current name is a little confusing and misleading. I suggest renaming this class to `FlowableDataCleaner`, which will be consistent with `FlowableHistoricDataCleaner`.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/jobs/FlowableOperationsCleanerTest.java
```diff
@@ -0,0 +1,68 @@
...
+    private FlowableOperationsCleaner flowableOperationsCleaner;
+
+    @BeforeEach
+    void setUp() {
+        MockitoAnnotations.initMocks(this);
+        flowableOperationsCleaner = new FlowableOperationsCleaner(flowableFacade);
+    }
+
+    static Stream<Arguments> testDeleteInvocation() {
+        // @formatter:off
```
nictas, line 39: Does the formatting look that bad without this comment?

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/jobs/FlowableOperationsCleanerTest.java
```diff
@@ -0,0 +1,68 @@
...
+                              Arguments.of(Collections.emptyList(), 0)
+             );
+        // @formatter:on
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void testDeleteInvocation(List<String> processIds, int deleteTimes) {
+        prepareFlowableFacade(processIds);
+        flowableOperationsCleaner.execute(null);
```
nictas, line 52: Pass some real `Date` object here and test whether it's being passed to the Flowable facade.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/jobs/FlowableOperationsCleanerTest.java
```diff
@@ -0,0 +1,68 @@
...
+        verify(flowableFacade, times(deleteTimes)).deleteProcessInstance(anyString(), anyString());
```
nictas, line 53: Make a separate verification call for each process ID in the list.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/jobs/FlowableOperationsCleanerTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.mockito.MockitoAnnotations;
+
+import com.sap.cloud.lm.sl.cf.process.flowable.FlowableFacade;
+
+public class FlowableOperationsCleanerTest {
+
+    @Mock
+    private FlowableFacade flowableFacade;
+
+    private FlowableOperationsCleaner flowableOperationsCleaner;
```
nictas, line 30: Use `@InjectMocks` instead of creating the object "manually" in the `setUp` method.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/jobs/AbortedOperationsCleanerTest.java
```diff
@@ -80,56 +78,53 @@ public void testExecute() {
...
                                               .type(EventType.ABORTED)
                                               .build();
     }
 
-    private Operation createOperation(String processId, State state) {
-        return ImmutableOperation.builder()
-                                 .processId(processId)
-                                 .state(state)
-                                 .build();
+    private class CustomOperation {
```
nictas, line 119: Don't use the term "operation". Use "process" instead.

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/jobs/AbortedOperationsCleaner.java
```diff
@@ -56,18 +51,7 @@ public void execute(Date expirationTime) {
...
-
-    private Operation getOperation(String operationId) {
-        try {
-            return operationService.createQuery()
-                                   .processId(operationId)
-                                   .singleResult();
-        } catch (NoResultException e) {
-            return null;
-        }
+        return flowableFacade.getProcessInstance(operationId) != null;
```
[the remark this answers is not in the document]
theghost5800: I had checked code in flowable and it will throw exception if result from sql query is more than 1 entry. If the result is empty, it will return null.

com.sap.cloud.lm.sl.cf.process/src/test/java/com/sap/cloud/lm/sl/cf/process/jobs/FlowableOperationsCleanerTest.java
theghost5800: Without formatting looks like this: 
`return Stream.of(Arguments.of(Arrays.asList("process-id-1"), 1), Arguments.of(Arrays.asList("process-id-1", "process-id-2"), 2),
                         Arguments.of(Collections.emptyList(), 0));`

PR #897, opened 2020-07-03
Skip user filtering if needed

IvanBorislavovDimitrov:
Do not filter by a user during health check monitoring if the user is not set.

LMCROSSITXSADEPLOY-1940

### DIFF
--- a/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/health/HealthRetriever.java
+++ b/com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/health/HealthRetriever.java
@@ -11,6 +11,7 @@ import javax.inject.Named;
 import com.sap.cloud.lm.sl.cf.core.health.model.Health;
 import com.sap.cloud.lm.sl.cf.core.health.model.HealthCheckConfiguration;
 import com.sap.cloud.lm.sl.cf.core.persistence.OrderDirection;
+import com.sap.cloud.lm.sl.cf.core.persistence.query.OperationQuery;
 import com.sap.cloud.lm.sl.cf.core.persistence.service.OperationService;
 import com.sap.cloud.lm.sl.cf.core.util.ApplicationConfiguration;
 import com.sap.cloud.lm.sl.cf.web.api.model.Operation;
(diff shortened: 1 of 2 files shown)


com.sap.cloud.lm.sl.cf.core/src/main/java/com/sap/cloud/lm/sl/cf/core/health/HealthRetriever.java
```diff
@@ -38,15 +39,21 @@ public Health getHealth() {
...
 
+    private List<Operation> getHealthCheckOperations(HealthCheckConfiguration healthCheckConfiguration, ZonedDateTime xSecondsAgo) {
+        OperationQuery healthCheckOperationsQuery = operationService.createQuery()
+                                                                    .mtaId(healthCheckConfiguration.getMtaId())
+                                                                    .spaceId(healthCheckConfiguration.getSpaceId())
+                                                                    .endedAfter(Date.from(xSecondsAgo.toInstant()))
+                                                                    .inFinalState()
+                                                                    .orderByEndTime(OrderDirection.DESCENDING);
+        if (healthCheckConfiguration.getUserName() != null) {
+            healthCheckOperationsQuery.user(healthCheckConfiguration.getUserName());
```
radoslav-d, line 54: Is there a chance that something else is null?

com.sap.cloud.lm.sl.cf.core/src/test/java/com/sap/cloud/lm/sl/cf/core/health/HealthRetrieverTest.java
```diff
@@ -101,15 +106,49 @@ public void testGetHealth() {
...
+                                                              .spaceId(SPACE_ID)
+                                                              .timeRangeInSeconds(TIME_RANGE_IN_SECONDS)
+                                                              .build());
+        Health health = healthRetriever.getHealth();
+
+        assertTrue(health.isHealthy());
+        HealthCheckOperation healthCheckOperation = health.getHealthCheckOperations()
+                                                          .get(0);
+        validateHealthCheckOperationParameters(healthCheckOperation);
+        Mockito.verify(operationQuery, times(0))
```
radoslav-d, line 128: I think that there is `never()` or something similar, which can be used instead of `times(0)`.

PR #909, opened 2020-07-22
Add metric for Flowable Queue size

theghost5800:

### DIFF
--- a/com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/monitoring/FlowableJobExecutorInformation.java
+++ b/com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/monitoring/FlowableJobExecutorInformation.java
@@ -1,0 +1,49 @@
+package com.sap.cloud.lm.sl.cf.web.monitoring;
+
+import java.time.Instant;
+
+import javax.inject.Inject;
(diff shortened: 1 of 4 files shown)


com.sap.cloud.lm.sl.cf.web/src/main/java/com/sap/cloud/lm/sl/cf/web/monitoring/FlowableJobExecutorInformation.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private boolean isPastCacheTimeout() {
+        return Instant.now()
+                      .minusSeconds(CACHE_TIMEOUT_IN_SECONDS)
+                      .isAfter(lastUpdateTime);
+    }
+
+    private void updateCurrentJobExecutorQueueSize() {
+        currentJobExecutorQueueSize = jobExecutor.getThreadPoolQueue()
```
nictas, line 40: Is this operation expensive enough to justify the cache?

theghost5800: For every execution of size() method, it obtains lock over blocking queue and this could slow add/remove jobs from queue because of frequent polling of jmx beans from Dynatrace.

PR #927, opened 2020-08-11
Replace DynatraceMeterRegistry to JMXmeterRegistry

boyan-velinov:
Jira: LMCROSSITXSADEPLOY-2178


pom.xml
```diff
@@ -310,10 +310,10 @@
                 <version>${snakeyaml.version}</version>
             </dependency>
             <dependency>
-                <groupId>io.micrometer</groupId>
```
nictas, line 313: Replace the tabs with spaces.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/MicrometerConfiguration.java
```diff
@@ -76,42 +49,20 @@ public Duration step() {
...
-    }
-
-    private String getDynatraceUrl(DynatraceServiceInfo serviceInfo) {
-        if (serviceInfo == null) {
-            return null;
-        }
-        return serviceInfo.getMicrometerUrl();
-    }
-
     private DynatraceServiceInfo getDynatraceServiceInfo() {
```
nictas, line 66: Do we need this now?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/MicrometerConfiguration.java
```diff
@@ -32,41 +32,14 @@
 
     @Inject
     @Bean
-    public DynatraceMeterRegistry dynatraceMeterRegistry(ApplicationConfiguration appConfig) {
+    public JmxMeterRegistry jmxMeterRegistry(ApplicationConfiguration appConfig) {
         DynatraceServiceInfo serviceInfo = getDynatraceServiceInfo();
```
nictas, line 36: This should no longer be needed, right?

boyan-velinov: DynatraceMeterRegistry  is removed. ReactorNetty use directly Metrics.globalRegistry that propagates meters in all registered registries.

pom.xml
boyan-velinov: Done

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/MicrometerConfiguration.java
boyan-velinov: It is not mandatory to check if DynatraceServiceInfo exists before registering JmxMeterRegistry.
However, it we check that we can optimize the things as not registering the registry at all.
In this way we will avoid unecesarry preparation of our own metrics and send them to micrometer

PR #928, opened 2020-08-12
Add ConfigurationSubscriptionCleaner

radoslav-d:
Add new clean-up job, which will delete all orphaned configuration
subscription, which were not handled by the UserDataCleaner. The cleaner
will execute once per DS application run. The cleaner queries for all
configuration subscriptions and checks each one whether its space is
existing. All configuration subscriptions with non-existence space are
deleted.
NOTE: this cleaner should be removed for the next takt.
JIRA: LMCROSSITXSADEPLOY-2130

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/jobs/ConfigurationSubscriptionCleaner.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/jobs/ConfigurationSubscriptionCleaner.java
@@ -1,0 +1,107 @@
+package org.cloudfoundry.multiapps.controller.process.jobs;
+
+import org.cloudfoundry.client.lib.CloudControllerClient;
+import org.cloudfoundry.client.lib.CloudControllerClientImpl;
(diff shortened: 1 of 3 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/jobs/ConfigurationSubscriptionCleaner.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+        try {
+            cfClient.getSpace(UUID.fromString(spaceId));
+            return true;
+        } catch (CloudOperationException e) {
+            if (e.getStatusCode()
+                 .equals(HttpStatus.NOT_FOUND)) {
+                return false;
+            }
+            throw e;
```
boyan-velinov, line 80: If one exception is thrown at the begining of stream processing, the deletion of other entries will not happen at all.
So, let's not rethrow the exception.

radoslav-d: only logging as a handling?

PR #930, opened 2020-08-13
Refactor rest api packages to match new versioning; fixes to swagger comments

ikasarov:
see ticket LMCROSSITXSADEPLOY-1714

### DIFF
--- a/multiapps-controller-api/pom.xml
+++ b/multiapps-controller-api/pom.xml
@@ -107,9 +107,8 @@
                         <configuration>
                             <apiSources>
                                 <apiSource>
-                                    <locations>org.cloudfoundry.multiapps.controller.api</locations>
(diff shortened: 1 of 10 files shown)


multiapps-controller-api/pom.xml
```diff
@@ -128,6 +130,11 @@
                                 <apiSource>
                                     <locations>org.cloudfoundry.multiapps.controller.api.v2</locations>
                                     <swaggerFileName>mtarest_v2</swaggerFileName>
+                                    <info>
```
theghost5800, line 133: Why there is no basePath field for api v2?

ikasarov: removed the other one as well, as swagger does not consider that  parameter correctly - the paths still contain the base path

PR #931, opened 2020-08-14
Change response to 404 for requests with non UUID space ID

Boris-ILIEV:


multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/security/AuthorizationChecker.java
```diff
@@ -69,6 +70,8 @@ public void ensureUserIsAuthorized(HttpServletRequest request, UserInfo userInfo
                 String message = MessageFormat.format(Messages.UNAUTHORISED_OPERATION_SPACE_ID, action, spaceGuid);
                 failWithForbiddenStatus(message);
             }
+        } catch (IllegalArgumentException e) {
```
nictas, line 73: This is not the right place for this `catch`. As it is now, it would catch **all** `IllegalArgumentException`s, whereas you only want to catch the one that may occur when parsing the space GUID. For example, it would also catch an exception coming from the following line:
```
UUID userGuid = UUID.fromString(userInfo.getId());
```
If this were to happen, the user would receive `404 Not Found`, whereas the appropriate status code would be either `400 Bad Request`, `401 Unauthorized` or `403 Forbidden`.
And this is just an example. In the future, we may add more code to the `checkPermissions` method, and it may also throw `IllegalArgumentException`s.

Furthermore, when reading this code, it's impossible to know why the `try` block would even throw such exceptions. You have to actually go inside the `checkPermissions` method to figure out what's going on, and this is never a good sign. When reading a method, a person should be able to figure out why things are the way they are, without having to dig into the lower levels of abstraction.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/security/AuthorizationChecker.java
```diff
@@ -94,8 +95,12 @@ boolean checkPermissions(UserInfo userInfo, String spaceId, boolean readOnly) {
...
             return true;
         }
-
-        UUID spaceGuid = UUID.fromString(spaceId);
+        UUID spaceGuid = null;
+        try {
+            spaceGuid = UUID.fromString(spaceId);
+        } catch (IllegalArgumentException e) {
+            throw new NotFoundException(e.getMessage(), e);
+        }
```
radoslav-d, line 103: I would suggest extracting these lines of code in separate method, because try-catch blocks are extremely ugly.
Consider something like the following example:
```
private UUID convertToUUID(String spaceId) { // you may come with better name
    try {
            return UUID.fromString(spaceId);
     } catch (IllegalArgumentException e) {
            throw new NotFoundException(e.getMessage(), e); // this exception will propagate in case it is thrown
     }
}
```

multiapps-controller-web/src/test/java/org/cloudfoundry/multiapps/controller/web/security/AuthorizationCheckerTest.java
```diff
@@ -158,6 +161,13 @@ public void testSpaceDevelopersCacheNegativeResult() {
                .getSpaceDevelopers(Mockito.eq(UUID.fromString(THIRD_SPACE_ID)));
     }
 
+    @Test
+    void testCheckPermissionsWithNonUUIDSpaceIDString() {
+        setUpMocks(true, true, null);
+
+        assertThrows(NotFoundException.class, () -> authorizationChecker.checkPermissions(userInfo, "foobar", true));
```
radoslav-d, line 168: Better replace `"foobar"` with something like `"non-uuid-space-id"`. Currently when I look at this method call, I am not sure what "foobar" stands for and I need to see the actual method to verify that the second param is spaceId.

multiapps-controller-web/src/test/java/org/cloudfoundry/multiapps/controller/web/security/AuthorizationCheckerTest.java
```diff
@@ -59,8 +62,8 @@
     private AuthorizationChecker authorizationChecker;
 
     @BeforeEach
-    public void setUp() {
-        MockitoAnnotations.initMocks(this);
+    void setUp() {
```
boyan-velinov, line 65: Why public is removed?

PR #936, opened 2020-08-20
fix for partial deploy with blue-green, LMCROSSITXSADEPLOY-2154

ikasarov:
taking selected modules '-m module-name' into consideration when renaming applications at start of blue-green strategy deply

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetectApplicationsToRenameStep.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetectApplicationsToRenameStep.java
@@ -8,6 +8,7 @@ import java.util.stream.Collectors;
 
 import javax.inject.Named;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.cloudfoundry.client.lib.CloudControllerClient;
 import org.cloudfoundry.client.lib.domain.CloudApplication;
 import org.cloudfoundry.multiapps.controller.core.model.BlueGreenApplicationNameSuffix;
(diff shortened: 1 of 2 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetectApplicationsToRenameStep.java
```diff
@@ -49,13 +51,28 @@ protected StepPhase executeStep(ProcessContext context) {
         return StepPhase.DONE;
     }
 
-    private List<String> computeOldAppsToRename(Set<String> appsToProcess) {
-        return appsToProcess.stream()
-                            .filter(appName -> !BlueGreenApplicationNameSuffix.isSuffixContainedIn(appName))
-                            .collect(Collectors.toList());
+    private List<String> computeOldAppsToRename(DeployedMta deployedMta, List<String> selectedModulesList) {        
+        Set<String> selectedModulesSet = new TreeSet<>();
```
radito3, line 55: I don't understand the need for this variable.
Just pass the list (and remove the `List` suffix from its name) to the other method

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetectApplicationsToRenameStep.java
```diff
@@ -49,13 +51,28 @@ protected StepPhase executeStep(ProcessContext context) {
...
+
+        return deployedMta.getApplications()
+                          .stream()
+                          .filter(app -> !BlueGreenApplicationNameSuffix.isSuffixContainedIn(app.getName()))
+                          .filter(app -> moduleIsSelectedForDeployment(app.getModuleName(), selectedModulesSet))
+                          .map(DeployedMtaApplication::getName)
+                          .collect(Collectors.toList());
+    }
+
+    private boolean moduleIsSelectedForDeployment(String module, Set<String> selectedModules) {
```
radito3, line 66: Wouldn't `isModuleSelectedForDeployment` be better?
And rename the first parameter to `moduleName`

radito3 (changes requested):

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetectApplicationsToRenameStep.java
ikasarov: the set will have better performance when later making the contains check

ikasarov: will remove in service of code clarity, performance isn't really a concern here

Issue #942, opened 2020-08-31
Liquibase changes should be refactored


nictas:
Most file explorers, including the GitHub UI and IDEs like Eclipse and IntelliJ, sort files alphabetically. This is how our Liquibase changelogs look when sorted like this:
```
db-changelog-add_audit_log_table.xml
db-changelog-add_configuration_registry_table.xml
db-changelog-add_configuration_subscription_table.xml
db-changelog-add_historic_operation_event_table.xml
db-changelog-add_new_configuration_registry_unique_constraint.xml
db-changelog-add_oauth_access_token_table.xml
db-changelog-add_ongoing_operation_table.xml
db-changelog-add_progress_message_table.xml
db-changelog-add_quartz_tables.xml
db-changelog-add_task_extension_table.xml
db-changelog-alter_configuration_registry_table_add_content_id_column.xml
db-changelog-alter_configuration_registry_table_add_namespace_column.xml
db-changelog-alter_configuration_registry_table_add_space_id_column.xml
db-changelog-alter_configuration_registry_table_add_target_org_column.xml
db-changelog-alter_configuration_registry_table_add_visible_in_column.xml
db-changelog-alter_configuration_registry_table_extend_provider_id_column.xml
db-changelog-alter_configuration_registry_table_extend_provider_version_column.xml
db-changelog-alter_configuration_registry_table_split_target_space_column.xml
db-changelog-alter_configuration_subscription_table_add_module_id_column.xml
db-changelog-alter_configuration_subscription_table_add_resource_id_column.xml
db-changelog-alter_ongoing_operation_table_add_final_state_column.xml
db-changelog-alter_ongoing_operation_table_rename_slp_states.xml
db-changelog-alter_operation_table_add_ended_at_column.xml
db-changelog-alter_operation_table_add_namespace_column.xml
db-changelog-alter_operation_table_drop_cleaned_up_column.xml
db-changelog-alter_operation_table_timestamp_storing_columns.xml
db-changelog-alter_progress_message_add_task_execution_id_column.xml
db-changelog-alter_progress_message_add_timestamp_column.xml
db-changelog-alter_progress_message_drop_task_execution_id_column.xml
db-changelog-alter_task_extension_add_task_execution_id_column.xml
db-changelog-delete_task_startup_progress_messages.xml
db-changelog-drop_configuration_registry_unique_constraint.xml
db-changelog-drop_context_extension_table.xml
[...]

PR #943, opened 2020-08-31
Prepare changelogs for Liquibase versions >= 3.5.2

nictas:
Since https://liquibase.jira.com/projects/CORE/issues/CORE-1863 Liquibase maps the generic BLOB type to OID instead of BYTEA for PostgreSQL databases.

This commit adapts our existing changelogs to that breaking change and also solves #942.


theghost5800 (commented):
I have one general question. With newer version of liquibase does it mean that liquibase table will be recreated/cleaned and all changes will be executed again?

multiapps-controller-core/src/main/resources/org/cloudfoundry/multiapps/controller/core/db/changelog/db-changelog.xml
```diff
@@ -4,84 +4,13 @@
...
-    <include
-        file="org/cloudfoundry/multiapps/controller/core/db/changelog/db-changelog-alter_configuration_registry_table_add_content_id_column.xml" />
-    <include
-        file="org/cloudfoundry/multiapps/controller/core/db/changelog/db-changelog-alter_configuration_subscription_table_add_module_id_column.xml" />
-    <include
-        file="org/cloudfoundry/multiapps/controller/core/db/changelog/db-changelog-alter_configuration_subscription_table_add_resource_id_column.xml" />
+    <property name="small-blob.type" value="BYTEA" dbms="postgresql" />
+    <property name="large-blob.type" value="OID" dbms="postgresql" />
+
+    <property name="small-blob.type" value="BLOB" dbms="all" />
```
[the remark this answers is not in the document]
nictas: There are unit tests that use the h2 in-memory database. They pass so I'd assume that this is fine.

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog-1.110.0-persistence.xml
```diff
@@ -0,0 +1,86 @@
...
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="process_log" />
+            </not>
+        </preConditions>
+        <createTable tableName="process_log">
+            <column name="file_id" type="CHAR(36)">
+                <constraints primaryKey="true" primaryKeyName="log_file_id_pk" nullable="false" />
+            </column>
+            <column name="content" type="${large-blob.type}" />
```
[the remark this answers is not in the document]
nictas: The goal of these changes is to refactor the structure of the changelogs and not the tables. The tables are left exactly as they were. If the `content` column is no longer needed it can be dropped in another change.

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/sql/create_unlink_file_content_function.sql
```diff
@@ -0,0 +1,11 @@
...
+$$
+  BEGIN
+    PERFORM lo_unlink(OLD.content);
+    RETURN NULL;
+  EXCEPTION
+    WHEN undefined_object THEN RAISE NOTICE 'Referenced large object could not be deleted, because it does not exist!';
+    RETURN NULL;
+  END;
+$$
+LANGUAGE 'plpgsql' STABLE;
```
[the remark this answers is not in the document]
nictas: Whatever it meant before in: https://github.com/cloudfoundry-incubator/multiapps-controller/blob/43c220afbc0596cb04870f0d0964e5e0d056f882/multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog-add_trigger_for_lm_sl_persistance_file_process_log_cf.xml#L14
There's no change here.

nictas:
No, using a newer version of Liquibase does not mean that the tables will be recreated, dropped, etc. The reason why there's such a massive refactoring in this PR is because the change of types (BLOB -> BYTEA/OID) already required the modification of existing changesets. Since we're doing it anyway, it makes sense to also refactor the structure of the changelogs, so that they're easier to manage/navigate.

multiapps-controller-core/src/main/resources/org/cloudfoundry/multiapps/controller/core/db/changelog/db-changelog.xml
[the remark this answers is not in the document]
nictas: Yup, I just tested it. Changing this line to `<property name="small-blob.type" value="BLAH" dbms="all" />` makes the tests fail.

radoslav-d (approved):

multiapps-controller-core/src/main/resources/org/cloudfoundry/multiapps/controller/core/db/changelog/db-changelog-1.110.0-core.xml
```diff
@@ -0,0 +1,265 @@
...
+                <constraints nullable="false" />
+            </column>
+            <column name="event" type="VARCHAR(100)">
+                <constraints nullable="false" />
+            </column>
+            <column name="timestamp" type="TIMESTAMP">
+                <constraints nullable="false" />
+            </column>
+        </createTable>
+        <createIndex tableName="historic_operation_event" indexName="idx_historic_operation_event_process_id ">
```
theghost5800, line 169: Previously we have added java classes which perform async index creation with `CONCURRENTLY` keyword which create index without locking the whole table. Does it mean now liquibase will use this feature of PostgreSQL to create indexes without locking whole tables?

nictas: Nope. Indexes will now be created non-concurrently, but this is fine since these changesets will only be executed if the table/index/sequence doesn't already exist.

theghost5800 (approved):

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/bootstrap/BootstrapServlet.java
```diff
@@ -110,31 +102,4 @@ private void initializeProviders() {
         AuditLoggingProvider.setFacade(new AuditLoggingFacadeSLImpl(dataSource, getUserInfoProvider()));
     }
 
-    private void executeAsyncDatabaseChanges() {
```
radito3, line 113: Now that this is gone, when will the changes be made?
Or do we even need to make these changes on app startup?

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/sql/create_unlink_file_content_function.sql
```diff
@@ -0,0 +1,11 @@
+CREATE OR REPLACE FUNCTION UNLINK_FILE_CONTENT() RETURNS TRIGGER AS
```
radito3, line 1: Since in both normal execution and in the event of an exception we return  NULL, wouldn't it be better if this is a procedure rather than a function?

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/sql/delete_unlink_file_content_triggers.sql
```diff
@@ -0,0 +1,5 @@
+DROP TRIGGER IF EXISTS DELETE_FROM_SYSTEM_TABLE_LM_SL_PERSISTANCE_FILE ON LM_SL_PERSISTENCE_FILE;
```
radito3, line 1: I understand that this file is called in case of a rollback but wouldn't it be cleaner if these DROP statemets are executed immediately prior to the trigger's creation? 
Or is it neccessary to have them deleted in a rollback?

radito3 (approved):
I greatly appreciate the deletion of the dead code and the refactoring and cleaning up of the liquibase changelog files. Just make sure you haven't missed any changelogs or deleted something you shouldn't have.

PR #966, opened 2020-09-30
Properly fix failing unit test in process module

nictas:


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/CloudPackagesGetter.java
```diff
@@ -35,7 +35,7 @@
         if (!currentCloudPackage.isPresent()) {
             return Optional.empty();
         }
-        LOGGER.info(MessageFormat.format(Messages.CURRENTLY_USED_PACKAGE_0, SecureSerialization.toJson(currentCloudPackage)));
```
nictas, line 38: Here we tried to serialize a `java.util.Optional` object, which is no longer possible with Java 11, since classes from the `java.base` module can't be accessed via reflection.

boyan-velinov (approved):

PR #967, opened 2020-10-01
Make Spring configuration resistant to renames

nictas:
This commit makes sure we don't hardcode any package/class names in our Spring configuration.


multiapps-controller-web/src/main/resources/log4j.properties
```diff
@@ -1,4 +0,0 @@
-# Root logger option
```
IvanBorislavovDimitrov, line 1: Maybe we should return this file?

IvanBorislavovDimitrov (approved):

PR #973, opened 2020-10-05
Fix database connection pool exhaustion (attempt #2)

nictas:


multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/DataSourceFactory.java
```diff
@@ -24,7 +24,6 @@ private HikariConfig createHikariConfig(CfJdbcService service, Integer maximumPo
         hikariConfig.setUsername(service.getUsername());
         hikariConfig.setPassword(service.getPassword());
         hikariConfig.setJdbcUrl(service.getJdbcUrl());
-        hikariConfig.setMaxLifetime(0);
```
nictas, line 27: Added in https://github.com/cloudfoundry-incubator/multiapps-controller/commit/3bee50c55434cf3ac3ba4c90cb6396c965f67207 as this was the major difference between the Tomcat and Hikari connection pools. As it turned out, however, this had nothing to do with the problem.

boyan-velinov (approved):

PR #981, opened 2020-10-27
Remove connection test query from data source

nictas:
As per HikariCP's recommendation (see https://github.com/brettwooldridge/HikariCP#configuration-knobs-baby).


boyan-velinov (approved):
top

PR #983, opened 2020-10-30
Namespace to routes merge

theghost5800:

### DIFF
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/helpers/SystemParameters.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/helpers/SystemParameters.java
@@ -24,8 +24,6 @@ public class SystemParameters {
     public static final String DEFAULT_IDLE_URL = "${protocol}://${default-idle-uri}";
     public static final String DEFAULT_URL = "${protocol}://${default-uri}";
 
-    private static final HostValidator HOST_VALIDATOR = new HostValidator();
-
     private final CredentialsGenerator credentialsGenerator;
     private final String targetName;
     private final String organizationName;
@@ -39,6 +37,7 @@ public class SystemParameters {
     private final String deployServiceUrl;
     private final boolean reserveTemporaryRoutes;
     private final Supplier<String> timestampSupplier;
+    private final HostValidator hostValidator;
 
     public SystemParameters(Builder builder) {
         this.targetName = builder.organizationName + " " + builder.spaceName;
@@ -54,6 +53,7 @@ public class SystemParameters {
         this.credentialsGenerator = builder.credentialsGenerator;
         this.reserveTemporaryRoutes = builder.reserveTemporaryRoutes;
         this.timestampSupplier = builder.timestampSupplier;
+        this.hostValidator = builder.hostValidator;
     }
 
     public void injectInto(DeploymentDescriptor descriptor) {
@@ -180,8 +180,8 @@ public class SystemParameters {
     private String getDefaultHost(String moduleName) {
         String host = (targetName + " " + moduleName).replaceAll("\\s", "-")
                                                      .toLowerCase();
-        if (!HOST_VALIDATOR.isValid(host, null)) {
-            return HOST_VALIDATOR.attemptToCorrect(host, null);
+        if (!hostValidator.isValid(host, Collections.emptyMap())) {
+            return hostValidator.attemptToCorrect(host, Collections.emptyMap());
         }
         return host;
     }
@@ -208,6 +208,7 @@ public class SystemParameters {
         private String deployServiceUrl;
         private boolean reserveTemporaryRoutes;
         private Supplier<String> timestampSupplier;
+        private HostValidator hostValidator;
 
         public Builder credentialsGenerator(CredentialsGenerator credentialsGenerator) {
             this.credentialsGenerator = credentialsGenerator;
@@ -266,6 +267,11 @@ public class SystemParameters {
 
         public Builder timestampSupplier(Supplier<String> timestampSupplier) {
             this.timestampSupplier = timestampSupplier;
+            return this;
+        }
+
+        public Builder hostValidator(HostValidator hostValidator) {
+            this.hostValidator = hostValidator;
             return this;
         }
 
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/HostValidator.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/HostValidator.java
@@ -1,9 +1,39 @@
 package org.cloudfoundry.multiapps.controller.core.validators.parameters;
 
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+import org.cloudfoundry.multiapps.common.util.MapUtil;
 import org.cloudfoundry.multiapps.controller.core.Messages;
 import org.cloudfoundry.multiapps.controller.core.model.SupportedParameters;
+import org.cloudfoundry.multiapps.controller.core.util.NameUtil;
 
 public class HostValidator extends RoutePartValidator {
+
+    private final String namespace;
+    private final boolean applyNamespaceGlobal;
+
+    public HostValidator() {
+        this.namespace = null;
+        this.applyNamespaceGlobal = false;
+    }
+
+    public HostValidator(String namespace, boolean applyNamespaceGlobal) {
+        this.namespace = namespace;
+        this.applyNamespaceGlobal = applyNamespaceGlobal;
+    }
+
+    @Override
+    public boolean isValid(Object routePart, final Map<String, Object> context) {
+        if (!super.isValid(routePart, context)) {
+            return false;
+        }
+        if (shouldApplyNamespace(context)) {
+            return ((String) routePart).startsWith(namespace);
+        }
+        return true;
+    }
 
     @Override
     public String getParameterName() {
@@ -28,6 +58,21 @@ public class HostValidator extends RoutePartValidator {
     @Override
     protected String getRoutePartPattern() {
         return "^([a-z0-9]|[a-z0-9][a-z0-9\\-]{0,61}[a-z0-9])|\\*$";
+    }
+
+    @Override
+    protected String modifyAndShortenRoutePart(String routePart, Map<String, Object> context) {
+        return NameUtil.computeNamespacedNameWithLength(routePart, namespace, shouldApplyNamespace(context), getRoutePartMaxLength());
+    }
+
+    private boolean shouldApplyNamespace(Map<String, Object> context) {
+        boolean applyNamespaceLocal = MapUtil.parseBooleanFlag(context, SupportedParameters.APPLY_NAMESPACE, true);
+        return applyNamespaceGlobal && applyNamespaceLocal;
+    }
+
+    @Override
+    public Set<String> getRelatedParameterNames() {
+        return Collections.singleton(SupportedParameters.APPLY_NAMESPACE);
     }
 
 }
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/IdleRoutesValidator.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/IdleRoutesValidator.java
@@ -1,0 +1,24 @@
+package org.cloudfoundry.multiapps.controller.core.validators.parameters;
+
+import java.util.Map;
+
+import org.cloudfoundry.multiapps.controller.core.model.SupportedParameters;
+
+public class IdleRoutesValidator extends RoutesValidator {
+
+    public IdleRoutesValidator(String namespace, boolean applyNamespaceGlobal) {
+        super(namespace, applyNamespaceGlobal);
+    }
+
+    @Override
+    protected void initRoutesValidators(String namespace, boolean applyNamespaceGlobal) {
+        ParameterValidator idleRouteValidator = new IdleRouteValidator(namespace, applyNamespaceGlobal);
+        this.validators = Map.of(idleRouteValidator.getParameterName(), idleRouteValidator);
+    }
+
+    @Override
+    public String getParameterName() {
+        return SupportedParameters.IDLE_ROUTES;
+    }
+
+}
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutePartValidator.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutePartValidator.java
@@ -15,17 +15,20 @@ public abstract class RoutePartValidator implements ParameterValidator {
             throw new SLException(getErrorMessage(), routePart);
         }
 
-        String result = (String) routePart;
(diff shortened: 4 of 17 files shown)


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutesValidator.java
```diff
@@ -21,15 +20,17 @@
...
-        this.validators.put(routeValidator.getParameterName(), routeValidator);
-
-        this.supportedParamsWithoutValidators = new TreeSet<>();
+    public RoutesValidator(String namespace, boolean applyNamespaceGlobal) {
+        this.validators = getRoutesValidators(namespace, applyNamespaceGlobal);
+		this.supportedParamsWithoutValidators = new TreeSet<>();
         this.supportedParamsWithoutValidators.add(SupportedParameters.NO_HOSTNAME);
     }
 
+    protected Map<String, ParameterValidator> getRoutesValidators(String namespace, boolean applyNamespaceGlobal) {
```
ikasarov, line 27: maybe rename this to initValidators instead of get, since it should be called more than once in the constructor of an instance; or just put this in the constructor and overwrite in child class

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutePartValidator.java
```diff
@@ -1,22 +1,37 @@
...
+
     @Override
     public String attemptToCorrect(Object routePart, final Map<String, Object> context) {
         if (!(routePart instanceof String)) {
             throw new SLException(getErrorMessage(), routePart);
         }
 
-        String result = (String) routePart;
-        result = NameUtil.getNameWithProperLength(result, getRoutePartMaxLength());
+        // FIXME This logic does not work in case where apply-namespace is set to false
```
ikasarov: same here

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutePartValidator.java
```diff
@@ -43,6 +27,10 @@ public String attemptToCorrect(Object routePart, final Map<String, Object> conte
         return result;
     }
 
+    protected String getRouteName(Object routePart, Map<String, Object> context) {
```
ikasarov, line 30: does this need to be in a separate method? it seems to only hide what the method is actually doing with the generic 'getRouteName'

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutesValidator.java
```diff
@@ -74,22 +73,42 @@ public Object attemptToCorrect(Object routes, final Map<String, Object> context)
...
 
-    private Map<String, Object> attemptToCorrectParameterMap(Map<String, Object> originalElem, Map<String, Object> context) {
+    private Map<String, Object> getRelatedParametersForContext(Map<String, Object> routesElement, Map<String, Object> context) {
+        Map<String, Object> updatedContext = new HashMap<>(routesElement);
+        Set<String> relatedParameters = getRelatedParameterNames();
+        Map<String, Object> updatedParamters = relatedParameters.stream()
+                                                                .filter(relatedParameter -> doesParameterMissingInRoutesElement(relatedParameter,
+                                                                                                                                routesElement,
+                                                                                                                                context))
+                                                                .collect(Collectors.toMap(Function.identity(), context::get));
```
ikasarov, line 88: is it guaranteed that these parameters are in the context? this way of processing does not specify a default value if they are not present.

Also this whole method is a bit hard to understand. Maybe also name 'updatedParameters' something more clear e.g. globalParameterValues

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutesValidator.java
```diff
@@ -74,22 +73,42 @@ public Object attemptToCorrect(Object routes, final Map<String, Object> context)
...
+        updatedContext.putAll(updatedParamters);
+        return updatedContext;
+    }
+
+    private boolean doesParameterMissingInRoutesElement(String relatedParameter, Map<String, Object> routesElement,
```
ikasarov, line 93: rename this method to isParameterMissingInRoutes or maybe even isGlobalParameterValueOverridden

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutesValidator.java
theghost5800: If make it as first suggestion, I have to move initialization of validators in method but then I have to remove final modifier. In the second option you cannot override parent constructor because child class always calls parent constructor and then initialize itself.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutePartValidator.java
theghost5800: Actually this is fixed with next commit "Add namespace to routes parameters only"

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutePartValidator.java
theghost5800: In the frist place I placed the logic about namespace in `RoutePartValidator` class but then I realized that namespace is used only on `HostValidator` implementation, so I put it there. So If remove `getRouteName` method, I have to override the whole `attemptToCorrect` method and execute almost the same with only difference calling another method from `NameUtil` class.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutesValidator.java
theghost5800: About the first question the answer is yes, in `isGlobalParameterValueOverridden` method there is a check does context contains this parameter key. I have tried to reproduce scenario where the key is present but the value is null but this is not possible because when in yaml set value of key to null, then it is ignored.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutesValidator.java
ikasarov: Maybe just rename and either remove final or keep final and setting outside then. Can always keep it as it is, I just find the 'get' word to be confusing in this case.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutePartValidator.java
ikasarov: yeah, forgot to delete this comment as I saw the other commit afterwards

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutePartValidator.java
ikasarov: ah, I see; still, maybe make method name more descriptive i.e. modifyAndShortenRoutePart or similar;
also, just realized that
result = result.replaceAll("^(-*)", "");
result = result.replaceAll("(-*)$", "");
this code should probably happen before that method invocation, as it also has the potential of shortening the value and possibly bringing it down to acceptable size.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutesValidator.java
ikasarov: yes

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/helpers/SystemParameters.java
```diff
@@ -54,6 +55,9 @@ public SystemParameters(Builder builder) {
         this.credentialsGenerator = builder.credentialsGenerator;
         this.reserveTemporaryRoutes = builder.reserveTemporaryRoutes;
         this.timestampSupplier = builder.timestampSupplier;
+        this.namespace = builder.namespace;
+        this.applyNamespace = builder.applyNamespace;
+        this.hostValidator = new HostValidator(namespace, applyNamespace);
```
IvanBorislavovDimitrov, line 60: Can we move this in the Builder class?

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/HostValidator.java
```diff
@@ -1,15 +1,39 @@
...
+    }
+
+    @Override
+    public boolean isValid(Object routePart, final Map<String, Object> context) {
+        boolean validRoutePart = super.isValid(routePart, context);
+        if (!validRoutePart) {
+            return false;
+        }
+        if (shouldApplyNamespace(context)) {
+            return ((String) routePart).startsWith(namespace);
```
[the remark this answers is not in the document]
theghost5800: Why? StringUtils.startsWith require again to cast Object to String. The only difference which I see is that if `namespace` is null, it will not throw NPE but this scenario shouldn't be possible to occur.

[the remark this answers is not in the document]
IvanBorislavovDimitrov: I agree, it cannot be null, but it just seems a bit more error-prone without a null check but ok

IvanBorislavovDimitrov (approved):

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/IdleRoutesValidator.java
```diff
@@ -0,0 +1,24 @@
+package org.cloudfoundry.multiapps.controller.core.validators.parameters;
+
+import java.util.Map;
+
+import org.cloudfoundry.multiapps.controller.core.model.SupportedParameters;
+
+public class IdleRoutesValidator extends RoutesValidator {
+
+    public IdleRoutesValidator(String namespace, boolean applyNamespaceGlobal) {
```
ikasarov, line 9: Is this constructor even needed now?

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RoutesValidator.java
```diff
@@ -71,22 +74,42 @@ public Object attemptToCorrect(Object routes, final Map<String, Object> context)
...
+        Map<String, Object> globalParametersValues = relatedParameters.stream()
+                                                                      .filter(relatedParameter -> isGlobalParameterValueOverridden(relatedParameter,
+                                                                                                                                   routeElement,
+                                                                                                                                   context))
+                                                                      .collect(Collectors.toMap(Function.identity(), context::get));
+        updatedAllRouteElements.putAll(globalParametersValues);
+        return updatedAllRouteElements;
+    }
+
+    private boolean isGlobalParameterValueOverridden(String relatedParameter, Map<String, Object> routeElement,
```
ikasarov, line 94: doesn't this check the opposite of what the method name implies? my mistake - you can negate the check or name it isGlobalParameterValueInAction or isGlobalParameterValueNotOverridden :D

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/IdleRoutesValidator.java
theghost5800: Yes, because it is necessary to initialize parent constructor

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/CollectSystemParametersStep.java
```diff
@@ -111,6 +113,8 @@ private SystemParameters createSystemParameters(ProcessContext context, CloudCon
                                              .reserveTemporaryRoutes(reserveTemporaryRoutes)
                                              .credentialsGenerator(credentialsGeneratorSupplier.get())
                                              .timestampSupplier(timestampSupplier)
+                                             .namespace(namespace)
+                                             .applyNamespace(applyNamespace)
```
IvanBorislavovDimitrov, line 117: Do we really need namespace and applyNamespace in the Builder class? Is it possible to have a method in CollectSystemParametersStep which creates HostValidator and pass it to the builder? This way the builder won't need to create the HostValidator it will be injected instead.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/RouteValidator.java
```diff
@@ -17,8 +17,8 @@
 
     private final List<ParameterValidator> validators;
 
-    public RouteValidator() {
-        this.validators = Arrays.asList(new HostValidator(), new DomainValidator());
+    public RouteValidator(String namespace, boolean applyNamespaceGlobal) {
+        this.validators = Arrays.asList(new HostValidator(namespace, applyNamespaceGlobal), new DomainValidator());
```
IvanBorislavovDimitrov, line 21: List.of() ?

multiapps-controller-core/src/test/java/org/cloudfoundry/multiapps/controller/core/helpers/SystemParametersTest.java
```diff
@@ -54,26 +61,18 @@ void initialize() throws Exception {
...
-
-    @Test
-    void testGeneralParametersWithReserveTemporaryRoutes() throws Exception {
-        testGeneralParameters(true);
-    }
-
-    private void testGeneralParameters(Boolean reserveTemporaryRoutes) throws Exception {
-        SystemParameters testedClass = createSystemParameters(reserveTemporaryRoutes);
+    @ParameterizedTest
+    @ValueSource(booleans = { false, true })
```
IvanBorislavovDimitrov, line 65: Cool!

multiapps-controller-core/src/test/java/org/cloudfoundry/multiapps/controller/core/helpers/SystemParametersTest.java
```diff
@@ -194,6 +198,12 @@ private void verifyModuleParameters(String moduleName, Map<String, Object> modul
             assertFalse(moduleParameters.containsKey(SupportedParameters.IDLE_DOMAIN));
         }
 
+        if (applyNamespace) {
```
IvanBorislavovDimitrov, line 201: Can we use assertEquals instead of if/else?

IvanBorislavovDimitrov (approved):

PR #990, opened 2020-11-09
Add functionality to upload files without temp file

radito3:
#### Description: 
When making API calls from our multiapps cli, we can pass the file size in a custom header.
That way, we don't need a temporary file to calculate size and digest.
The digest can be calculated by passing a `DigestInputStream` to the underlying file storage when uploading.


nictas (approved):

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/services/DatabaseFileService.java
```diff
@@ -65,17 +65,17 @@ public int deleteFilesEntriesWithoutContent() throws FileStorageException {
...
-    private boolean storeFileWithContent(FileEntry fileEntry, InputStream fileStream) throws FileStorageException {
+    protected FileEntry storeFile(FileEntry fileEntry, InputStream fileStream) throws FileStorageException {
         try {
-            return getSqlQueryExecutor().execute(getSqlFileQueryProvider().getStoreFileQuery(fileEntry, fileStream));
+            if (fileEntry.getDigest() != null) {
+                getSqlQueryExecutor().execute(getSqlFileQueryProvider().getStoreFileQuery(fileEntry, fileStream));
+                return fileEntry;
+            }
+            String digest = getSqlQueryExecutor().execute(getSqlFileQueryProvider().getStoreFileWithoutDigestQuery(fileEntry, fileStream));
+            getSqlQueryExecutor().execute(getSqlFileQueryProvider().getUpdateFileDigestQuery(fileEntry, digest));
```
nictas, line 75: Is this in the same transaction and if not, can this be a problem? For example, having two people upload the same file in parallel. The first stores the content in the database successfully, but before updating the digest, the second person tries to retrieve the list of files to check whether he can skip the upload. He sees the file entry, tries to use the digest and fails with a nil pointer dereference, because it is nil.

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog-1.110.0-persistence.xml
```diff
@@ -15,9 +15,7 @@
                 <constraints primaryKey="true" primaryKeyName="file_id_pk" nullable="false" />
             </column>
             <column name="content" type="${large-blob.type}" />
-            <column name="digest" type="VARCHAR(255)">
-                <constraints nullable="false" />
-            </column>
+            <column name="digest" type="VARCHAR(255)" />
```
nictas, line 18: You can't do this, as it will break the migration on existing databases. Modifying an existing Liquibase changelog is very dangerous - see https://www.liquibase.org/blog/dealing-with-changing-changesets. You need to create a new changelog - `db-changelog-1.115.0-persistence.xml` and use the following change there:
https://docs.liquibase.com/change-types/community/drop-not-null-constraint.html

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/sql/create_unlink_file_content_triggers.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TRIGGER DELETE_FROM_SYSTEM_TABLE_LM_SL_PERSISTENCE_FILE AFTER UPDATE OR DELETE
+CREATE TRIGGER DELETE_FROM_SYSTEM_TABLE_LM_SL_PERSISTENCE_FILE AFTER DELETE
```
nictas, line 1: The same comment applies here. You can't modify existing Liquibase changes like this - you need to create new ones.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -87,7 +87,13 @@
             }
 
             try (InputStream in = item.openStream()) {
-                FileEntry entry = fileService.addFile(spaceGuid, namespace, item.getName(), in);
+                FileEntry entry;
+                if (request.getHeader("X-File-Size") != null) {
+                    entry = fileService.addFile(spaceGuid, namespace, item.getName(), in,
+                                                Long.parseLong(request.getHeader("X-File-Size")));
```
nictas, line 93: What if some moron passes a non-integer here? Will this be handled gracefully by our exception handlers?

nictas (changes requested):

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/services/DatabaseFileService.java
radito3: You're right. It's currently not in a single transaction but I will change it to be.

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog-1.116.0-persistence.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.6.xsd">
+
+    <changeSet author="sap.com" id="alter_table_lm_sl_persistence_file">
+        <preConditions onFail="MARK_RAN">
```
nictas, line 8: The `db-changelog-1.110.0-xxx.xml` files are full of preconditions, because they replace previous Liquibase changelogs and they must be skipped in case their previous versions were already executed. However, this is a brand new change and it should always be executed, so that the `lm_sl_persistence_file` table is in a consistent state across all landscapes.

Remove this precondition and let the change fail if there are any problems.

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog-1.116.0-persistence.xml
```diff
@@ -0,0 +1,25 @@
...
+            <tableExists tableName="lm_sl_persistence_file" />
+        </preConditions>
+        <dropNotNullConstraint tableName="lm_sl_persistence_file" columnName="digest" />
+    </changeSet>
+
+    <changeSet author="sap.com" id="create_new_unlink_file_content_trigger">
+        <preConditions onFail="MARK_RAN">
```
nictas, line 15: Same.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -87,7 +88,16 @@
             }
 
             try (InputStream in = item.openStream()) {
-                FileEntry entry = fileService.addFile(spaceGuid, namespace, item.getName(), in);
+                FileEntry entry;
+                String fileSizeHeader = request.getHeader("X-File-Size");
```
nictas, line 92: You could extract this and the following 3 lines of code in a separate method called `getFileSizeHeaderValue` and let it return a Long instead of a String.

nictas (approved):

PR #1002, opened 2020-12-01
Add option to download MTAR from url

radito3:
#### Description: 
Extends FilesApi and its service delegate to handle a file url query parameter.

#### Issue: LMCROSSITXSADEPLOY-1981

### DIFF
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/FileUtils.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/FileUtils.java
@@ -12,6 +12,7 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.text.MessageFormat;
 
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.cloudfoundry.multiapps.controller.core.Messages;
 import org.slf4j.Logger;
 
@@ -19,6 +20,7 @@ public class FileUtils {
 
     public static final String PATH_SHOULD_NOT_BE_ABSOLUTE = "Archive entry name \"{0}\" should not be absolute";
     public static final String PATH_SHOULD_BE_NORMALIZED = "Archive entry name \"{0}\" should be normalized";
+    public static final String FILE_DOES_NOT_HAVE_EXTENSION = "File \"{0}\" does not have extension";
 
     private FileUtils() {
     }
@@ -47,6 +49,13 @@ public class FileUtils {
         if (Paths.get(path)
                  .isAbsolute()) {
             throw new IllegalArgumentException(MessageFormat.format(PATH_SHOULD_NOT_BE_ABSOLUTE, path));
+        }
+    }
+
+    public static void validateFileHasExtension(String fileName) {
+        String extension = FilenameUtils.getExtension(fileName);
+        if (StringUtils.isEmpty(extension)) {
+            throw new IllegalArgumentException(MessageFormat.format(FILE_DOES_NOT_HAVE_EXTENSION, fileName));
         }
     }
 
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/UriUtil.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/UriUtil.java
@@ -1,7 +1,10 @@
 package org.cloudfoundry.multiapps.controller.core.util;
 
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.text.MessageFormat;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -46,4 +49,13 @@ public class UriUtil {
                      .map(CloudRouteSummary::toUriString)
                      .collect(Collectors.joining(", "));
     }
+
+    public static void validateUrl(String url) {
+        try {
+            new URL(url).toURI();
+        } catch (URISyntaxException | MalformedURLException e) {
+            throw new IllegalArgumentException(MessageFormat.format(Messages.INVALID_URL, url), e);
+        }
+    }
+
 }
--- a/multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
+++ b/multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
@@ -2,7 +2,16 @@ package org.cloudfoundry.multiapps.controller.web.api.impl;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
+import java.net.URLDecoder;
+import java.net.http.HttpClient;
+import java.net.http.HttpClient.Redirect;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.net.http.HttpResponse.BodyHandlers;
+import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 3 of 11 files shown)


multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -138,4 +154,55 @@ private FileMetadata parseFileEntry(FileEntry fileEntry) {
                                     .namespace(fileEntry.getNamespace())
                                     .build();
     }
+
+    private FileEntry uploadFileFromUrl(String spaceGuid, String namespace, String mtarEndpoint) throws FileStorageException, IOException {
+        if (!UriUtil.isValidUrl(mtarEndpoint)) {
+            throw new SLException("Invalid URL");
```
nictas, line 160: Add the URL to the exception and make sure that the response contains the error message, so that users will be able to see what went wrong.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/UriUtil.java
```diff
@@ -46,4 +48,14 @@ public static String prettyPrintRoutes(Set<CloudRouteSummary> routes) {
                      .map(CloudRouteSummary::toUriString)
                      .collect(Collectors.joining(", "));
     }
+
+    public static boolean isValidUrl(String url) {
```
nictas, line 52: Wouldn't it be better to call the method `validateUrl` and throw an exception if the URl is invalid, so that the error message shown to the user contains some additional details?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -138,4 +154,55 @@ private FileMetadata parseFileEntry(FileEntry fileEntry) {
...
+        }
+        String decodedUrl = URLDecoder.decode(mtarEndpoint, StandardCharsets.UTF_8);
+        WebClient client = new RestUtil().createWebClient(false);
```
nictas, line 163: Not sure if this is a good idea. `RestUtil` is more of an internal class for the client facade. It'd be nice to have the option of changing it whenever we want, without having to touch the deploy-service codebase.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -138,4 +154,55 @@ private FileMetadata parseFileEntry(FileEntry fileEntry) {
...
+
+        ClientResponse response = client.get()
+                                        .uri(decodedUrl)
+                                        .exchange()
+                                        .block();
+        long fileSize = response.headers()
+                                .contentLength()
+                                .orElseThrow(() -> new SLException("MTAR endpoint did not return Content-Length header"));
+        long maxUploadSize = new Configuration().getMaxUploadSize();
+        if (fileSize > maxUploadSize) {
```
nictas, line 173: Is this the only validation for downloading a bigger MTAR than allowed? From a security PoV, the server could return a Content-Length header with a small value, but proceed to send an enormous MTAR.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/FileUtils.java
```diff
@@ -72,6 +72,16 @@ public static void cleanUp(Path filePath, Logger logger) {
         }
     }
 
+    public static boolean isFilenameValid(String file) {
+        File f = new File(file);
+        try {
+            f.getCanonicalPath();
```
nictas, line 78: When will this throw an exception? Also, isn't there anything here that would do the job:
https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FilenameUtils.html

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -138,4 +154,55 @@ private FileMetadata parseFileEntry(FileEntry fileEntry) {
...
+        }
+
+        String fileName = extractFileName(decodedUrl);
+        try (InputStream content = response.body(this::extractContent)) {
+            return fileService.addFile(spaceGuid, namespace, fileName, content, fileSize);
+        }
+    }
+
+    private String extractFileName(String url) {
+        int fragmentIdx = url.indexOf('#');
```
nictas, line 184: Why not:
```
String[] pathFragments = new URI(uri).getPath().split("/");
// Get the last element of pathFragments
```

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -138,4 +154,55 @@ private FileMetadata parseFileEntry(FileEntry fileEntry) {
...
+        }
+        int startIdx = url.lastIndexOf('/') + 1;
+        String fileName = url.substring(startIdx);
+        if (!FileUtils.isFilenameValid(fileName)) {
+            throw new SLException("Invalid MTAR name");
+        }
+        return fileName;
+    }
+
+    private InputStream extractContent(ReactiveHttpInputMessage httpResponse, BodyExtractor.Context context) {
```
nictas, line 200: The `context` doesn't seem to be used here.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
```diff
@@ -138,4 +154,55 @@ private FileMetadata parseFileEntry(FileEntry fileEntry) {
...
+        return httpResponse.getBody()
+                           .map(dataBuffer -> dataBuffer.asInputStream(true))
+                           .collectList()
+                           .map(list -> new SequenceInputStream(Collections.enumeration(list)))
```
nictas, line 204: I think this will create a SequenceInputStream from multiple ByteArrayInputStreams, which means that the entire file content will be stored in-memory. Can you check, please? If I'm right, find another way to download the file.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
radito3: Yes, but we need it to use this method as a method reference

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
radito3: Agreed. 
Changed to `WebClientFactory::getWebClient()` which is in the deploy-service codebase

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
radito3: Done as suggested

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/FilesApiServiceImpl.java
radito3: Added a `LimitedInputStream` wrapper to the content stream

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/FileUtils.java
radito3: Changed to `FilenameUtils::validatePath`

nictas (approved):

PR #1006, opened 2020-12-07
Applications v3

IvanBorislavovDimitrov:
LMCROSSITXSADEPLOY-1749

### DIFF
--- a/multiapps-controller-process/src/main/java/module-info.java
+++ b/multiapps-controller-process/src/main/java/module-info.java
@@ -50,6 +50,8 @@ open module org.cloudfoundry.multiapps.controller.process {
     requires spring.core;
     requires spring.security.oauth2;
     requires spring.web;
+    requires reactor.netty;
+    requires io.netty.handler;
 
     requires static java.compiler;
     requires static org.immutables.value;
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStep.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStep.java
@@ -8,6 +8,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
 import javax.inject.Inject;
@@ -63,17 +64,21 @@ public class UploadAppStep extends TimeoutAsyncFlowableStep {
 
         MtaArchiveElements mtaArchiveElements = context.getVariable(Variables.MTA_ARCHIVE_ELEMENTS);
         String moduleFileName = mtaArchiveElements.getModuleFileName(applicationToProcess.getModuleName());
+        CloudControllerClient client = context.getControllerClient();
         if (moduleFileName == null) {
             getStepLogger().debug(Messages.NO_CONTENT_TO_UPLOAD);
+            if (applicationToProcess.getDockerInfo() != null) {
+                CloudPackage cloudPackage = createDockerPackage(client, applicationToProcess);
+                context.setVariable(Variables.CLOUD_PACKAGE, cloudPackage);
+            }
             return StepPhase.DONE;
         }
 
         String newApplicationDigest = getNewApplicationDigest(context, context.getRequiredVariable(Variables.APP_ARCHIVE_ID),
                                                               moduleFileName);
-        CloudControllerClient client = context.getControllerClient();
         CloudApplication cloudApp = client.getApplication(applicationToProcess.getName());
 
-        boolean contentChanged = detectApplicationFileDigestChanges(context, cloudApp, newApplicationDigest);
+        boolean contentChanged = detectApplicationFileDigestChanges(cloudApp, newApplicationDigest);
         if (contentChanged) {
             proceedWithUpload(context, applicationToProcess, moduleFileName);
             attemptToUpdateApplicationDigest(context.getControllerClient(), cloudApp, newApplicationDigest);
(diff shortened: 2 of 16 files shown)


radoslav-d (approved):
Everything seems right

radoslav-d (approved):

multiapps-controller-process/src/main/java/module-info.java
```diff
@@ -50,6 +50,8 @@
     requires spring.core;
     requires spring.security.oauth2;
     requires spring.web;
+    requires reactor.netty;
+    requires io.netty.handler;
```
radito3, line 54: Why are these needed?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStep.java
```diff
@@ -94,6 +99,34 @@ public StepPhase executeAsyncStep(ProcessContext context) throws FileStorageExce
         return StepPhase.DONE;
     }
 
+    private CloudPackage createDockerPackage(CloudControllerClient client, CloudApplicationExtended application) {
+        UUID applicationGuid = client.getApplicationGuid(application.getName());
+        return client.createDockerPackage(applicationGuid, application.getDockerInfo());
+    }
+
+    private String getNewApplicationDigest(ProcessContext context, String appArchiveId, String fileName) throws FileStorageException {
+        return fileService.processFileContent(context.getVariable(Variables.SPACE_GUID), appArchiveId,
```
radito3, line 108: Can `appArchiveId` be retrieved in this method instead of passing it as a parameter?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStep.java
```diff
@@ -63,17 +64,21 @@ public StepPhase executeAsyncStep(ProcessContext context) throws FileStorageExce
 
         MtaArchiveElements mtaArchiveElements = context.getVariable(Variables.MTA_ARCHIVE_ELEMENTS);
         String moduleFileName = mtaArchiveElements.getModuleFileName(applicationToProcess.getModuleName());
+        CloudControllerClient client = context.getControllerClient();
         if (moduleFileName == null) {
             getStepLogger().debug(Messages.NO_CONTENT_TO_UPLOAD);
+            if (applicationToProcess.getDockerInfo() != null) {
```
radito3, line 70: If I understand correctly, this flow means that there isn't a file entry in the MTAR associated with this module because it is described as a docker application, is this right?
If so, how did we operate with docker apps so far?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStep.java
```diff
@@ -94,6 +99,34 @@ public StepPhase executeAsyncStep(ProcessContext context) throws FileStorageExce
         return StepPhase.DONE;
     }
 
+    private CloudPackage createDockerPackage(CloudControllerClient client, CloudApplicationExtended application) {
+        UUID applicationGuid = client.getApplicationGuid(application.getName());
```
radito3, line 103: Why isn't the guid in the `CloudApplicationExtended` object used but a request to the CC is made to fetch the guid?

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/CreateOrUpdateAppStepTest.java
```diff
@@ -92,16 +92,19 @@ void testCreateApplicationFromDockerImage() {
                                                                                           .password("somePassword")
                                                                                           .build())
                                                    .build();
+        Staging staging = ImmutableStaging.builder()
```
radito3, line 95: Maybe `dockerStaging` would be more descriptive

multiapps-controller-process/src/main/java/module-info.java
IvanBorislavovDimitrov: Error occurred during initialization of boot layer
java.lang.module.FindException: Module reactor.netty not found, required by com.sap.cloudfoundry.client.facade
They are required as far as I understood

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStep.java
IvanBorislavovDimitrov: Yes, it can.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStep.java
IvanBorislavovDimitrov: In v2 the docker apps do not require packages, they are just created as docker image apps and started. In v3 it is required to create a docker package to deploy the application. If there is no module name the application is probably a docker app but I added this check just in case if the user does not provide docker parameters otherwise a null pointer will occur.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStep.java
IvanBorislavovDimitrov: Because it set in BuildApplicationDeployModelStep and its guid is unknown -> the id is null.

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/CreateOrUpdateAppStepTest.java
IvanBorislavovDimitrov: Done

radito3 (approved):

PR #1007, opened 2020-12-16
SLP accepts org/space names with slash

boyan-velinov:
This is a regression on our side for SLP.
In the past, this scenario was working. Most probably,
the regiression comes from migration to WebMVC from Jersey

Jira: LMCROSSITXSADEPLOY-2291


multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/WebMvcConfiguration.java
```diff
@@ -39,6 +40,9 @@ public void addInterceptors(InterceptorRegistry registry) {
     @Override
     public void configurePathMatch(PathMatchConfigurer configurer) {
         configurer.setUseSuffixPatternMatch(false);
+        UrlPathHelper urlPathHelper = new UrlPathHelper();
+        urlPathHelper.setUrlDecode(false);
```
theghost5800, line 44: It should be check does handle correctly plus(+) signs in org/space names. We had previously another issues related with url decode/encode.

boyan-velinov: It accepts + sign. This is covered by integration tests.

theghost5800 (approved):

PR #1012, opened 2021-01-13
Do not fail on error when generating javadoc

radoslav-d:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -941,7 +941,10 @@
                         <configuration>
                             <source>8</source>
                             <detectJavaApiLink>false</detectJavaApiLink>
-                            <additionalparam>-Xdoclint:none</additionalparam>
+                            <doclint>none</doclint>
+                            <!--TODO remove once the javadoc generation is stable-->
+                            <!--Added because failing javadoc generation blocks the deploy to nexus-->
+                            <failOnError>false</failOnError>
                         </configuration>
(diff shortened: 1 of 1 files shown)


nictas (approved):

pom.xml
```diff
@@ -941,7 +941,8 @@
                         <configuration>
                             <source>8</source>
```
theghost5800, line 942: Is this version of Java? Why we will ignore failure while generating javadoc?

radoslav-d: The maven-javadoc-plugin has a regression when compiling with java 11 and fails with unknown error during deploy to nexus. The solution, which the community proposes, is to use java 8, which worked for us quite a long time. However, during release procedure, the javadoc generation failed again and again even after trying numerous combinations of configurations. Me and Sunny decided on setting `<failOnError>false</failOnError>` for now because javadoc is not vital part of our release.

theghost5800: okay

theghost5800 (approved):

pom.xml
theghost5800: Maybe some comment should be added with reason why this is configured like that. Probably in future we should delete this configuration.

radoslav-d: Yes, for sure. This is temp workaround to execute the deployment to nexus.

PR #1013, opened 2021-01-14
Remove dynamic operation state calculation

radoslav-d:
Rely on the current state, which is persisted in the database instead of
dynamically calculating it. Map current_state in the database to
Operation.state and remove Operation.cachedState.
JIRA: LMCROSSITXSADEPLOY-2128

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/OperationsHelper.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/OperationsHelper.java
@@ -1,27 +1,20 @@
 package org.cloudfoundry.multiapps.controller.process.util;
 
-import java.text.MessageFormat;
 import java.util.List;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.inject.Named;
 
-import org.apache.commons.collections4.CollectionUtils;
 import org.cloudfoundry.multiapps.controller.api.model.ErrorType;
 import org.cloudfoundry.multiapps.controller.api.model.ImmutableOperation;
 import org.cloudfoundry.multiapps.controller.api.model.Operation;
 import org.cloudfoundry.multiapps.controller.persistence.model.HistoricOperationEvent;
 import org.cloudfoundry.multiapps.controller.persistence.services.OperationService;
-import org.cloudfoundry.multiapps.controller.process.Messages;
 import org.cloudfoundry.multiapps.controller.process.metadata.ProcessTypeToOperationMetadataMapper;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 @Named
 public class OperationsHelper {
-
-    private static final Logger LOGGER = LoggerFactory.getLogger(OperationsHelper.class);
 
     private final OperationService operationService;
     private final ProcessTypeToOperationMetadataMapper metadataMapper;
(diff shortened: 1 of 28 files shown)


multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog-1.117.0-core.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        logicalFilePath="org/cloudfoundry/multiapps/controller/core/db/changelog/db-changelog-1.117.0-core.xml"
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.6.xsd">
+
+    <changeSet author="sap.com" id="alter_operation_table_add_not_null_constraint_current_state">
```
boyan-velinov, line 8: Did you check that there is not NULLs at the canary/live landscapes?

In addition, I would refrain in the current tact to introduce the change, if we revert the release.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/OperationsHelper.java
```diff
@@ -67,52 +56,4 @@ private ErrorType toErrorType(HistoricOperationEvent.EventType historicType) {
         }
         return null;
     }
-
-    public Operation addState(Operation operation) {
-        if (operation.getState() != null) {
-            return operation;
-        }
-        Operation.State state = computeState(operation);
-        // Fixes bug XSBUG-2035: Inconsistency in 'operation', 'act_hi_procinst' and 'act_ru_execution' tables
```
ikasarov, line 76: was this code used to release a lock on every operation state poll? will the mentioned inconsistency come up again with it deleted?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/OperationsApiServiceImpl.java
```diff
@@ -214,26 +209,13 @@ private Operation getOperation(String operationId) {
             Operation operation = operationService.createQuery()
                                                   .processId(operationId)
                                                   .singleResult();
-            operation = addOperationState(operation);
             operation = operationsHelper.addErrorType(operation);
-            return operation;
+            return operationsHelper.releaseLockIfNeeded(operation);
```
theghost5800, line 213: Will this lead to more db queries when we get operation?

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog-1.119.0-core.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.6.xsd">
+
+    <changeSet author="sap.com" id="alter_operation_table_add_not_null_constraint_current_state">
+        <addNotNullConstraint tableName="operation" columnName="current_state" columnDataType="VARCHAR(50)"/>
```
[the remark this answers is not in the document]
boyan-velinov: Agree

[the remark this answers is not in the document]
radoslav-d: It should not be a problem. When I went through kibana logs, there was only one peek of logs, where the current state was null and that correlated with the lss script. There were no further logs, where the current state was null. However, we may skip this liquibase change and add it in the next execution of this task (when the final state is removed).

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/api/impl/OperationsApiServiceImpl.java
radoslav-d: Will lead to 1 more db query only in case the lock is not released and the operation is in final state. Apparently there was/is an inconsistency in flowable tables, which this method fixes.

multiapps-controller-persistence/src/main/resources/org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog-1.119.0-core.xml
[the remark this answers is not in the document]
radoslav-d: I have removed it for now in order to add it in the next execution.

theghost5800 (approved):

PR #1020, opened 2021-01-28
Response timeout

theghost5800:


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/ApplicationConfiguration.java
```diff
@@ -130,6 +131,7 @@
     public static final int DEFAULT_CONTROLLER_CLIENT_CONNECT_TIMEOUT_IN_SECONDS = (int) TimeUnit.MINUTES.toSeconds(2);
     public static final int DEFAULT_CONTROLLER_CLIENT_CONNECTION_POOL_SIZE = 192;
     public static final int DEFAULT_CONTROLLER_CLIENT_THREAD_POOL_SIZE = 64;
+    public static final int DEFAULT_CONTROLLER_CLIENT_RESPONSE_TIMEOUT_IN_SECONDS = (int) TimeUnit.MINUTES.toSeconds(20);
```
radoslav-d, line 134: What is the reasoning behind these 20 minutes? Isn`t this too much?

radoslav-d (approved):

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/ApplicationConfiguration.java
theghost5800: Yeah, the value is little bit high but I don't want to hit some timeout issue cause of slow response from CF controller. We had previously issues where deployments in big spaces with applications and services exceed some timeouts and that's why I want to guarantee that it won't break deployments in big cf spaces. I just want to have some timeout because by default reactor-netty doesn't have any and this potentially leads to bigger issues.

boyan-velinov (approved):

boyan-velinov (approved):

PR #1030, opened 2021-02-23
Do not delete discontinued services in CTS deploy

boyan-velinov:
Jira: LMCROSSITXSADEPLOY-2337


theghost5800 (approved):
It looks like but it should be tested what will happen when `deleteServices` variable is not set(has null value) and flowable try to resolve the expression

PR #1032, opened 2021-02-25
Do not set error state if operation is aborted

radoslav-d:
Flowable call the ErrorProcessListener after FailedJobCommandFactory.
Our FailedJobCommandFactory calls AbortProcessAction (which sets state
to aborted) in the case where --abort-on-error is set. In other words,
the process is initially aborted in case --abort-on-error and then the
ErrorProcessListener is called.
JIRA: LMCROSSITXSADEPLOY-2128

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/OperationInFinalStateHandler.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/OperationInFinalStateHandler.java
@@ -90,6 +90,9 @@ public class OperationInFinalStateHandler {
         Operation operation = operationService.createQuery()
                                               .processId(processInstanceId)
                                               .singleResult();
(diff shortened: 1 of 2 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/OperationInFinalStateHandler.java
```diff
@@ -90,6 +90,9 @@ protected void setOperationState(String processInstanceId, Operation.State state
         Operation operation = operationService.createQuery()
                                               .processId(processInstanceId)
                                               .singleResult();
+        if (isOperationAlreadyFinal(operation)) {
```
radoslav-d, line 93: This change is not related to the problem, but rather fixes the issue where endedAt is set again to set, which is already aborted.

ikasarov (approved):

theghost5800 (approved):

PR #1036, opened 2021-03-01
Print operation ID

nictas:


radoslav-d (approved):
Have you tested what is printed in other cases: abort, resume, undeploy, etc? Because as I remember, there might be duplications. For example:
```
Executing action 'abort' on operation e7c6860f-7aa2-11eb-b8ab-eeee0a99d2e1...
Operation ID: e7c6860f-7aa2-11eb-b8ab-eeee0a99d2e1
```

nictas:
I've tested deploy, blue-green deploy and undeploy. I'll test abort as well, but I doubt it'll print anything considering that the progress message is saved in `StartProcessListener`.

nictas:
Confirmed: The message is not printed in case or `retry`, `abort` or `resume`.

PR #1052, opened 2021-04-02
Add correlation id header to CC client

radoslav-d:
JIRA: LMCROSSITXSADEPLOY-1673


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ClientReleaser.java
```diff
@@ -27,10 +27,13 @@ public void releaseClientFor(HistoryService historyService, String processInstan
         String organizationName = HistoryUtil.getVariableValue(historyService, processInstanceId, Variables.ORGANIZATION_NAME.getName());
         String spaceName = HistoryUtil.getVariableValue(historyService, processInstanceId, Variables.SPACE_NAME.getName());
         String spaceGuid = HistoryUtil.getVariableValue(historyService, processInstanceId, Variables.SPACE_GUID.getName());
+        String correlationId = HistoryUtil.getVariableValue(historyService, processInstanceId, Variables.CORRELATION_ID.getName());
 
         try {
-            clientProvider.releaseClient(user, organizationName, spaceName);
-            clientProvider.releaseClient(user, spaceGuid);
+            clientProvider.releaseClient(user, organizationName, spaceName, null);
```
boyan-velinov, line 33: I think, that correlationid should be not null at this point

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
```diff
@@ -123,28 +112,19 @@ private OAuth2AccessTokenWithAdditionalInfo getValidToken(String userName) {
...
 
-    private CloudControllerClient getClientFromCache(String userName, String spaceId) {
-        String key = getKey(userName, spaceId);
-        return clients.computeIfAbsent(key, k -> clientFactory.createClient(getValidToken(userName), spaceId));
+    private CloudControllerClient getClientFromCache(String userName, String spaceId, String correlationId) {
+        String key = getKey(userName, spaceId, correlationId);
+        return clients.computeIfAbsent(key, k -> clientFactory.createClient(getValidToken(userName), spaceId, correlationId));
     }
 
     private String getKey(String... args) {
```
boyan-velinov, line 125: What is the advantage of having correlation id in the key at all?
I see 2 disadvantages:
1.  In this way the key become non-deterministic. E.g:
(user_X, my-guid, null) AND (user_X, null, my-guid) will produce the same result 
2. the cache could have more elements, potentially harder to manage

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
```diff
@@ -35,44 +38,28 @@
...
+    public CloudControllerClient getControllerClient(String userName, String org, String space, String correlationId) {
         try {
-            return getClientFromCache(userName, org, space, processId);
+            return getClientFromCache(userName, org, space, correlationId);
         } catch (CloudOperationException e) {
             throw new SLException(e, Messages.CANT_CREATE_CLIENT_2, org, space);
         }
     }
 
-    /**
```
boyan-velinov, line 49: I do not see advantage of the removal. In this way it is expected you force the java client code to pass null values, which should be handled afterwards.
Check https://stackoverflow.com/questions/4535112/is-null-arguments-a-bad-practice

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ClientReleaser.java
```diff
@@ -27,10 +27,13 @@ public void releaseClientFor(HistoryService historyService, String processInstan
...
+            clientProvider.releaseClient(user, spaceGuid, null);
```
boyan-velinov, line 34: Can we reduce somehow the invocation of all these functions? I would be easy to forget what and when to release

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/ClientFactory.java
```diff
@@ -10,19 +10,19 @@ public CloudControllerClient createClient(OAuth2AccessTokenWithAdditionalInfo to
         return createClient(createCredentials(token));
     }
 
-    public CloudControllerClient createClient(OAuth2AccessTokenWithAdditionalInfo token, String org, String space) {
-        return createClient(createCredentials(token), org, space);
+    public CloudControllerClient createClient(OAuth2AccessTokenWithAdditionalInfo token, String org, String space, String correlationId) {
```
boyan-velinov, line 13: Can we prepare some builder where some of the values are required and others, as correlationid, not?
The direction where any new argument is added to all methods is not good.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ClientReleaser.java
radoslav-d: It can be null if the request to CC is not performed from process i.e. `cf mtas`.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
radoslav-d: Because the client can be created with correlation id or not - imagine the following example:
a user is performing parallel `cf deploy`s in the same space, but for different mta. For each of the deployments, a client will be cached with the correlation id of the process. If the correlation id is not taken into account when managing the clients the might be a process, which uses a client with different correlation id. In fact for a space-user there will be only one client and therefore this feature becomes useless.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
radoslav-d: I will think of better way to do what is intended.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/ClientFactory.java
radoslav-d: will think of a way

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
boyan-velinov: This is the case with the cached token so far and there are not problems.

theghost5800: Currently cache is based on username,org,space or username and spaceguid. Now if we want to tag requests for single process, we need to add in cache CloudControllerClient for every single process, not per space. You can set additional http headers when you create cloudfoundry client from os project.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudFoundryClientFactory.java
```diff
@@ -46,35 +49,44 @@ private ImmutableCloudControllerRestClientFactory createClientFactory(Applicatio
     @Override
     protected CloudControllerClient createClient(CloudCredentials credentials) {
         OAuthClient oAuthClient = oAuthClientFactory.createOAuthClient();
-        List<ExchangeFilterFunction> exchangeFilters = getExchangeFiltersList(null, null);
+        List<ExchangeFilterFunction> exchangeFilters = getExchangeFiltersList(null, null,null);
```
theghost5800, line 52: Now I am thinking that this was not good idea to add just another null. Maybe we have to create separate method without parameters.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ClientReleaser.java
theghost5800: The idea behind this was just to try to release clients created during polling of operation

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/resources/ConfigurationEntriesResource.java
```diff
@@ -37,7 +37,7 @@
     @PostMapping("/purge")
     public ResponseEntity<Void> purgeConfigurationRegistry(@RequestParam(REQUEST_PARAM_ORGANIZATION) String organization,
                                                            @RequestParam(REQUEST_PARAM_SPACE) String space) {
-        CloudControllerClient client = createClient(organization, space);
+        CloudControllerClient client = createClient();
```
theghost5800, line 40: Why you don't use organization and space to create client?

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
```diff
@@ -21,8 +18,10 @@
 @Named
 public class CloudControllerClientProvider {
 
+    private static final String NO_CORRELATION_ID = "NO-ID"; // TODO decide better dummy value
```
theghost5800, line 21: You will use this constant only to build key which will be used only in cache map. It won't be used to add correlation id header with this constant, right?

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientFactory.java
```diff
@@ -46,51 +47,60 @@ private ImmutableCloudControllerRestClientFactory createClientFactory(Applicatio
                                                         .build();
     }
 
-    @Override
-    protected CloudControllerClient createClient(CloudCredentials credentials) {
+    public CloudControllerClient createClient(OAuth2AccessTokenWithAdditionalInfo token) {
```
theghost5800, line 50: I think it is possible in CloudCredentials to set target of org and space. How this will be possible now?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/resources/ConfigurationEntriesResource.java
radoslav-d: Firstly because I wanted to remove the factory method for (user, org, space). It is used only here. Secondly getting client for the concrete space is not needed here, as the spaceid is retrieved again later when purging the config entries. Thirdly, one call to the CC could be skipped here, because creating client with org and space name involves additional call to obtain the space guid.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
radoslav-d: Yes, that is right. You may propose better value if anything comes to your mind.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientFactory.java
radoslav-d: I removed the methods with `CloudCredentials` parameter, because they were not used actually. But could be easily added again if we decide.

boyan-velinov (approved):

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
theghost5800: I am fine with this value

theghost5800 (approved):

PR #1054, opened 2021-04-06
Fix client creation for config registry purge

radoslav-d:
JIRA: LMCROSSITXSADEPLOY-1673


multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/resources/ConfigurationEntriesResource.java
```diff
@@ -37,7 +37,7 @@
     @PostMapping("/purge")
     public ResponseEntity<Void> purgeConfigurationRegistry(@RequestParam(REQUEST_PARAM_ORGANIZATION) String organization,
                                                            @RequestParam(REQUEST_PARAM_SPACE) String space) {
-        CloudControllerClient client = createClient();
+        CloudControllerClient client = createClient(organization, space);
```
theghost5800, line 40: So, organization and space is necessary to create client?

radoslav-d: It is actually not necessary in order to create a client, but is mandatory if you want to target a space. Actually the targeting happens during client creation.

theghost5800 (approved):

theghost5800 (approved):

PR #1062, opened 2021-05-13
Get recent logs with time frame

radoslav-d:
JIRA: LMCROSSITXSADEPLOY-2084

### DIFF
--- a/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
+++ b/com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
@@ -41,27 +41,28 @@ public class PollStageAppStatusExecution implements AsyncExecution {
 
         try {
             execution.getStepLogger()
-                .debug(Messages.CHECKING_APP_STATUS, app.getName());
+                     .debug(Messages.CHECKING_APP_STATUS, app.getName());
             
             StagingState state = applicationStager.getStagingState(execution, client);
      
             
             ProcessLoggerProvider processLoggerProvider = execution.getStepLogger()
-                .getProcessLoggerProvider();
+                                                                   .getProcessLoggerProvider();
             StepsUtil.saveAppLogs(execution.getContext(), client, recentLogsRetriever, app, LOGGER, processLoggerProvider);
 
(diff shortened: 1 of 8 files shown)


com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
```diff
@@ -41,67 +41,70 @@ public AsyncExecutionState execute(ExecutionWrapper execution) {
...
         if (state.getStagingLogs() != null) {
             StagingLogs logs = state.getStagingLogs();
             String stagingLogs = logs.getLogs();
             int offset = logs.getOffset();
             
             stagingLogs = new XMLValueFilter(stagingLogs).getFiltered();
             execution.getStepLogger()
-                .debug(stagingLogs);
+                     .debug(stagingLogs);
+            StepsUtil.appLog(execution.getContext(), appName, stagingLogs, LOGGER, processLoggerProvider);
```
theghost5800, line 104: Is this about add staging logs into application log file?

pom.xml
```diff
@@ -58,7 +58,7 @@
         <jackson-asl.version>1.9.13</jackson-asl.version>
         <liquibase.version>3.4.1</liquibase.version>
         <liquibase-slf4j.version>1.2.1</liquibase-slf4j.version>
-        <cloudfoundry.client.version>1.3.0</cloudfoundry.client.version>
+        <cloudfoundry.client.version>1.3.1-SNAPSHOT</cloudfoundry.client.version>
```
theghost5800, line 61: Do not forget to adopt release version before merge

theghost5800 (approved):

com.sap.cloud.lm.sl.cf.process/src/main/java/com/sap/cloud/lm/sl/cf/process/steps/PollStageAppStatusExecution.java
radoslav-d: yes, and actually this whole if is applicable only to xs because there is no staging logs in the state object for cf

pom.xml
radoslav-d: yes, that is why I left this with `SNAPSHOT` suffix

theghost5800 (approved):

PR #1065, opened 2021-06-02
Refactor step timeout handling & update timeouts

radito3:
#### Description: 

Change `getTimeout` method to return Duration instead of a simple integer.
This will prevent time unit inconsistencies.

### DIFF
--- a/multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStepTest.java
+++ b/multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStepTest.java
@@ -2,6 +2,7 @@ package org.cloudfoundry.multiapps.controller.process.steps;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import java.time.Duration;
 import java.util.Map;
 
 import org.cloudfoundry.multiapps.controller.client.lib.domain.CloudApplicationExtended;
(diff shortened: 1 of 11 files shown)


multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStepTest.java
```diff
@@ -49,8 +50,9 @@ void testGetTimeoutWithAppParameter() {
         private void testGetTimeout(CloudApplicationExtended app, int expectedUploadTimeout) {
             context.setVariable(Variables.APP_TO_PROCESS, app);
 
-            int uploadTimeout = step.getTimeout(context);
-            assertEquals(expectedUploadTimeout, uploadTimeout);
+            var expectedTimeout = Duration.ofSeconds(expectedUploadTimeout);
+            var actualTimeout = step.getTimeout(context);
```
IvanBorislavovDimitrov, line 54: I've seen that Sonar forces us to use var but do we really want it?

IvanBorislavovDimitrov (approved):

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/UploadAppStepTest.java
radito3: Maybe not in every instance Sonar suggests but I think this is appropriate.

ikasarov (approved):

PR #1078, opened 2021-07-26
Throws exception on zero declared instances

vkalapov:

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ScaleAppStep.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ScaleAppStep.java
@@ -27,7 +27,8 @@ public class ScaleAppStep extends SyncFlowableStep {
         CloudControllerClient client = context.getControllerClient();
 
         String appName = app.getName();
-        Integer instances = (app.getInstances() != 0) ? app.getInstances() : null;
+
+        Integer instances = app.getInstances();
 
         if (instances != null && (existingApp == null || !instances.equals(existingApp.getInstances()))) {
             getStepLogger().info(Messages.SCALING_APP_0_TO_X_INSTANCES, appName, instances);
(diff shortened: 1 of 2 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ScaleAppStep.java
```diff
@@ -27,7 +28,12 @@ protected StepPhase executeStep(ProcessContext context) {
         CloudControllerClient client = context.getControllerClient();
 
         String appName = app.getName();
-        Integer instances = (app.getInstances() != 0) ? app.getInstances() : null;
+
+        if (app.getInstances() == 0) {
```
theghost5800, line 32: Did you try to deploy mta withhout specified instances parameter? What will happen if instances parameter is negative value?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ScaleAppStep.java
```diff
@@ -27,7 +28,12 @@ protected StepPhase executeStep(ProcessContext context) {
...
+            throw new SLException(Messages.ERROR_INVALID_ZERO_INSTANCE_COUNT);
+        }
+
+        Integer instances = app.getInstances();
```
theghost5800, line 36: I think you can leave this line before if statement because you are using it there

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ScaleAppStep.java
vkalapov: The deployment of mta with no specified instances parameter is tested and uses the default instance count. In the case of the instances parameter being a negative value, a SLException is thrown.

vkalapov:
The current version of cf cli allows apps with 0 instances to be started and no longer fails when attempting to push with 0 instances specified. The deploy process behavior is inconsistent with this behavior.

theghost5800 (approved):

PR #1099, opened 2021-08-16
Shallow cloning on travis

YunLemon:
According to [https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth](url), Travis CI provide a way to shallow clone a repository. This has the obvious benefit of speed, since you only need to download a small number of commits.


theghost5800:
Hi @YunLemon ,
We are configure it in this way because it is necessary to pick all commits for sonar scan https://github.com/cloudfoundry-incubator/multiapps-controller/commit/d5c0015c521086b5ff39c58d5a8bf2315b30b9ef#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485  
Is there are any other reasons to propose this change?

YunLemon:
no other reasons. I understand, thank you.

theghost5800:
Hi @YunLemon ,
What is the reason for reopening of this pr?

PR #1107, opened 2021-09-24
Fix some sonar issues & update UAA client to use v3 info endpoint

radito3:


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/listeners/StartProcessListener.java
```diff
@@ -45,11 +45,11 @@
     @Inject
     protected ProcessTypeParser processTypeParser;
     @Autowired(required = false)
-    private ProcessTypeToOperationMetadataMapper operationMetadataMapper;
+    private transient ProcessTypeToOperationMetadataMapper operationMetadataMapper;
```
IvanBorislavovDimitrov, line 48: why transient is needed at all? and why only to operationMetadataMapper and currentTimeSupplier

IvanBorislavovDimitrov (approved):

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/listeners/StartProcessListener.java
radito3: Don't know. Sonar complains about only these two

PR #1129, opened 2021-12-15
Increasing versions for log4j and jclouds. Explicitly setting version of gson

abdermendz:
In order to fix some vulnerabilities increasing versions of the following was required:
- jclouds
- gson
- log4j

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,7 @@
         <jclouds.version>2.3.0</jclouds.version>
         <guava.version>30.1.1-jre</guava.version>
         <guice.version>5.0.1</guice.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -770,6 +771,12 @@
                 <artifactId>azureblob</artifactId>
                 <version>${jclouds.version}</version>
             </dependency>
+            <!-- https://mvnrepository.com/artifact/com.google.code.gson/gson -->
```
theghost5800, line 774: Add comment for this transitive dependency. Similar to this one
``

theghost5800 (changes requested):

theghost5800 (approved):

PR #1145, opened 2022-01-11
Fix deletion of service keys for empty key list

vkalapov:

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetermineServiceCreateUpdateServiceActionsStep.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetermineServiceCreateUpdateServiceActionsStep.java
@@ -47,7 +47,6 @@ public class DetermineServiceCreateUpdateServiceActionsStep extends SyncFlowable
     protected StepPhase executeStep(ProcessContext context) throws Exception {
         CloudControllerClient client = context.getControllerClient();
         CloudServiceInstanceExtended serviceToProcess = context.getVariable(Variables.SERVICE_TO_PROCESS);
-
         getStepLogger().info(Messages.PROCESSING_SERVICE, serviceToProcess.getName());
         CloudServiceInstance existingService = client.getServiceInstance(serviceToProcess.getName(), false);
         setServiceParameters(context, serviceToProcess);
@@ -85,10 +84,8 @@ public class DetermineServiceCreateUpdateServiceActionsStep extends SyncFlowable
 
     private List<ServiceAction> determineActions(ProcessContext context, CloudServiceInstanceExtended service,
                                                  CloudServiceInstance existingService) {
-        Map<String, List<CloudServiceKey>> serviceKeys = context.getVariable(Variables.SERVICE_KEYS_TO_CREATE);
         List<ServiceAction> actions = new ArrayList<>();
-        List<CloudServiceKey> keys = serviceKeys.get(service.getResourceName());
-        if (shouldUpdateKeys(service, keys)) {
+        if (shouldUpdateKeys(service, existingService, context)) {
             getStepLogger().debug("Service keys should be updated");
             actions.add(ServiceAction.UPDATE_KEYS);
         }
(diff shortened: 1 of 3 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetermineServiceCreateUpdateServiceActionsStep.java
```diff
@@ -224,8 +224,12 @@ private CloudServiceInstanceExtended mergeCredentials(CloudServiceInstanceExtend
                                                     .withCredentials(result);
     }
 
-    private boolean shouldUpdateKeys(CloudServiceInstanceExtended service, List<CloudServiceKey> serviceKeys) {
-        return !(service.isUserProvided() || CollectionUtils.isEmpty(serviceKeys));
+    private boolean shouldUpdateKeys(CloudServiceInstanceExtended service, List<CloudServiceKey> serviceKeys, ProcessContext context) {
+        return !(service.isUserProvided() || CollectionUtils.isEmpty(serviceKeys) && shouldNotDeleteServiceKeys(context));
```
radito3, line 228: Due to the moderate complexity of this statement, it's a bit hard to discern the result.
It would be clearer if the 3rd expression is outside the initial parenthesis.
As far as I understand, right now, this would return true (as in, the keys should be updated) when the service is not user-provided or has at least one service key and its service keys _should_ be deleted (because the new function returns true if service keys should not be deleted and we are inverting that result).
Is this the correct behaviour?

vkalapov: I agree that the statement is rather complicated and I will try to simplify it with your suggestion.  
As for the behaviour, the method would currently return true (meaning the keys should be updated) when the service is not user provided and either there is at least one service key or service keys should be deleted.

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/DetermineServiceCreateUpdateServiceActionsStepTest.java
```diff
@@ -99,6 +100,27 @@ void testServiceParametersFetchingErrorHandling(boolean isOptionalService) {
...
+                   .thenReturn(service);
+        }
+        step.execute(execution);
+        assertStepFinishedSuccessfully();
+        List<ServiceAction> serviceActionsToExecute = context.getVariable(Variables.SERVICE_ACTIONS_TO_EXCECUTE);
+        if (serviceExists) {
+            assertTrue(serviceActionsToExecute.contains(ServiceAction.UPDATE_KEYS), "Actions should contain " + ServiceAction.UPDATE_KEYS);
+            return;
+        }
+        assertFalse(serviceActionsToExecute.contains(ServiceAction.UPDATE_KEYS), "Actions should not contain " + ServiceAction.UPDATE_KEYS);
```
radito3, line 121: Add some empty lines to create a semantic separation: setup - execution - validation.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/DetermineServiceCreateUpdateServiceActionsStep.java
```diff
@@ -224,8 +221,24 @@ private CloudServiceInstanceExtended mergeCredentials(CloudServiceInstanceExtend
...
-    private boolean shouldUpdateKeys(CloudServiceInstanceExtended service, List<CloudServiceKey> serviceKeys) {
-        return !(service.isUserProvided() || CollectionUtils.isEmpty(serviceKeys));
+    private boolean shouldUpdateKeys(CloudServiceInstanceExtended service, CloudServiceInstance existingService, ProcessContext context) {
+        if (service.isUserProvided()) {
+            return false;
+        }
+        return serviceHasServiceKeysToCreate(context, service) || shouldDeleteServiceKeys(existingService, context);
+    }
+
+    private boolean serviceHasServiceKeysToCreate(ProcessContext context, CloudServiceInstanceExtended service) {
```
radito3, line 231: I don't think we need "service" as a prefix here.
You can omit it and name the method `hasServiceKeysForCreation`.

radito3 (approved):

PR #1179, opened 2022-04-08
Add configurable delay after stopping app

vkalapov:
LMCROSSITXSADEPLOY-2577

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/StopAppStep.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/StopAppStep.java
@@ -10,6 +10,7 @@ import javax.inject.Named;
 import org.cloudfoundry.multiapps.controller.api.model.ProcessType;
 import org.cloudfoundry.multiapps.controller.core.model.HookPhase;
 import org.cloudfoundry.multiapps.controller.process.Messages;
+import org.cloudfoundry.multiapps.controller.process.util.ApplicationWaitAfterStopHandler;
 import org.cloudfoundry.multiapps.controller.process.util.ProcessTypeParser;
 import org.cloudfoundry.multiapps.controller.process.variables.Variables;
 import org.springframework.beans.factory.config.BeanDefinition;
@@ -26,25 +27,25 @@ public class StopAppStep extends SyncFlowableStepWithHooks implements BeforeStep
     @Inject
     private ProcessTypeParser processTypeParser;
 
+    @Inject
+    private ApplicationWaitAfterStopHandler waitAfterStopHandler;
+
     @Override
     public StepPhase executeStepInternal(ProcessContext context) {
-        // Get the next cloud application from the context
+
         CloudApplication app = context.getVariable(Variables.APP_TO_PROCESS);
 
-        // Get the existing application from the context
         CloudApplication existingApp = context.getVariable(Variables.EXISTING_APP);
-
         if (existingApp != null && !existingApp.getState()
                                                .equals(State.STOPPED)) {
             getStepLogger().info(Messages.STOPPING_APP, app.getName());
 
-            // Get a cloud foundry client
             CloudControllerClient client = context.getControllerClient();
 
-            // Stop the application
             client.stopApplication(app.getName());
 
             getStepLogger().debug(Messages.APP_STOPPED, app.getName());
+            waitAfterStopHandler.configureDelayAfterAppStop(context, app.getName());
         } else {
             getStepLogger().debug("Application \"{0}\" already stopped", app.getName());
         }
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopHandler.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopHandler.java
@@ -1,0 +1,47 @@
+package org.cloudfoundry.multiapps.controller.process.util;
+
+import java.time.Duration;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import org.cloudfoundry.multiapps.controller.process.Messages;
+import org.cloudfoundry.multiapps.controller.process.steps.ProcessContext;
+import org.cloudfoundry.multiapps.controller.process.variables.Variables;
+
+@Named
+public class ApplicationWaitAfterStopHandler {
+
+    protected ApplicationWaitAfterStopVariableGetter delayVariableGetter;
+
+    @Inject
+    public ApplicationWaitAfterStopHandler(ApplicationWaitAfterStopVariableGetter delayVariableGetter) {
+        this.delayVariableGetter = delayVariableGetter;
+    }
+
+    public void configureDelayAfterAppStop(ProcessContext context, String appName) {
+        boolean shouldWaitAfterAppStop = shouldWaitAfterAppStop(context);
+        if (shouldWaitAfterAppStop) {
+            setWaitAfterAppStopVariable(context, appName);
+            context.getStepLogger()
+                   .info(Messages.DELAYING_APP_0_FOR_1_SECONDS, appName, context.getVariable(Variables.DELAY_AFTER_APP_STOP)
+                                                                                .toSeconds());
+        }
+    }
+
+    private boolean shouldWaitAfterAppStop(ProcessContext context) {
+        if (!delayVariableGetter.isAppStopDelayEnvVariableSet(context)) {
+            return false;
+        }
+        Duration durationToWait = delayVariableGetter.getDelayDurationFromAppEnv(context);
+        return !durationToWait.isZero();
+    }
+
+    private void setWaitAfterAppStopVariable(ProcessContext context, String appName) {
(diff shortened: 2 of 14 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopVariableGetter.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private static Duration convertDelayToSeconds(int delay, ProcessContext context) {
+        if (delay > ApplicationConfiguration.DEFAULT_MAX_STOP_DELAY_IN_SECONDS) {
+            context.getStepLogger()
+                   .warnWithoutProgressMessage(Messages.DELAY_AFTER_APP_STOP_0_ABOVE_MAX_VALUE, delay);
+            return Duration.ofSeconds(ApplicationConfiguration.DEFAULT_MAX_STOP_DELAY_IN_SECONDS);
+        }
+        Duration delayAfterAppStop = Duration.ofSeconds(delay);
+        return delayAfterAppStop.isNegative() ? Duration.ZERO : delayAfterAppStop;
```
IvanBorislavovDimitrov, line 53: I'd also warn if the value is negative

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopVariableGetter.java
```diff
@@ -0,0 +1,59 @@
...
+        return processDurationFromAppEnv(stopDelayFromEnv, context);
+    }
+
+    private static Duration processDurationFromAppEnv(String stopDelayFromEnv, ProcessContext context) {
+        try {
+            int delay = Integer.parseInt(stopDelayFromEnv);
+            return convertDelayToSeconds(delay, context);
+        } catch (NumberFormatException e) {
+            context.getStepLogger()
+                   .debug(Messages.ERROR_WHILE_READING_APP_STOP_DELAY, e.getMessage());
```
IvanBorislavovDimitrov, line 41: Warn or Error?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/StopAppStep.java
```diff
@@ -42,8 +43,7 @@ public StepPhase executeStepInternal(ProcessContext context) {
             CloudControllerClient client = context.getControllerClient();
 
             // Stop the application
```
IvanBorislavovDimitrov, line 45: Delet dis :D

IvanBorislavovDimitrov (changes requested):
Do not use so many static methods. Create them as beans and inject them. Static code cannot be mocked/tested well.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopVariableGetter.java
vkalapov: I think that changing the debug to warn might be more appropriate. In addition I'll rename the message constant name from "ERROR_..." to something else to avoid confusion.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationStopHandler.java
```diff
@@ -0,0 +1,43 @@
...
+        this.delayVariableGetter = delayVariableGetter;
+    }
+
+    public void stopAppWithDelayIfConfigured(CloudApplication app, CloudControllerClient client, ProcessContext context) {
+        client.stopApplication(app.getName());
+        if (shouldWaitAfterAppStop(context)) {
+            Duration durationToWait = delayVariableGetter.getDelayDurationFromAppEnv(context);
+            context.getStepLogger()
+                   .info(Messages.WAITING_FOR_0_SECONDS, durationToWait.toSeconds());
+            MiscUtil.sleep(durationToWait.toMillis());
```
theghost5800, line 31: One other approach for this is using flowable entities like [Timer](https://www.flowable.com/open-source/docs/bpmn/ch07b-BPMN-Constructs#timer-intermediate-catching-event) in bpmn diagram to wait some time before to continue with next flowable steps. In this way flowable step won't be in "busy" state and hold db transaction until Thread sleep finish.
At some point it's good that we have some restriction about maximum delay time because if we don't have such configuration, some client can make attack by applying very long stop delays and trigger multiple deployments which will consume all available resources.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopVariableGetter.java
```diff
@@ -0,0 +1,76 @@
...
+        return getWaitAfterStopFromApp(existingApp);
+    }
+
+    private String getWaitAfterStopFromApp(CloudApplication app) {
+        if (ObjectUtils.isEmpty(app)) {
+            return null;
+        }
+        return app.getEnv()
+                  .get(Constants.VAR_WAIT_AFTER_APP_STOP);
+    }
```
IvanBorislavovDimitrov, line 40: Utilize usage of Optional<> instead of null checks. Return Optional instead of null

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopHandler.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public boolean shouldWaitAfterAppStop(ProcessContext context) {
+        if (!delayVariableGetter.isAppStopDelayEnvVariableSet(context)) {
+            return false;
+        }
+        Duration durationToWait = delayVariableGetter.getDelayDurationFromAppEnv(context);
+        return !durationToWait.isZero();
+    }
+
+    public void setDelayAfterAppStop(ProcessContext context, String appName) {
```
theghost5800, line 31: This method is quite strange because it is named with `set` prefix and I am expecting to see to pass delay value to the method. Either change the name of the method or pass delay value from outside the method as parameter.

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopHandlerTest.java
```diff
@@ -0,0 +1,109 @@
...
+    void testDurationToDelayIsZero() {
+        Mockito.when(delayVariableGetter.isAppStopDelayEnvVariableSet(context))
+               .thenReturn(true);
+        Mockito.when(delayVariableGetter.getDelayDurationFromAppEnv(context))
+               .thenReturn(Duration.ofSeconds(0));
+
+        assertFalse(stopHandler.shouldWaitAfterAppStop(context));
+    }
+
+    @Test
```
theghost5800, line 59: These two tests can be also combined into parameterized test

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopHandler.java
```diff
@@ -0,0 +1,52 @@
...
+            setWaitAfterAppStopVariable(context, appName);
+            context.getStepLogger()
+                   .info(Messages.DELAYING_APP_0_FOR_1_SECONDS, appName, context.getVariable(Variables.DELAY_AFTER_APP_STOP)
+                                                                                           .toSeconds());
+        }
+    }
+
+    private void setWaitAfterAppStopVariable(ProcessContext context, String appName) {
+        Duration durationToWait = delayVariableGetter.getDelayDurationFromAppEnv(context);
+        Duration currentDurationToWait = context.getVariable(Variables.DELAY_AFTER_APP_STOP);
```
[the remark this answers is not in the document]
vkalapov: No, unless this method has been invoked from a previous process or operation, since this method is the only place that the variable is set currently.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationWaitAfterStopHandler.java
```diff
@@ -0,0 +1,52 @@
...
+public class ApplicationWaitAfterStopHandler {
+
+    protected ApplicationWaitAfterStopVariableGetter delayVariableGetter;
+
+    @Inject
+    public ApplicationWaitAfterStopHandler(ApplicationWaitAfterStopVariableGetter delayVariableGetter) {
+        this.delayVariableGetter = delayVariableGetter;
+    }
+
+    public boolean shouldWaitAfterAppStop(ProcessContext context) {
```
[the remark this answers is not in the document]
vkalapov: No, the public scope is a leftover from previous changes and is only used for unit testing.

theghost5800 (approved):

PR #1190, opened 2022-04-27
Adopt cf-java-client-sap 2.17.0

radito3:


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/clients/ApplicationRoutesGetter.java
```diff
@@ -16,24 +18,64 @@ public ApplicationRoutesGetter(CloudControllerClient client) {
...
-        return "/v2/apps/" + appGuid.toString() + "/routes?inline-relations-depth=1";
+        return "/v3/routes?include=domain&app_guids=" + appGuid.toString();
+    }
+
+    private String getServiceRouteBindingsUrl(String routeGuids) {
+        return "/v3/service_route_bindings?route_guids=" + routeGuids;
+    }
+
+    private String getRouteGuidsAsString(List<Map<String, Object>> routeResources) {
+        return routeResources.stream()
```
theghost5800, line 45: We have to add check for length of route guids because there is length limit for url/query parameters. We have already such logic for batching in CloudControllerRestClientImp class -> `getApplicationResourcesByNamesInBatches`

theghost5800 (approved):

PR #1193, opened 2022-05-19
Prevent concurrently running opperations on one MTA

radito3:
LMCROSSITXSADEPLOY-2574


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ProcessConflictPreventer.java
```diff
@@ -70,7 +81,7 @@ private List<Operation> findConflictingOperations(String mtaId, String namespace
                                .mtaId(mtaId)
                                .namespace(namespace)
                                .spaceId(spaceId)
-                               .acquiredLock(true)
+                               .orderByStartTime(OrderDirection.DESCENDING)
```
theghost5800, line 84: Why acquired lock check is done now in the code?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ProcessConflictPreventer.java
```diff
@@ -45,17 +46,27 @@ public synchronized void acquireLock(String mtaId, String namespace, String spac
...
 
-        if (conflictingOperations.size() == 1) {
+        if (conflictingOperations.stream()
+                                 .anyMatch(Operation::hasAcquiredLock)) {
+            if (conflictingOperations.stream()
+                                     .filter(Operation::hasAcquiredLock)
+                                     .count() >= 2) {
+                List<String> operationIds = getOperationIds(conflictingOperations);
+                throw new SLException(Messages.MULTIPLE_OPERATIONS_WITH_LOCK_FOUND, qualifiedMtaId, spaceId, operationIds);
+            }
```
theghost5800, line 60: Extract this in new method

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ProcessConflictPreventer.java
```diff
@@ -45,17 +46,27 @@ public synchronized void acquireLock(String mtaId, String namespace, String spac
...
             Operation conflictingOperation = conflictingOperations.get(0);
             throw new SLException(Messages.CONFLICTING_PROCESS_FOUND, conflictingOperation.getProcessId(), qualifiedMtaId);
         }
-        if (conflictingOperations.size() >= 2) {
-            List<String> operationIds = getOperationIds(conflictingOperations);
-            throw new SLException(Messages.MULTIPLE_OPERATIONS_WITH_LOCK_FOUND, qualifiedMtaId, spaceId, operationIds);
+
+        if (!conflictingOperations.isEmpty()) {
+            var latestOperation = conflictingOperations.get(0);
```
theghost5800, line 66: Did you consider/make test with SLP scenario where you can trigger multiple mta deployments with single process?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ProcessConflictPreventer.java
radito3: If one process hangs prior to the step that sets this lock and another process is started in that time, there would be 2 processes with `hasAcquiredLock` -> false. The current query didn't handle that scenario, that's why we need to manually check the lock.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ProcessConflictPreventer.java
radito3: Would they be concurrent for the same MTA?

theghost5800 (approved):

PR #1204, opened 2022-06-24
Adopt cf-java-client-sap

IvanBorislavovDimitrov:
LMCROSSITXSADEPLOY-2547

Adapt interfaces


radito3 (approved):
Don't forget to bump cf java client version

PR #1222, opened 2022-08-30
Fix enablement of objectstore for Alicloud

theghost5800:
JIRA:LMCROSSITXSADEPLOY-2132


multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/bootstrap/BootstrapServlet.java
```diff
@@ -77,12 +77,15 @@ private void initializeApplicationConfiguration() {
...
-            LOGGER.info(MessageFormat.format(Messages.FILE_SERVICE_DELETED_FILES, deletedFiles));
-        } catch (FileStorageException e) {
-            LOGGER.error(MessageFormat.format(Messages.FILE_SERVICE_CLEANUP_FAILED, e.getMessage()), e);
-        }
+        new Thread(() -> {
+            try {
+                LOGGER.info(Messages.FILE_SERVICE_DELETING_FILES);
+                int deletedFiles = fileService.deleteFilesEntriesWithoutContent();
+                LOGGER.info(MessageFormat.format(Messages.FILE_SERVICE_DELETED_FILES, deletedFiles));
+            } catch (FileStorageException e) {
```
IvanBorislavovDimitrov, line 85: Maybe catch Exception, because if something happens we will have no idea of the root cause

IvanBorislavovDimitrov (approved):

PR #1227, opened 2022-09-16
Add jdbc configuration for sslkey

vkalapov:
LMCROSSITXSADEPLOY-2639

### DIFF
--- a/multiapps-controller-persistence/pom.xml
+++ b/multiapps-controller-persistence/pom.xml
@@ -198,5 +198,9 @@
             <groupId>org.eclipse.persistence</groupId>
             <artifactId>org.eclipse.persistence.extension</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk18on</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/Constants.java
+++ b/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/Constants.java
@@ -20,6 +20,8 @@ public class Constants {
     public static final String FILE_ENTRY_DIGEST = "DIGEST";
     public static final String FILE_ENTRY_ID = "FILE_ID";
 
+    public static final String SSL_CLIENT_KEY_FILE_NAME = "ClientKey.der";
+
     protected Constants() {
     }
 
--- a/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
+++ b/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
@@ -1,0 +1,29 @@
+package org.cloudfoundry.multiapps.controller.persistence.util;
+
+import java.io.IOException;
+import java.io.StringReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.bouncycastle.openssl.PEMParser;
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
+import org.cloudfoundry.multiapps.common.SLException;
+import org.cloudfoundry.multiapps.controller.persistence.Messages;
+
+public class ClientKeyConfigurationHandler {
+
+    public Path createEncodedKeyFile(String pemKey, String fileName) {
+        try (PEMParser pemParser = new PEMParser(new StringReader(pemKey))) {
+            PEMToEncodedKeyConverter pemKeyConverter = new PEMToEncodedKeyConverter();
+            byte[] encodedKey = pemKeyConverter.getPrivateEncodedKey(pemParser);
+            return writeEncodedKeyToFile(encodedKey, fileName);
+        } catch (IOException e) {
+            throw new SLException(e, Messages.GENERATING_KEY_FILE_FAILED, fileName);
+        }
+    }
+
+    private Path writeEncodedKeyToFile(byte[] encodedKey, String fileName) throws IOException {
+        return Files.write(Paths.get(fileName), encodedKey);
+    }
+}
--- a/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/DataSourceFactory.java
+++ b/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/DataSourceFactory.java
@@ -1,7 +1,11 @@
 package org.cloudfoundry.multiapps.controller.persistence.util;
 
+import java.nio.file.Path;
+
 import javax.inject.Named;
 import javax.sql.DataSource;
+
+import org.cloudfoundry.multiapps.controller.persistence.Constants;
 
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
@@ -28,11 +32,29 @@ public class DataSourceFactory {
         hikariConfig.setIdleTimeout(60000);
         hikariConfig.setMinimumIdle(10);
         hikariConfig.addDataSourceProperty("tcpKeepAlive", true);
+
+        configureSSLClientKeyIfExists(service, hikariConfig);
+
         if (maximumPoolSize != null) {
             hikariConfig.setMaximumPoolSize(maximumPoolSize);
         }
         hikariConfig.setRegisterMbeans(true);
         return hikariConfig;
+    }
+
+    private void configureSSLClientKeyIfExists(CfJdbcService service, HikariConfig hikariConfig) {
+        String clientKey = (String) service.getCredentials()
+                                           .getMap()
+                                           .get("sslkey");
+        if (clientKey != null) {
+            configureClientCertificate(clientKey, hikariConfig);
+        }
+    }
+
+    private void configureClientCertificate(String clientKey, HikariConfig hikariConfig) {
+        ClientKeyConfigurationHandler sslKeyHandler = new ClientKeyConfigurationHandler();
+        Path encodedKeyPath = sslKeyHandler.createEncodedKeyFile(clientKey, Constants.SSL_CLIENT_KEY_FILE_NAME);
+        hikariConfig.addDataSourceProperty("sslkey", encodedKeyPath.toAbsolutePath());
     }
 
 }
--- a/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/PEMToEncodedKeyConverter.java
+++ b/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/PEMToEncodedKeyConverter.java
@@ -1,0 +1,26 @@
+package org.cloudfoundry.multiapps.controller.persistence.util;
+
+import java.io.IOException;
+import java.security.KeyPair;
+import java.text.MessageFormat;
+
+import org.bouncycastle.openssl.PEMKeyPair;
+import org.bouncycastle.openssl.PEMParser;
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
+import org.cloudfoundry.multiapps.controller.persistence.Messages;
+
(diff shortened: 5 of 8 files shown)


multiapps-controller-persistence/pom.xml
```diff
@@ -198,5 +198,10 @@
             <groupId>org.eclipse.persistence</groupId>
             <artifactId>org.eclipse.persistence.extension</artifactId>
         </dependency>
+        <dependency>
+			<groupId>org.bouncycastle</groupId>
+			<artifactId>bcpkix-jdk15on</artifactId>
+			<version>1.70</version>
+		</dependency>
```
IvanBorislavovDimitrov, line 205: Extract the dependency in the parent pom, make the version as a property and just refer it in this maven module

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/DataSourceFactory.java
```diff
@@ -28,11 +28,29 @@ private HikariConfig createHikariConfig(CfJdbcService service, Integer maximumPo
...
             hikariConfig.setMaximumPoolSize(maximumPoolSize);
         }
         hikariConfig.setRegisterMbeans(true);
         return hikariConfig;
     }
 
+    private void configureSSLClientKeyIfExists(CfJdbcService service, HikariConfig hikariConfig) {
+        Object clientKey = service.getCredentials()
+                                  .getMap()
+                                  .get("sslkey");
```
IvanBorislavovDimitrov, line 44: Isn't this property a String, can you cast it?

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
```diff
@@ -0,0 +1,32 @@
...
+import org.bouncycastle.openssl.PEMParser;
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
+import org.cloudfoundry.multiapps.controller.persistence.Messages;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ClientKeyConfigurationHandler {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClientKeyConfigurationHandler.class);
+
+    public void createEncodedKeyFile(String pemKey, String fileName) {
```
IvanBorislavovDimitrov, line 18: This method should return the File

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/DataSourceFactory.java
```diff
@@ -28,11 +28,29 @@ private HikariConfig createHikariConfig(CfJdbcService service, Integer maximumPo
...
+        if (clientKey != null) {
+            configureClientCertificate(clientKey.toString(), hikariConfig);
+        }
+    }
+
+    private void configureClientCertificate(String clientKey, HikariConfig hikariConfig) {
+        ClientKeyConfigurationHandler sslKeyHandler = new ClientKeyConfigurationHandler();
+        sslKeyHandler.createEncodedKeyFile(clientKey, "ClientKey.der");
```
IvanBorislavovDimitrov, line 52: Extract the filename as constant.
If this method returns a File, you can use it, this way you will not hardcode paths in the filesystem. Not really recomended

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/PEMToEncodedKeyConverter.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class PEMToEncodedKeyConverter {
+
+    public byte[] getPrivateEncodedKey(PEMParser pemParser, JcaPEMKeyConverter converter) throws IOException {
+        Object pemContent = pemParser.readObject();
+        if (pemContent instanceof PEMKeyPair) {
+            KeyPair keyPair = converter.getKeyPair((PEMKeyPair) pemContent);
+            return keyPair.getPrivate()
+                          .getEncoded();
+        } else {
```
IvanBorislavovDimitrov, line 20: This else is not needed

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
```diff
@@ -0,0 +1,32 @@
...
+        try (PEMParser pemParser = new PEMParser(new StringReader(pemKey))) {
+            PEMToEncodedKeyConverter derKeyHandler = new PEMToEncodedKeyConverter();
+            JcaPEMKeyConverter converter = new JcaPEMKeyConverter();
+            byte[] encodedKey = derKeyHandler.getPrivateEncodedKey(pemParser, converter);
+            writeEncodedKeyToFile(encodedKey, fileName);
+        } catch (IOException e) {
+            LOGGER.error(MessageFormat.format(Messages.GENERATING_KEY_FILE_FAILED, e.getMessage()));
```
IvanBorislavovDimitrov, line 25: Shouldn't we rethrow the error?

multiapps-controller-persistence/pom.xml
theghost5800: Add it in parent pom.xml under `<dependencyManagement>` section and here just specify the dependency group id and artifact id without version.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
theghost5800: If this method won't return anything, then change the name of method. `fileName` parameter could be Path object instead String.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
```diff
@@ -0,0 +1,32 @@
...
+import org.cloudfoundry.multiapps.controller.persistence.Messages;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ClientKeyConfigurationHandler {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClientKeyConfigurationHandler.class);
+
+    public void createEncodedKeyFile(String pemKey, String fileName) {
+        try (PEMParser pemParser = new PEMParser(new StringReader(pemKey))) {
+            PEMToEncodedKeyConverter derKeyHandler = new PEMToEncodedKeyConverter();
```
theghost5800, line 20: I am little bit confused. How `derKeyHandler` comes from? The object is named `PEMToEcnodedKeyConverter` so it is again converter from PEM to encoded format. Variable name should be different in my opinion.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/DataSourceFactory.java
vkalapov: Yes, I assume there wouldn't be any problem with casting the property to String. I'll change it to that.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/PEMToEncodedKeyConverter.java
```diff
@@ -0,0 +1,26 @@
...
+import java.text.MessageFormat;
+
+import org.bouncycastle.openssl.PEMKeyPair;
+import org.bouncycastle.openssl.PEMParser;
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
+import org.cloudfoundry.multiapps.controller.persistence.Messages;
+
+public class PEMToEncodedKeyConverter {
+
+    public byte[] getPrivateEncodedKey(PEMParser pemParser, JcaPEMKeyConverter converter) throws IOException {
```
theghost5800, line 14: I think this method could be inside ClientKeyConfigurationHandler class or move `JcaPEMKeyConverter` initialization inside this method.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/DataSourceFactory.java
theghost5800: +1

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/DataSourceFactory.java
```diff
@@ -28,11 +28,29 @@ private HikariConfig createHikariConfig(CfJdbcService service, Integer maximumPo
...
+        hikariConfig.addDataSourceProperty("sslkey", "/home/vcap/app/ClientKey.der");
```
theghost5800, line 53: Use Files api of java, do not hardcode filepath as String.

theghost5800 (changes requested):

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
vkalapov: To clarify If I understand correctly - by returning the File, do you mean returning the Path object returned from the method Files.write()?

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
vkalapov: I agree that the variable should be renamed. The reason why it's named `derKeyHandler` is that the .der format is basically the binary version of a PEM extension, so an encoded pem file can be seen as a der file. But even then the name is confusing. Something along the lines of `pemKeyConverter` I think might be more appropriate

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/util/ClientKeyConfigurationHandler.java
vkalapov: I agree the exception should be changed to be rethrown.

IvanBorislavovDimitrov (approved):

pom.xml
```diff
@@ -792,6 +793,12 @@
                 <artifactId>resilience4j-ratelimiter</artifactId>
                 <version>${resilience4j.version}</version>
             </dependency>
+            <!-- https://mvnrepository.com/artifact/org.bouncycastle/bcpkix-jdk15on -->
+            <dependency>
+                <groupId>org.bouncycastle</groupId>
+                <artifactId>bcpkix-jdk15on</artifactId>
```
theghost5800, line 799: Actually this versions is intended for Java 5 and version 1.70 is from 2021. From their site https://www.bouncycastle.org/latest_releases.html we can use artifactId: jdk18on and version 1.71.1 https://mvnrepository.com/artifact/org.bouncycastle/bcpkix-jdk18on

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/Constants.java
```diff
@@ -20,6 +20,8 @@ public class Constants {
     public static final String FILE_ENTRY_DIGEST = "DIGEST";
     public static final String FILE_ENTRY_ID = "FILE_ID";
 
+    public static final String SSL_CLIENT_KEY_FILE = "ClientKey.der";
```
theghost5800, line 23: Rename it to SSL_CLIENT_KEY_FILE_NAME because this constant provide the name of the file, not the file itself.

theghost5800 (approved):

PR #1258, opened 2022-12-06
Add Cloud Controller header configuration

vkalapov:
LMCROSSITXSADEPLOY-2678


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/clients/CustomControllerClient.java
```diff
@@ -41,6 +57,14 @@ private PaginationV3 addPageOfResources(String uri, ResourcesResponseMapper<?> r
         return PaginationV3.fromResponse(responseMap);
     }
 
+    private MultiValueMap<String, String> generateRequestHeaders(String correlationId) {
+        Map<String, List<String>> map = headerConfiguration.generateHeaders(correlationId)
+                                                           .entrySet()
+                                                           .stream()
+                                                           .collect(Collectors.toMap(Map.Entry::getKey, e -> Arrays.asList(e.getValue())));
+        return CollectionUtils.toMultiValueMap(map);
```
radito3, line 65: I would recommend doing this instead of using streams:
```java
var result = LinkedMultiValueMap<String, String>();
headerConfi.generateHeaders(correlationId).forEach(result::add);
return result;
```
The `add` method accepts a key and a single value and inserts it into a list value.
That way we can avoid some of the technicalities when converting from a normal to a multi-value map.

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/CreateOrUpdateStepWithExistingAppTest.java
```diff
@@ -202,8 +201,11 @@ private void prepareClient(CloudApplication application, Set<CloudRoute> routes,
...
                                                                                                   .diskInMb(disk == null ? 1024 : disk)
-                                                                                                  .memoryInMb(memory == null ? 1024 : memory)
-                                                                                                  .healthCheckType(hcType == null ? HealthCheckType.PORT : HealthCheckType.valueOf(hcType.toUpperCase()))
+                                                                                                  .memoryInMb(memory == null ? 1024
+                                                                                                      : memory)
+                                                                                                  .healthCheckType(hcType == null
+                                                                                                      ? HealthCheckType.PORT
+                                                                                                      : HealthCheckType.valueOf(hcType.toUpperCase()))
```
radito3, line 208: Is this a formatter change?

vkalapov: Yes, the lines were automatically changed by the formatter.

radito3 (approved):

PR #1265, opened 2023-01-09
Fix deploy from URL while using multipart upload

abdermendz:
Issue: LMCROSSITXSADEPLOY-2669

### DIFF
--- a/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/services/ObjectStoreFileStorage.java
+++ b/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/services/ObjectStoreFileStorage.java
@@ -1,5 +1,6 @@
 package org.cloudfoundry.multiapps.controller.persistence.services;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.text.MessageFormat;
 import java.util.Date;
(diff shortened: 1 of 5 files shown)


multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/services/ObjectStoreFileStorage.java
```diff
@@ -44,17 +47,19 @@ public ObjectStoreFileStorage(BlobStore blobStore, String container) {
...
     @Override
-    public void addFile(FileEntry fileEntry, InputStream content) throws FileStorageException {
+    public void addFile(FileEntry fileEntry, InputStream content) throws FileStorageException, IOException {
         String entryName = fileEntry.getId();
         long fileSize = fileEntry.getSize()
                                  .longValue();
+        LOGGER.trace("Building blob ...");
         Blob blob = blobStore.blobBuilder(entryName)
-                             .payload(content)
+                             .payload(ByteSource.wrap(ByteStreams.toByteArray(content)))
```
radito3, line 56: This will read the entire input stream in memory - [source code](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/ByteStreams.java#L173)
This wouldn't be a problem for small MTARs but it does not scale.
The issue is somewhat mitigated by the small usage of the _deploy from URL_ feature but is still worth considering.

If we are going to read the whole file in memory, prefer to use the standard library or apache commons (IOUtils.toByteArray) so we wouldn't need to include the Guava library as a dependency (it is quite sizable and will only bloat our executable)

PR #1281, opened 2023-02-22
Adopt new Azure objectstore endpoint

boyan-velinov:
Jira: LMCROSSITXSADEPLOY-2727

### DIFF
--- a/multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/service/ObjectStoreServiceInfoCreator.java
+++ b/multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/service/ObjectStoreServiceInfoCreator.java
@@ -1,5 +1,7 @@
 package org.cloudfoundry.multiapps.controller.web.configuration.service;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.Map;
 
 import io.pivotal.cfenv.core.CfService;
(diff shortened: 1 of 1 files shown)


multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/service/ObjectStoreServiceInfoCreator.java
```diff
@@ -42,10 +44,17 @@ private ObjectStoreServiceInfo createServiceInfoForAzure(Map<String, Object> cre
...
+        try {
+            containerUri = new URL((String) credentials.get("container_uri"));
+        } catch (MalformedURLException e) {
+            throw new IllegalStateException("Cannot parse container_uri of objectstore", e);
+        }
         return ImmutableObjectStoreServiceInfo.builder()
                                               .provider("azureblob")
                                               .identity(accountName)
                                               .credential(sasToken)
+                                              .endpoint(containerUri.getProtocol() + "://" + containerUri.getHost())
```
radito3, line 57: Can we use `containerUri.toString()`?

boyan-velinov: Seems not, because in the `container_uri` that we have in test environment, URI contains also the path in the end, which would be appended in  `toString()`.
I did not find any validation in Jclouds that strips down the whole URI into only schema+host

PR #1287, opened 2023-03-01
Adopt azure container uri

IvanBorislavovDimitrov:
LMCROSSITXSADEPLOY-2727

### DIFF
--- a/multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/service/ObjectStoreServiceInfoCreator.java
+++ b/multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/service/ObjectStoreServiceInfoCreator.java
@@ -1,6 +1,10 @@
 package org.cloudfoundry.multiapps.controller.web.configuration.service;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.Map;
+
+import org.cloudfoundry.multiapps.controller.web.Messages;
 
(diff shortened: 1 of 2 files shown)


multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/service/ObjectStoreServiceInfoCreator.java
```diff
@@ -46,10 +50,20 @@ private ObjectStoreServiceInfo createServiceInfoForAzure(Map<String, Object> cre
...
                                               .credential(sasToken)
+                                              .endpoint(getContainerUriEndpoint(credentials).toString())
                                               .container(containerName)
                                               .build();
     }
 
+    private URL getContainerUriEndpoint(Map<String, Object> credentials) {
+        try {
+            URL containerUri = new URL((String) credentials.get("container_uri"));
+            return new URL(containerUri.getProtocol(), containerUri.getHost(), containerUri.getPort(), "/");
```
radito3, line 61: Does the `getHost()` method return the domain as well?
And why do we need a forwards slash at the end if the path isn't provided?

IvanBorislavovDimitrov: yes, it returns everything up to the first slash.
The path will be provided by the jclouds library, it is the container_name

radito3 (approved):

PR #1288, opened 2023-03-02
Fix NPE when determining which binding to delete during service recre…

IvanBorislavovDimitrov:
LMCROSSITXSADEPLOY-2734


IvanBorislavovDimitrov:
Consider changing ManageAppServiceBindingEndListener as APP_TO_PROCESS is used there as well

radito3 (approved):

PR #1294, opened 2023-03-15
Improve service key recreation logic

IvanBorislavovDimitrov:
[LMCROSSITXSADEPLOY-2716](https://jira.tools.sap/browse/LMCROSSITXSADEPLOY-2716)


vkalapov (commented):
Is the potential NPE possible due to the changes for service key recreation during bg-deploy or it was always possible? 

Edit: It seems to be the latter

vkalapov (approved):

PR #1300, opened 2023-04-03
Explicitly hint the data source config about the postgresql driver

danail-branekov:
#### Description:
It seems that in some environments the postgresql driver is not available on the mta controller classpath, thus causing connection to the postgre datatabase fail with `java.lang.RuntimeException: Failed to get driver instance for jdbcUrl=<some-valid-url>`. Here is a separate slack conversation on the issue: https://cloudfoundry.slack.com/archives/C9V0Z8ZPF/p1680505547626489

This PR introduces an explicit dependency to the postgresql driver, its latest is currently `42.6.0`. This seems to solve the issue.

However, one should keep in mind that doing that prevents a buildpack from managing the driver version, e.g. it would not be able to bring potential security fixes without changing this dependency.

### DIFF
--- a/multiapps-controller-persistence/pom.xml
+++ b/multiapps-controller-persistence/pom.xml
@@ -202,5 +202,10 @@
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-jdk18on</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.postgresql</groupId>
+            <artifactId>postgresql</artifactId>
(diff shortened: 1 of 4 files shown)


boyan-velinov:
Thanks for the PR!
It is worth to having it in place but for the time being we will refrain from having it in master. It is good for testing purposes only

bonzofenix:
@boyan-velinov this would be useful for us in the Autoscaler team. Is there a reason for not having it in the master version ?

PR #1305, opened 2023-05-05
Unlock running jobs during app instance restart

IvanBorislavovDimitrov:
When app instance crashes / flowable engine gets shutdown a lock_owner and lock_exp_time should also be removed from act_ru_jobs and the process lock_owner and lock_time should be removed.
The commit removes the lock_onwer and time in act_ru_job
Flowable forum thread: https://forum.flowable.org/t/clear-lock-owner-and-lock-exp-time-from-act-ru-job/10465

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/flowable/commands/ClearJobLockOwnersCmd.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/flowable/commands/ClearJobLockOwnersCmd.java
@@ -1,0 +1,44 @@
+package org.cloudfoundry.multiapps.controller.process.flowable.commands;
+
+import java.text.MessageFormat;
+import java.util.List;
+
+import org.cloudfoundry.multiapps.controller.process.Messages;
+import org.flowable.common.engine.impl.interceptor.Command;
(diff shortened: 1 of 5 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/flowable/commands/ClearJobLockOwnersCmd.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public ClearJobLockOwnersCmd(List<Job> lockedJobs, JobServiceConfiguration jobServiceConfiguration) {
+        this.lockedJobs = lockedJobs;
+        this.jobServiceConfiguration = jobServiceConfiguration;
+    }
+
+    @Override
+    public Void execute(CommandContext commandContext) {
+        JobEntityManager jobEntityManager = jobServiceConfiguration.getJobEntityManager();
+        lockedJobs.forEach(job -> {
```
radito3, line 31: I think it would be cleaner to use a regular for loop.
The main use case for the `Collection.forEach` method is, in my opinion, for one-liners:
```java
list.forEach(this::doSomething);
```
Otherwise, a regular for loop is cleaner and easier to read

radito3 (approved):

radito3:
LGMT

PR #1315, opened 2023-05-17
Remove objectstore implementation

tilieva:
Objectstore is not used in on-premise delivery

Jira: LMCROSSITXSADEPLOY-2736


boyan-velinov (commented):
In addition, the objectstore removal needs:
* Removal of class com.sap.cloud.lm.sl.cf.web.configuration.service.ObjectStoreServiceInfoCreator
* Change in org.springframework.cloud.cloudfoundry.CloudFoundryServiceInfoCreator to remove reference of the above file

boyan-velinov (approved):

PR #1329, opened 2023-06-16
Refactor usage of in-memory caches

radito3:


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/CachedMap.java
```diff
@@ -26,12 +27,25 @@ public V get(K key) {
...
+        }
+        value = creator.get();
+        put(key, value);
+        return value;
+    }
+
     public void put(K key, V value) {
+        if (value == null) {
+            return;
+        }
```
IvanBorislavovDimitrov, line 43: Why is this check? Do we really put nulls?

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/security/data/termination/DataTerminationService.java
```diff
@@ -83,28 +79,20 @@ private List<String> getSpaceDeleteEvents() {
...
+        return new CFOptimizedEventGetter(cloudCredentials);
     }
 
     private String getDateBeforeDays(int numberOfDays) {
         ZonedDateTime dateBeforeTwoDays = ZonedDateTime.now()
                                                        .minus(Duration.ofDays(numberOfDays));
-        String result = DATE_TIME_FORMATTER.format(dateBeforeTwoDays);
+        //TODO test this format...
+        String result = DateTimeFormatter.ISO_DATE_TIME
+                                         .format(dateBeforeTwoDays);
```
IvanBorislavovDimitrov, line 95: Complete the todo

multiapps-controller-core/src/test/java/org/cloudfoundry/multiapps/controller/core/helpers/ClientHelperTest.java
```diff
@@ -67,37 +52,37 @@ void testComputeSpaceIdIfSpaceIsNull() {
...
+//    }
+//
+//    @Test
+//    void testComputeTargetCloudOperationExceptionBadRequestThrown() {
+//        Mockito.when(client.getSpace(ArgumentMatchers.any(UUID.class)))
+//               .thenThrow(new CloudOperationException(HttpStatus.BAD_REQUEST));
+//        CloudOperationException cloudOperationException = Assertions.assertThrows(CloudOperationException.class,
+//                                                                                  () -> clientHelper.computeTarget(SPACE_ID));
+//        Assertions.assertEquals(HttpStatus.BAD_REQUEST, cloudOperationException.getStatusCode());
+//    }
```
IvanBorislavovDimitrov, line 85: I guess you will fix them

IvanBorislavovDimitrov (changes requested):

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/CloudControllerClientProvider.java
```diff
@@ -79,116 +54,29 @@ public CloudControllerClient getControllerClientWithNoCorrelation(String userNam
...
+    public void releaseClient(String userName, String spaceGuid) {
+        clients.remove(getKey(userName, spaceGuid));
     }
 
     private CloudControllerClient getClientFromCacheWithNoCorrelation(String userName, String spaceId) {
-        String key = getKey(userName, spaceId, NO_CORRELATION_ID);
+        String key = getKey(spaceId, userName);
         return clients.computeIfAbsent(key,
-                                       k -> clientFactory.createClient(getValidToken(userName), spaceId, null));
+                                       () -> clientFactory.createClient(tokenService.getToken(userName), spaceId, null));
```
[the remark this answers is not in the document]
radito3: Yes, it's used.
The flow for the tokens goes like this:
```
- get token from cache
- create oauth client object (contains a web client, a token factory, last used token and others..)
- create ClientCredentials object with the retrieved token
- initialize the oauth client with those credentials (sets the token field in OAuthClient)
- create the cloud controller client
```
Then, on calls to CF:
```
- OAuthClient.getToken() 
- check if the token field is set
   - if set and the token expires after more than 50 seconds, return it
   - otherwise, make a call to the UAA endpoint to fetch a new one and update the field
- add token to authorization header
```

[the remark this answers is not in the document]
IvanBorislavovDimitrov: Ok, I agree it is used. Do you think it makes sense (with another backlog) to introduce a unified way of getting tokens? Currently, we use:
oAuthClient.getToken() - For Oauth and Basic depending if an extended client is created or not.
oAuthClient.getAuthorizationHeaderValue() - to get the token value

What seems the best in my opinion is some sort of a filter as in "CloudControllerRestClientRequestFilterFunction", right before every call to CC to renew the token. The custom clients have to be changed as well as the logging client.
There are a lot of places where we invoke oAuthClient.init which might be missed.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/clients/CfRolesGetter.java
```diff
@@ -0,0 +1,51 @@
...
+        return new CustomControllerClientErrorHandler().handleErrorsOrReturnResult(() -> doGetRoles(spaceGuid, userGuid));
+    }
+
+    private Set<UserRole> doGetRoles(UUID spaceGuid, UUID userGuid) {
+        String url = String.format(GET_ROLES_URL, spaceGuid, userGuid);
+        var list = getListOfResources(new UserRoleMapper(), url);
+        var result = EnumSet.noneOf(UserRole.class);
+        result.addAll(list);
+        return result;
+    }
```
IvanBorislavovDimitrov, line 35: Use EnumSet.copyOf

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/security/AuthorizationChecker.java
```diff
@@ -91,23 +95,21 @@ boolean checkPermissions(UserInfo userInfo, String orgName, String spaceName, bo
         var userToken = tokenService.getToken(userInfo.getName());
         var spaceClient = clientFactory.createSpaceClient(userToken);
         var space = spaceClient.getSpace(orgName, spaceName);
-        var spaceGuid = space.getGuid()
-                             .toString();
 
-        CloudControllerClient client = clientFactory.createClient(userToken, spaceGuid, null);
+        CfRolesGetter rolesGetter = new CfRolesGetter(webClientFactory, new CloudCredentials(userToken));
```
IvanBorislavovDimitrov, line 99: You can extract the creation in a method, so it could be mocked in tests

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/cf/clients/CfRolesGetter.java
radito3: This will throw an exception if the collection is empty

IvanBorislavovDimitrov (approved):

PR #1411, opened 2024-01-06
Update optional services to don't fail sometimes

Yavor16:
Issue: [LMCROSSITXSADEPLOY-2784](https://jira.tools.sap/browse/LMCROSSITXSADEPLOY-2784)

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
@@ -202,6 +202,11 @@ public class Messages {
     public static final String COULD_NOT_BIND_OPTIONAL_SERVICE_TO_APP = "Could not bind optional service \"{0}\" to application \"{1}\"";
     public static final String COULD_NOT_UNBIND_OPTIONAL_SERVICE_TO_APP = "Could not unbind optional service \"{0}\" to application \"{1}\"";
     public static final String COULD_NOT_CREATE_OPTIONAL_SERVICE = "Could not create optional service \"{0}\"";
+    public static final String COULD_NOT_UPDATE_TAGS_OF_OPTIONAL_SERVICE = "Could not update tags of optional service \"{0}\" : {1}";
+    public static final String COULD_NOT_UPDATE_METADATA_OF_OPTIONAL_SERVICE = "Could not update metadata of optional service \"{0}\" : {1}";
+    public static final String COULD_NOT_UPDATE_PARAMETERS_OPTIONAL_SERVICE = "Could not update parameters of optional service \"{0}\" : {1}";
+    public static final String COULD_NOT_UPDATE_PLAN_OPTIONAL_SERVICE = "Could not update plan of optional service \"{0}\" : {1}";
+    public static final String COULD_NOT_UPDATE_SYSLOG_DRAIN_URL_OPTIONAL_SERVICE = "Could not update syslog drain url of optional service \"{0}\" : {1}";
     public static final String COULD_NOT_GET_SERVICE_KEYS_FOR_OPTIONAL_SERVICE = "Could not get service keys for optional service \"{0}\"";
     public static final String DEFAULT_FAILED_OPERATION_DESCRIPTION = "The service broker returned an error with no description!";
     public static final String ERROR_DURING_CLEAN_UP_0 = "Error during clean-up: {0}";
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/CreateServiceStep.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/CreateServiceStep.java
@@ -19,9 +19,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.util.Assert;
 
 import com.sap.cloudfoundry.client.facade.CloudControllerClient;
-import com.sap.cloudfoundry.client.facade.CloudControllerException;
 import com.sap.cloudfoundry.client.facade.CloudOperationException;
-import com.sap.cloudfoundry.client.facade.CloudServiceBrokerException;
 import com.sap.cloudfoundry.client.facade.domain.CloudServiceInstance;
 import com.sap.cloudfoundry.client.facade.domain.ServiceOperation;
 
@@ -44,7 +42,12 @@ public class CreateServiceStep extends ServiceStep {
             if (operationExecutionState.isPresent()) {
                 return operationExecutionState.get();
             }
-            processServiceCreationFailure(context, serviceInstance, e);
+            String exceptionDescription = MessageFormat.format(Messages.COULD_NOT_CREATE_OPTIONAL_SERVICE, serviceInstance.getName());
+            CloudOperationException cloudOperationException = new CloudOperationException(e.getStatusCode(),
+                                                                                          e.getStatusText(),
+                                                                                          exceptionDescription);
+
+            processServiceActionFailure(context, serviceInstance, cloudOperationException);
         }
         return OperationExecutionState.FINISHED;
     }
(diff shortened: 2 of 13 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/SyncFlowableStep.java
```diff
@@ -208,6 +211,31 @@ protected ProcessStepHelper getStepHelper() {
...
     }
 
+    protected void processServiceCreationFailure(ProcessContext context, CloudServiceInstanceExtended serviceInstance,
+                                               CloudOperationException e) {
+        processServiceActionFailure(context, serviceInstance, e, MessageFormat.format(Messages.COULD_NOT_CREATE_OPTIONAL_SERVICE, serviceInstance.getName()));
+    }
+
+    protected void processServiceUpdateFailure(ProcessContext context, CloudServiceInstanceExtended serviceInstance,
+                                               CloudOperationException e, String errorStep) {
+        processServiceActionFailure(context, serviceInstance, e, MessageFormat.format(Messages.COULD_NOT_UPDATE_OPTIONAL_SERVICE, serviceInstance.getName(), errorStep));
```
IvanBorislavovDimitrov, line 221: this method has a lot of parameters, let's try no to use more than 3. 
instead of adding a customErrorMessage can you add this message in the exception.
in the steps you can create a new CloudOperationException with more descriptive message and pass it to this method

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UpdateServiceParametersStep.java
```diff
@@ -18,6 +20,8 @@
 @Scope(BeanDefinition.SCOPE_PROTOTYPE)
 public class UpdateServiceParametersStep extends ServiceStep {
 
+    private static final String PARAMETERS = "parameters";
```
IvanBorislavovDimitrov, line 23: if you still need these parameters after fixing the above comment, move them to another class with constants/messages. 
Same applies for other tags, syslog, etc

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UpdateServiceParametersStep.java
```diff
@@ -27,9 +31,12 @@ protected OperationExecutionState executeOperation(ProcessContext context, Cloud
...
-
-        getStepLogger().debug(Messages.SERVICE_UPDATED, service.getName());
+        try {
+            client.updateServiceParameters(service.getName(), service.getCredentials());
+            getStepLogger().debug(Messages.SERVICE_UPDATED, service.getName());
+        } catch (CloudOperationException e) {
+            processServiceUpdateFailure(context, service, e, PARAMETERS);
+        }
 
         return OperationExecutionState.EXECUTING;
```
IvanBorislavovDimitrov, line 41: In case of error, you do not need to monitor the service again
return OperationExecutionState.FINISHED instead of EXECUTING

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UpdateServiceSyslogDrainUrlStep.java
```diff
@@ -28,9 +31,13 @@ protected OperationExecutionState executeOperation(ProcessContext context, Cloud
         }
         getStepLogger().info(Messages.UPDATING_SERVICE_SYSLOG_URL, service.getName());
 
-        client.updateServiceSyslogDrainUrl(service.getName(), service.getSyslogDrainUrl());
+        try {
+            client.updateServiceSyslogDrainUrl(service.getName(), service.getSyslogDrainUrl());
+            getStepLogger().debug(Messages.SERVICE_SYSLOG_URL_UPDATED, service.getName());
+        } catch (CloudOperationException e) {
+            processServiceUpdateFailure(context, service, e, SYSLOG_DRAIN_URL);
```
IvanBorislavovDimitrov, line 38: same, return FINISHED here,
check for the other usages as well

IvanBorislavovDimitrov (approved):

IvanBorislavovDimitrov:
Do not forget to squash some of the commits

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/SyncFlowableStep.java
```diff
@@ -208,6 +211,31 @@ protected ProcessStepHelper getStepHelper() {
         return stepHelper;
     }
 
+    protected void processServiceCreationFailure(ProcessContext context, CloudServiceInstanceExtended serviceInstance,
+                                               CloudOperationException e) {
+        processServiceActionFailure(context, serviceInstance, e, MessageFormat.format(Messages.COULD_NOT_CREATE_OPTIONAL_SERVICE, serviceInstance.getName()));
+    }
+
+    protected void processServiceUpdateFailure(ProcessContext context, CloudServiceInstanceExtended serviceInstance,
```
ikasarov: processService... methods don't really belong in this superclass. A more appropriate place would be ServiceStep or if that doesn't work the most concrete shared superclass between the steps. that require the methods

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
```diff
@@ -202,6 +202,7 @@ public class Messages {
     public static final String COULD_NOT_BIND_OPTIONAL_SERVICE_TO_APP = "Could not bind optional service \"{0}\" to application \"{1}\"";
     public static final String COULD_NOT_UNBIND_OPTIONAL_SERVICE_TO_APP = "Could not unbind optional service \"{0}\" to application \"{1}\"";
     public static final String COULD_NOT_CREATE_OPTIONAL_SERVICE = "Could not create optional service \"{0}\"";
+    public static final String COULD_NOT_UPDATE_OPTIONAL_SERVICE = "Could not update optional service \"{0}\" {1}";
```
ikasarov: Maybe reformat the message to be worded more clearly, like: 'Could not update **metadata** of option service "service-name"'

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/CreateServiceStep.java
```diff
@@ -44,7 +42,10 @@ protected OperationExecutionState executeOperation(ProcessContext context, Cloud
             if (operationExecutionState.isPresent()) {
                 return operationExecutionState.get();
             }
-            processServiceCreationFailure(context, serviceInstance, e);
+
+            processServiceActionFailure(context, serviceInstance,
```
ikasarov, line 46: If you're making the exception at the same time as calling that method - take the exception out of the method invocation on the line just above to make it more readable. e.g.:
  CloudOperationException createFailedException = createCloudOperationException(...);
  processServiceActionFailure(context, serviceInstance, createFailedException);

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ServiceStep.java
```diff
@@ -61,6 +61,10 @@ private OperationExecutionState executeOperationAndHandleExceptions(ProcessConte
     protected abstract OperationExecutionState executeOperation(ProcessContext context, CloudControllerClient controllerClient,
                                                                 CloudServiceInstanceExtended service);
 
+    protected CloudOperationException createCloudOperationException(CloudOperationException e, String message, String... params) {
```
ikasarov, line 64: This seems like an unnecessary method - it only calls one other method inside, with almost the same signature.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/SyncFlowableStep.java
```diff
@@ -211,29 +211,19 @@ protected ProcessStepHelper getStepHelper() {
         return stepHelper;
     }
 
-    protected void processServiceCreationFailure(ProcessContext context, CloudServiceInstanceExtended serviceInstance,
+    protected void processServiceActionFailure(ProcessContext context, CloudServiceInstanceExtended serviceInstance,
```
ikasarov, line 214: added a comment on this already

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/UpdateServiceMetadataStep.java
```diff
@@ -32,7 +31,10 @@ protected OperationExecutionState executeOperation(ProcessContext context, Cloud
             client.updateServiceInstanceMetadata(serviceGuid, service.getV3Metadata());
             getStepLogger().debug(Messages.UPDATING_METADATA_OF_SERVICE_INSTANCE_0_DONE, service.getName());
         } catch (CloudOperationException e) {
-            processServiceUpdateFailure(context, service, e, METADATA);
+            processServiceActionFailure(context, service,
+                                        createCloudOperationException(e, MessageFormat.format(Messages.COULD_NOT_UPDATE_OPTIONAL_SERVICE,
```
ikasarov, line 35: again, maybe take out the exception creation out of the method invocation, same for the ones below

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
```diff
@@ -673,6 +673,11 @@ public class Messages {
     public static final String GETTING_HISTORIC_VARIABLES_FOR_OPERATIONS_STARTED_AFTER_0 = "Getting historic variables for operations started after {0}";
     public static final String NO_ORPHANED_FILES_TO_DELETE = "No orphaned files to delete.";
     public static final String DELETING_ORPHANED_FILES_0 = "Deleting {0} orphaned files: {1}";
+    public static final String TAGS = "tags";
```
ikasarov, line 676: The names of these constants are not really descriptive, but rather only repeat the same words that they contain. Maybe rename to something more reasonable, like 'SERVICE_ATTRIBUTE_TAGS" etc. Or better yet just don't make these service parameters into different constants and have separate custom error messages (if it makes sense). For example, 'service tags' are already used in log messages without them being separated into constants (better to follow established use of messages):
    public static final String UPDATING_SERVICE_TAGS = "Updating service tags for service \"{0}\"...";
    public static final String SERVICE_TAGS_UPDATED = "Service tags for service \"{0}\" updated";

ikasarov (changes requested):

ikasarov (approved):

PR #1428, opened 2024-01-31
Do not always print service is optional during failure

Yavor16:
LMCROSSITXSADEPLOY-2878

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ServiceStep.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ServiceStep.java
@@ -6,8 +6,6 @@ import java.util.Map;
 
 import javax.inject.Inject;
 
-import com.sap.cloudfoundry.client.facade.CloudControllerException;
-import com.sap.cloudfoundry.client.facade.CloudServiceBrokerException;
 import org.cloudfoundry.multiapps.common.util.JsonUtil;
 import org.cloudfoundry.multiapps.controller.client.lib.domain.CloudServiceInstanceExtended;
 import org.cloudfoundry.multiapps.controller.core.util.OperationExecutionState;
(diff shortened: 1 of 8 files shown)


IvanBorislavovDimitrov:
instead of having the condition in every step can you extract it in:
ServiceStep::processServiceActionFailure only?

You can have 2 types of messages:
Error creating or updating optional service instance (for optional)
Error creating or updating service instance (for non-optional)

This way in the steps Update,CreateService, etc, you're just throwing the exception and its message will be added.

Thus you will not have the code duplication

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ServiceStep.java
```diff
@@ -67,15 +67,15 @@ protected abstract OperationExecutionState executeOperation(ProcessContext conte
...
                 context.setVariable(Variables.SERVICE_OFFERING, serviceInstance.getLabel());
-                throw new CloudServiceBrokerException(e.getStatusCode(), e.getStatusText(), detailedDescription);
+                throw new CloudServiceBrokerException(e.getStatusCode(), e.getStatusText(), e.getDescription());
             }
-            throw new CloudControllerException(e.getStatusCode(), e.getStatusText(), detailedDescription);
+            throw new CloudControllerException(e.getStatusCode(), e.getStatusText(), e.getDescription());
         }
-        getStepLogger().warn(e.getDescription());
+        String detailedMessage = serviceInstance.isOptional() ? Messages.ERROR_CREATING_OR_UPDATING_OPTIONAL_SERVICE_INSTANCE
+            : Messages.ERROR_CREATING_OR_UPDATING_SERVICE_INSTANCE;
```
IvanBorislavovDimitrov, line 77: Always optional true. Add the other message for not optional above
Use placeholders "{0}" instead of concatenation

IvanBorislavovDimitrov (approved):

PR #1445, opened 2024-04-02
Adopt java 17

IvanBorislavovDimitrov:
LMCROSSITXSADEPLOY-2662

### DIFF
--- a/multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationColorDetectorTest.java
+++ b/multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationColorDetectorTest.java
@@ -4,6 +4,7 @@ import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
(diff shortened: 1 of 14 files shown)


ikasarov (approved):

boyan-velinov (commented):
Change the java to 17 just in case:
https://github.com/cloudfoundry/multiapps-controller/blob/master/multiapps-controller-web/manifests/manifest.yml#L19

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationColorDetectorTest.java
```diff
@@ -135,9 +131,9 @@ void detectSingularDeployedApp3ModulesNoSuffix() {
     @Test
     void detectSingularDeployedApp3ModulesGreenBlueSuffix() {
         DeployedMta deployedMta = createMta("com.sap.sample.mta.consumer", Collections.emptySet(),
-                                            List.of(createMtaApplication("app-1", "app-1-blue", parseDate("2016-15-10")),
+                                            List.of(createMtaApplication("app-1", "app-1-blue", parseDate("2016-10-10")),
```
theghost5800, line 134: Is there a reason why month was changed?

pom.xml
```diff
@@ -16,47 +16,47 @@
...
         <log4j.version>2.23.1</log4j.version>
         <slf4j.version>2.0.12</slf4j.version>
-        <commons-io.version>2.15.1</commons-io.version>
+        <commons-io.version>2.16.1</commons-io.version>
         <commons-fileupload.version>1.5</commons-fileupload.version>
         <commons-collections4.version>4.4</commons-collections4.version>
         <snakeyaml.version>2.2</snakeyaml.version>
         <httpclient.version>4.5.14</httpclient.version>
-        <httpcore.version>4.4.13</httpcore.version>
+        <httpcore.version>4.4.16</httpcore.version>
```
theghost5800, line 30: By the way this version is from 2022 - https://mvnrepository.com/artifact/org.apache.httpcomponents/httpcore
There is a new major version 5 of httpclient with different artifact id and new package names, most probably we have to track separate task to adopt newer apache http client.

pom.xml
```diff
@@ -16,47 +16,47 @@
...
-        <liquibase.version>4.25.1</liquibase.version>
+        <liquibase.version>4.27.0</liquibase.version>
         <liquibase-slf4j.version>5.0.0</liquibase-slf4j.version>
         <cloudfoundry-client.version>2.41.0</cloudfoundry-client.version>
-        <swagger.version>1.6.9</swagger.version>
+        <swagger.version>1.6.14</swagger.version>
         <jclouds.version>2.6.0</jclouds.version>
-        <guava.version>32.0.1-jre</guava.version>
-        <resilience4j.version>1.7.1</resilience4j.version>
+        <guava.version>33.1.0-jre</guava.version>
```
theghost5800, line 45: Guava is used by jclouds and needs to check does everything works on all IaaS providers again.

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/util/ApplicationColorDetectorTest.java
IvanBorislavovDimitrov: yes 15 is invalid month and the method (in the test) was just printing stack traces for no reason and returning nulls

pom.xml
IvanBorislavovDimitrov: might give it a try with the next release, the client and core should be updated but will require some refactoring

theghost5800 (approved):

PR #1463, opened 2024-04-29
Remove failtOnCrash tag

Yavor16:


theghost5800:
As far as I know, previous modification https://github.com/cloudfoundry/multiapps-controller/commit/3816c71d99285c30fefcf2124bda9afb92cb6662 is not release yet, so this most probably must not be applied together, right?

Yavor16:
I think if they are applied in the same release won't be a problem

theghost5800 (approved):

PR #1464, opened 2024-05-10
Execution listener task

MNoeva:
#### Description: 
Refactor the "AbstractProcessExecutionListener" class so it does not use field injection.
Make sure all classes that extend this class do not use field injection as well.

LMCROSSITXSADEPLOY-2934


MNoeva:
Changed email address for the commit

PR #1477, opened 2024-05-21
Added timeout to CollectServicesInProgressStateStep

MNoeva:
#### Description: 
Async steps of CollectServicesInProgressStateStep doesn't timeout


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/CollectServicesInProgressStateStep.java
```diff
@@ -81,4 +81,9 @@ protected String getStepErrorMessage(ProcessContext context) {
         return Messages.ERROR_MONITORING_OPERATIONS_OVER_SERVICES;
     }
 
+    @Override
+    public Duration getTimeout(ProcessContext context) {
+        return context.getVariable(Variables.START_TIMEOUT);
```
IvanBorislavovDimitrov, line 86: I'd suggest creating a new variable for this step instead of using the START_TIMEOUT, it is already used improperly and soon will be removed

IvanBorislavovDimitrov (approved):

PR #1484, opened 2024-06-12
Skip ssl cf api

asalan316:
#### Description: 

The call to CF API to get [UAA Endpoint](https://github.com/cloudfoundry/multiapps-controller/blob/master/multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/configuration/UAAClientConfiguration.java#L27-L58) was possible only with SSL configured. For development reasons, it is sometimes needed to skip ssl configuration.

Change:

This PR allows users/operators to also make insecure ssl connection to CF API in order to retrieve UAA token url. This can be achieve by setting `SKIP_SSL_VALIDATION` environment variable ([already available](https://github.com/cloudfoundry/multiapps-controller/blob/master/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/ApplicationConfiguration.java#L423))


asalan316:
mistakenly - opened PR

PR #1493, opened 2024-07-04
Validate size of all uploaded parts per operation

theghost5800:


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ValidateDeployParametersStep.java
```diff
@@ -87,6 +89,25 @@ private void validateDescriptorSize(FileEntry file) {
...
+            sizeOfAllFiles = fileEntries.stream()
+                                        .mapToLong(fileEntry -> fileEntry.getSize()
+                                                                         .longValue())
+                                        .sum();
+        } catch (FileStorageException e) {
+            throw new SLException(e, MessageFormat.format(Messages.ERROR_OCURRED_DURING_VALIDATION_OF_FILES_0, e.getMessage()));
+        }
+
+        if (sizeOfAllFiles >= maxFileSizeLimit) {
+            throw new ContentException(Messages.SIZE_OF_ALL_OPERATIONS_FILES_0_EXCEEDS_MAX_UPLOAD_SIZE_1, sizeOfAllFiles, maxFileSizeLimit);
```
IvanBorislavovDimitrov, line 107: should we delete these files immediately as they only take space and deployment with them is not possible

PR #1530, opened 2024-09-13
Add namespace as suffix option

Yavor16:


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/NamespaceValidationUtil.java
```diff
@@ -1,35 +1,46 @@
...
+        this.applyNamespaceAsSuffixProcessVariable = applyNamespaceAsSuffixProcessVariable;
     }
 
     public String getNamespace() {
         return namespace;
     }
 
+    protected boolean shouldApplyNamespaceAsSuffix(final Map<String, Object> relatedParameters) {
+        Boolean applyNamespaceAsSuffixModuleLevel = MapUtil.parseBooleanFlag(relatedParameters, SupportedParameters.APPLY_NAMESPACE_AS_SUFFIX, null);
+        return Objects.requireNonNullElse(applyNamespaceAsSuffixProcessVariable,
```
ikasarov, line 40: it would be mroe readable and faster if you just add 
if (applyNamespaceAsSuffixProcessVariable != null) return applyNamespaceAsSuffixProcessVariable
at the start

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/HostValidator.java
```diff
@@ -23,11 +29,31 @@ public boolean isValid(Object routePart, final Map<String, Object> context) {
...
             return ((String) routePart).startsWith(getNamespace());
         }
         return true;
     }
 
+    private boolean isHostWithNamespaceAndSuffixValid(String routePart) {
+        if (routePart.endsWith(getNamespace())) {
+            return true;
+        } else if (routePart.endsWith(SystemParameters.IDLE_HOST_SUFFIX)) {
+            String routeStringWithoutIdleSuffix = routePart.substring(0, routePart.lastIndexOf(SystemParameters.IDLE_HOST_SUFFIX));
```
ikasarov, line 44: routeStringWithoutIdleSuffix if a bad name for all of the else statements; also a lot of the same 2 lines - maybe extract in a method 'boolean containsNamespaceBeforeSuffix(routePart, suffix)

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/NamespaceGlobalParameters.java
```diff
@@ -20,26 +20,30 @@ public NamespaceGlobalParameters(DeploymentDescriptor deploymentDescriptor) {
...
+        return getApplyNamespace(SupportedParameters.APPLY_NAMESPACE_ROUTES, Boolean.TRUE);
+
+    }
 
+    public boolean getApplyNamespaceAsSuffix() {
+        return getApplyNamespace(SupportedParameters.APPLY_NAMESPACE_AS_SUFFIX, Boolean.FALSE);
     }
 
-    private boolean getApplyNamespace(String applyNamespaceFlagName) {
+    private boolean getApplyNamespace(String applyNamespaceFlagName, boolean defaultValue) {
```
ikasarov, line 39: why is the signature 'boolean' but the constants we use on invocation is Boolean?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/NamespaceGlobalParameters.java
```diff
@@ -20,26 +20,30 @@ public NamespaceGlobalParameters(DeploymentDescriptor deploymentDescriptor) {
...
         if (!applyNamespaceGlobalLevel.isEmpty()) {
             Object value = applyNamespaceGlobalLevel.get(applyNamespaceFlagName);
             if (value != null && !(value instanceof Boolean)) {
                 throw new ContentException(NAMESPACE_PARSING_ERROR_MESSAGE, applyNamespaceFlagName);
             }
             return (boolean) applyNamespaceGlobalLevel.getOrDefault(applyNamespaceFlagName, true);
```
ikasarov, line 45: if APPLY_NAMESPACE_AS_SUFFIX exists at the global level but is 'nulll' this will return 'true' as opposed to the 'false' default value given below - is this correct behaviour?

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/NameUtil.java
```diff
@@ -73,19 +61,63 @@ public static String getNameWithProperLength(String name, int maxLength) {
...
+            name = placeNamespaceBeforeBlueGreenSuffix(name, namespaceSuffix, SystemParameters.GREEN_HOST_SUFFIX);
+        } else {
+            name += namespaceSuffix;
         }
 
-        return getNameWithProperLength(prefix + name, maxLength);
+        return name;
+    }
+
+    private static String placeNamespaceBeforeBlueGreenSuffix(String name, String namespaceSuffix, String blueGreenSuffix) {
```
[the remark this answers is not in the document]
Yavor16: The problem is when the 'name' is not shortend because it is already in the range. That means that the suffix will be left there and can happen something like test-name-idle-namespace-idle. It happened during testing so I did it this way

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/NameUtil.java
```diff
@@ -73,19 +61,63 @@ public static String getNameWithProperLength(String name, int maxLength) {
...
+        String nameWithoutShortening = name;
+        name = getNameWithProperLength(name, calculateNameLengthWithoutNamespaceAndBlueGreenSuffix(namespaceSuffix, maxLengthWithSuffix));
+
+        name = correctNameSuffix(name, nameWithoutShortening, namespaceSuffix);
+
+        return name;
+    }
+
+    private static int calculateNameLengthWithoutNamespaceAndBlueGreenSuffix(String namespaceSuffix, int maxLengthWithSuffix) {
+        return maxLengthWithSuffix - (namespaceSuffix.length() + SystemParameters.GREEN_HOST_SUFFIX.length());
```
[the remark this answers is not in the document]
Yavor16: Yes, I use because it is the longest. Will add comment

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/NameUtil.java
```diff
@@ -123,4 +155,21 @@ public static String getServiceName(Resource resource) {
                                 .get(SupportedParameters.SERVICE_NAME);
     }
 
+    public static class NameRequirements {
```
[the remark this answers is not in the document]
Yavor16: I didn't move them intentionally. Everytime I try to reformat the code they are automatically moved there

pom.xml
```diff
@@ -49,7 +49,7 @@
         <aliyun-sdk-oss.version>3.17.4</aliyun-sdk-oss.version>
         <auto-service.version>1.1.1</auto-service.version>
         <commons-codec.version>1.16.1</commons-codec.version>
-        <multiapps.version>2.24.0</multiapps.version>
+        <multiapps.version>2.25.0-SNAPSHOT</multiapps.version>
```
ikasarov, line 52: shouldn't be set to SNAPSHOT?

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/NameUtil.java
```diff
@@ -73,19 +59,62 @@ public static String getNameWithProperLength(String name, int maxLength) {
...
+            }
         }
+        return getNameWithProperLength(name, maxLength);
+    }
 
-        return getNameWithProperLength(prefix + name, maxLength);
+    private static String getNameWithNamespaceSuffix(String name, String namespace, int maxLength) {
+        String namespaceSuffix = getNamespaceSuffix(namespace);
+        String nameWithoutShortening = name;
+        name = getNameWithProperLength(name, calculateNameLengthWithoutNamespaceAndBlueGreenSuffix(namespaceSuffix, maxLength));
```
ikasarov, line 77: you can skip the above line by simply adding
String shortenedName = getNameWithProperLength(..) and don't need to change the value of 'name'

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/NameUtil.java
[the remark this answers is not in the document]
ikasarov: Does this mean that for a longer name_idle the shortening will also hash the idle part and replace it with the hash value leading the namespaced name to be 'nameHASH_namespace' without the idle in the end? Then shouldn't we extract the suffix before even shortening in the first place? That way the 'idle' or 'blue' won't get hashed in the shortening as well and will be present in a readable way and will be applied correctly.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/NameUtil.java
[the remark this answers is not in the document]
ikasarov: Intellij? 🥲

ikasarov (approved):

IvanBorislavovDimitrov (approved):

PR #1532, opened 2024-09-16
Fix log message and remove backwards logic

Yavor16:


theghost5800 (dismissed):
Do not forget to adopt released version of multiapps

theghost5800 (dismissed):

theghost5800 (approved):

PR #1553, opened 2024-10-25
Add header support for ProcessLogger

vkalapov:

### DIFF
--- a/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/services/ProcessLogger.java
+++ b/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/services/ProcessLogger.java
@@ -1,8 +1,5 @@
 package org.cloudfoundry.multiapps.controller.persistence.services;
 
-import java.util.Objects;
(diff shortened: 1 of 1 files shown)


multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/services/ProcessLogger.java
```diff
@@ -77,7 +79,13 @@ public OperationLogEntry getOperationLogEntry() {
     }
 
     private void createLogMessage(Object message, Level logLevel) {
-        logMessage = layout.toSerializable(createEvent(message, logLevel));
+        String formattedLogMessage = layout.toSerializable(createEvent((message), logLevel));
+        byte[] header = layout.getHeader();
+        if (header != null && !headerIsSet) {
```
boyan-velinov, line 84: Why this is not part of the other method?

Can we rely on having `header != null` then use it, without additional parameter `headerIsSet`?

There is one more createLogMessage method that should use header as well.

Yavor16 (approved):

PR #1598, opened 2025-02-27
Fix NPE when apps with multiple versions exists

theghost5800:


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ExistingAppsToBackupCalculator.java
```diff
@@ -74,14 +75,50 @@ private boolean isDeployedMtaBackupDescriptorMissing(ProcessContext context, Dep
...
+            return true;
+        }
+        return isBackupDescriptorMissing(context, deployedMtaVersion);
+    }
+
+    private Version getDeployedMtaVersion(DeployedMta deployedMta) {
+        Version deployedMtaVersion = deployedMta.getMetadata()
+                                                .getVersion();
+        if (deployedMtaVersion == null) {
+            deployedMtaVersion = findLiveProductizationStateVersion(deployedMta);
```
IvanBorislavovDimitrov, line 89: you can just return here

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ExistingAppsToBackupCalculator.java
```diff
@@ -74,14 +75,50 @@ private boolean isDeployedMtaBackupDescriptorMissing(ProcessContext context, Dep
...
+        }
+        return deployedMtaVersion;
+    }
+
+    private Version findLiveProductizationStateVersion(DeployedMta deployedMta) {
+        for (DeployedMtaApplication deployedMtaApplication : deployedMta.getApplications()) {
```
IvanBorislavovDimitrov, line 95: what will happen if there is a deployed MTA with services only that mismatch

IvanBorislavovDimitrov (approved):

PR #1607, opened 2025-03-12
Add warning for unsupported parameters

vkalapov:

### DIFF
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStep.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStep.java
@@ -1,8 +1,12 @@
 package org.cloudfoundry.multiapps.controller.process.steps;
 
+import java.text.MessageFormat;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
 import org.cloudfoundry.multiapps.controller.core.cf.CloudHandlerFactory;
 import org.cloudfoundry.multiapps.controller.core.helpers.MtaDescriptorMerger;
 import org.cloudfoundry.multiapps.controller.persistence.dto.BackupDescriptor;
@@ -10,15 +14,15 @@ import org.cloudfoundry.multiapps.controller.persistence.dto.ImmutableBackupDesc
 import org.cloudfoundry.multiapps.controller.persistence.services.DescriptorBackupService;
 import org.cloudfoundry.multiapps.controller.process.Messages;
 import org.cloudfoundry.multiapps.controller.process.util.NamespaceGlobalParameters;
+import org.cloudfoundry.multiapps.controller.process.util.UnsupportedParameterFinder;
 import org.cloudfoundry.multiapps.controller.process.variables.Variables;
 import org.cloudfoundry.multiapps.mta.model.DeploymentDescriptor;
 import org.cloudfoundry.multiapps.mta.model.ExtensionDescriptor;
 import org.cloudfoundry.multiapps.mta.model.Platform;
+import org.cloudfoundry.multiapps.mta.resolvers.ReferenceContainer;
+import org.cloudfoundry.multiapps.mta.resolvers.ReferencesFinder;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Scope;
-
-import jakarta.inject.Inject;
-import jakarta.inject.Named;
 
 @Named("mergeDescriptorsStep")
 @Scope(BeanDefinition.SCOPE_PROTOTYPE)
@@ -26,6 +30,9 @@ public class MergeDescriptorsStep extends SyncFlowableStep {
 
     @Inject
     private DescriptorBackupService descriptorBackupService;
+
+    @Inject
+    private UnsupportedParameterFinder unsupportedParameterFinder;
 
     protected MtaDescriptorMerger getMtaDescriptorMerger(CloudHandlerFactory factory, Platform platform) {
         return new MtaDescriptorMerger(factory, platform, getStepLogger());
@@ -42,9 +49,21 @@ public class MergeDescriptorsStep extends SyncFlowableStep {
                                                                                                  extensionDescriptors);
         context.setVariable(Variables.DEPLOYMENT_DESCRIPTOR, descriptor);
 
+        warnForUnsupportedParameters(descriptor);
+
         backupDeploymentDescriptor(context, descriptor);
         getStepLogger().debug(Messages.DESCRIPTORS_MERGED);
         return StepPhase.DONE;
+    }
+
+    private void warnForUnsupportedParameters(DeploymentDescriptor descriptor) {
+        List<ReferenceContainer> references = new ReferencesFinder().getAllReferences(descriptor);
+        Map<String, List<String>> unsupportedParameters = unsupportedParameterFinder.findUnsupportedParameters(descriptor,
+                                                                                                               references);
+        if (!unsupportedParameters.isEmpty()) {
+            getStepLogger().warn(MessageFormat.format(Messages.PARAMETERS_0_ARE_NOT_SUPPORTED_OR_REFERENCED_BY_ANY_OTHER_ENTITIES,
+                                                      unsupportedParameters));
+        }
     }
 
     private void backupDeploymentDescriptor(ProcessContext context, DeploymentDescriptor descriptor) {
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ModuleHooksAggregator.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ModuleHooksAggregator.java
@@ -28,17 +28,17 @@ public class ModuleHooksAggregator {
 
     private List<Hook> determineHooksForExecution(Map<String, List<String>> alreadyExecutedHooks,
                                                   List<HookPhase> hookPhasesForCurrentStepPhase) {
-        List<Hook> moduleHooksToExecuteOnCurrentStepPhase = collectHooksWithPhase(moduleToDeploy, hookPhasesForCurrentStepPhase);
+        List<Hook> moduleHooksToExecuteOnCurrentStepPhase = collectHooksWithPhase(hookPhasesForCurrentStepPhase);
         return getHooksForExecution(alreadyExecutedHooks, moduleHooksToExecuteOnCurrentStepPhase, hookPhasesForCurrentStepPhase);
     }
 
-    private List<Hook> collectHooksWithPhase(Module moduleToDeploy, List<HookPhase> hookPhasesForCurrentStepPhase) {
-        return getModuleHooks(moduleToDeploy).stream()
-                                             .filter(hook -> shouldCollectHook(hook.getPhases(), hookPhasesForCurrentStepPhase))
-                                             .collect(Collectors.toList());
+    private List<Hook> collectHooksWithPhase(List<HookPhase> hookPhasesForCurrentStepPhase) {
+        return getModuleHooks().stream()
+                               .filter(hook -> shouldCollectHook(hook.getPhases(), hookPhasesForCurrentStepPhase))
+                               .collect(Collectors.toList());
     }
 
-    private List<Hook> getModuleHooks(Module module) {
+    private List<Hook> getModuleHooks() {
         return moduleToDeploy.getMajorSchemaVersion() < MAJOR_SCHEMA_VERSION_THREE ? Collections.emptyList() : moduleToDeploy.getHooks();
     }
 
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ModuleHooksAggregator.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ModuleHooksAggregator.java
@@ -1,4 +1,3 @@
-
 package org.cloudfoundry.multiapps.controller.process.util;
 
 import java.util.Collections;
@@ -13,9 +12,7 @@ import org.cloudfoundry.multiapps.mta.model.Module;
 import org.flowable.engine.delegate.DelegateExecution;
 
 public class ModuleHooksAggregator {
-
     private static final int MAJOR_SCHEMA_VERSION_THREE = 3;
-
     private final DelegateExecution execution;
     private final Module moduleToDeploy;
 
@@ -41,7 +38,7 @@ public class ModuleHooksAggregator {
                                              .collect(Collectors.toList());
     }
 
-    private List<Hook> getModuleHooks(Module moduleToDeploy) {
+    private List<Hook> getModuleHooks(Module module) {
         return moduleToDeploy.getMajorSchemaVersion() < MAJOR_SCHEMA_VERSION_THREE ? Collections.emptyList() : moduleToDeploy.getHooks();
     }
 
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SchemaVersionUtils.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SchemaVersionUtils.java
@@ -1,0 +1,14 @@
+package org.cloudfoundry.multiapps.controller.process.util;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+
+public class SchemaVersionUtils {
+
+    private static final int MAJOR_SCHEMA_VERSION_THREE = 3;
+
+    public static <T, R> List<R> getEntityData(T entity, Function<T, Integer> versionGetter, Function<T, List<R>> dataGetter) {
+        return versionGetter.apply(entity) < MAJOR_SCHEMA_VERSION_THREE ? Collections.emptyList() : dataGetter.apply(entity);
+    }
+}
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
@@ -1,0 +1,56 @@
+package org.cloudfoundry.multiapps.controller.process.util;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+import org.cloudfoundry.multiapps.mta.model.DeploymentDescriptor;
+import org.cloudfoundry.multiapps.mta.resolvers.CustomParameterContainer;
+import org.cloudfoundry.multiapps.mta.resolvers.ReferenceContainer;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.context.annotation.Scope;
+
+@Named("UnsupportedParameterFinder")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class UnsupportedParameterFinder {
+
+    @Inject
+    private SupportedParameterChecker supportedParameterChecker;
+
+    @Inject
+    private ReferenceContainerMatcher referenceMatcher;
+
+    @Inject
+    private CustomParameterContainerProcessor containerProcessor;
+    
+    public Map<String, List<String>> findUnsupportedParameters(DeploymentDescriptor descriptor, List<ReferenceContainer> references) {
+        List<CustomParameterContainer> unknownParametersContainer = supportedParameterChecker.getCustomParameters(descriptor);
+        return collectUnsupportedParametersByOwner(unknownParametersContainer, references);
+    }
+
+    private Map<String, List<String>> collectUnsupportedParametersByOwner(List<CustomParameterContainer> containers,
+                                                                          List<ReferenceContainer> references) {
+        return containers.stream()
+                         .map(container -> containerProcessor.processCustomParameterContainer(container, references, referenceMatcher))
+                         .filter(entry -> !entry.getValue()
+                                                .isEmpty())
+                         .collect(unsupportedParametersCollector());
+    }
+
+    private Collector<Map.Entry<String, List<String>>, ?, Map<String, List<String>>> unsupportedParametersCollector() {
+        return Collectors.toMap(
+            Map.Entry::getKey,
+            Map.Entry::getValue,
+            this::mergeParameterLists
+        );
+    }
+
+    private List<String> mergeParameterLists(List<String> existing, List<String> replacement) {
+        existing.addAll(replacement);
+        return existing;
+    }
+
+}
--- a/multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStepTest.java
+++ b/multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStepTest.java
@@ -1,10 +1,5 @@
 package org.cloudfoundry.multiapps.controller.process.steps;
 
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.when;
-
 import java.util.Collections;
 
 import org.cloudfoundry.multiapps.common.ContentException;
@@ -13,12 +8,21 @@ import org.cloudfoundry.multiapps.common.test.Tester.Expectation;
 import org.cloudfoundry.multiapps.controller.core.cf.CloudHandlerFactory;
 import org.cloudfoundry.multiapps.controller.core.helpers.MtaDescriptorMerger;
 import org.cloudfoundry.multiapps.controller.core.test.DescriptorTestUtil;
+import org.cloudfoundry.multiapps.controller.process.util.UnsupportedParameterFinder;
 import org.cloudfoundry.multiapps.controller.process.variables.Variables;
 import org.cloudfoundry.multiapps.mta.model.DeploymentDescriptor;
 import org.cloudfoundry.multiapps.mta.model.Platform;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mock;
+import org.mockito.Mockito;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 class MergeDescriptorsStepTest extends SyncFlowableStepTest<MergeDescriptorsStep> {
 
@@ -39,6 +43,9 @@ class MergeDescriptorsStepTest extends SyncFlowableStepTest<MergeDescriptorsStep
     @Mock
     private MtaDescriptorMerger merger;
 
+    @Mock
+    private UnsupportedParameterFinder unsupportedParameterFinder;
+
     @BeforeEach
     public void setUp() {
         prepareContext();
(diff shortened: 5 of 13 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ReferenceFinder.java
```diff
@@ -0,0 +1,164 @@
...
+
+import jakarta.inject.Named;
+
+@Named("referenceFinder")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class ReferenceFinder {
+
+    private String parameterToFind;
+
+    public ReferenceFinder() {
```
s-yonkov-yonkov, line 28: Why this constructor is explicitly added?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/CollectSystemParametersStep.java
```diff
@@ -61,7 +61,6 @@ protected StepPhase executeStepInternal(ProcessContext context, boolean reserveT
         SystemParameters systemParameters = createSystemParameters(context, defaultDomainName, reserveTemporaryRoutes, descriptor);
```
s-yonkov-yonkov, line 61: Why this class is part of the PR?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SchemaVersionUtils.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+
+public class SchemaVersionUtils {
+
+    private static final int MAJOR_SCHEMA_VERSION_THREE = 3;
+
+    public static <T> List<?> getEntityData(T entity, Function<T, Integer> versionGetter, Function<T, List<?>> dataGetter) {
```
s-yonkov-yonkov, line 11: Please reformat this class -> 

`    
public static <T, R> List<R> getEntityData(T entity, Function<T, Integer> versionGetter, Function<T, List<R>> dataGetter) {
        return versionGetter.apply(entity) < MAJOR_SCHEMA_VERSION_THREE ? Collections.emptyList() : dataGetter.apply(entity);
    }
`

- By adding the Generics to the List we ensure that we always get the correct list type

- No need for casting on every method call

- Safer, clearer and prevents accidental type errors

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ModuleHooksAggregator.java
```diff
@@ -41,8 +38,8 @@ private List<Hook> collectHooksWithPhase(Module moduleToDeploy, List<HookPhase>
                                              .collect(Collectors.toList());
     }
 
-    private List<Hook> getModuleHooks(Module moduleToDeploy) {
-        return moduleToDeploy.getMajorSchemaVersion() < MAJOR_SCHEMA_VERSION_THREE ? Collections.emptyList() : moduleToDeploy.getHooks();
+    private List<Hook> getModuleHooks(Module module) {
+        return (List<Hook>) SchemaVersionUtils.getEntityData(module, Module::getMajorSchemaVersion, Module::getHooks);
```
s-yonkov-yonkov, line 42: Update SchemaVersionUtils and remove the cast

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ReferenceFinder.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private void setParameterToFind(String parameterToFind) {
```
s-yonkov-yonkov, line 31: Consider refactoring the design to pass the parameter as an argument through the call hierarchy instead of using an instance field.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ReferenceFinder.java
```diff
@@ -0,0 +1,164 @@
...
+                      .stream()
+                      .anyMatch(this::hasReferenceInValue);
+        } else if (value instanceof List<?> list) {
+            return list.stream()
+                       .anyMatch(this::hasReferenceInValue);
+        }
+        return false;
+    }
+
+    private boolean referencePatternMatches(List<String> patterns, Object value) {
```
s-yonkov-yonkov, line 102: this method compiles regex patterns on every invocation, precompiled Pattern instances can be cached via a static final List<Pattern> so that the compilation is done only once:

`
private static final List<Pattern> COMPILED_PATTERNS = Arrays.asList(
    Pattern.compile(ReferencePattern.PLACEHOLDER.getPattern()),
    Pattern.compile(ReferencePattern.SHORT.getPattern()),
    Pattern.compile(ReferencePattern.FULLY_QUALIFIED.getPattern())
);
`

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
```diff
@@ -0,0 +1,115 @@
...
+import org.springframework.context.annotation.Scope;
+
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
+@Named("supportedParametersChecker")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class SupportedParametersChecker {
+
+    private List<String> unknownParameters;
```
s-yonkov-yonkov, line 25: Why this is declared as a class field?

s-yonkov-yonkov:
Check the failing checks

s-yonkov-yonkov (changes requested):
Please review comments

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ReferenceFinder.java
vkalapov: Leftover from previous changes during development. Removed.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SchemaVersionUtils.java
vkalapov: Great suggestion! Done.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ModuleHooksAggregator.java
vkalapov: Done. Also removed casts in in other calls of SchemaVersionUtils.getEntityData.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/CollectSystemParametersStep.java
vkalapov: Changes were done to it during development, which were removed, leaving only the changes to formatting to be committed. I've removed the formatting changes, so it doesn't appear as a file change.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ReferenceFinder.java
vkalapov: Done

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
```diff
@@ -0,0 +1,115 @@
...
+import org.cloudfoundry.multiapps.mta.model.Resource;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.context.annotation.Scope;
+
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
+@Named("supportedParametersChecker")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class SupportedParametersChecker {
```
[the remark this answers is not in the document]
vkalapov: In practice if the Resource, Hook etc. parameters are null, the operation will fail before that in a previous step. So I don't believe a check for that is needed.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ReferenceFinder.java
vkalapov: This would introduce additional complexity to the method calls.  What are the concerns for using an instance field in this case?

vkalapov:
Tests fail due to missing version of the multiapps dependency, which is not released yet.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
```diff
@@ -0,0 +1,114 @@
...
+        checkRequiredDependencyParameters(hook.getRequiredDependencies());
+    }
+
+    private void checkRequiredDependencyParameters(List<RequiredDependency> requiredDependencies) {
+        requiredDependencies.forEach(requiredDependency -> checkParameters(getDependencyParameters(), requiredDependency.getParameters()));
+    }
+
+    private void checkParameters(Set<String> supportedParameters, Map<String, Object> parameters) {
+        for (String parameterName : parameters.keySet()) {
+            if (!supportedParameters.contains(parameterName) && !getGlobalParameters().contains(parameterName)) {
```
ikasarov, line 100: what is the point of allowing global parameters to pass the checks on all levels? e.g. APPS_START_TIMEOUT would pass these checks if defined at any level, even where it's not appropriate

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
```diff
@@ -0,0 +1,114 @@
...
+
+    protected Set<String> getModuleHookParameters() {
+        return SupportedParameters.MODULE_HOOK_PARAMETERS;
+    }
+
+    protected Set<String> getResourceParameters() {
+        return SupportedParameters.RESOURCE_PARAMETERS;
+    }
+
+    protected Set<String> getGlobalParameters() {
```
ikasarov, line 80: don't see a need to wrap there SupportedParameters static sets in method calls - just use thes as constants

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
```diff
@@ -0,0 +1,114 @@
...
+        for (String parameter : findUnsupportedParameters(descriptor)) {
+            if (!this.finder.isParameterReferenced(descriptor, parameter)) {
+                unsupportedAndNotReferencedParameters.add(parameter);
+            }
+        }
+        return unsupportedAndNotReferencedParameters;
+    }
+
+    private List<String> findUnsupportedParameters(DeploymentDescriptor descriptor) {
+        this.unknownParameters = new ArrayList<>();
```
ikasarov, line 41: don't like using void methods for filling this list, but initializing the list at the start of the method, instead of the constructor. Maybe just wrap the unknownParameters.add in an instance method taht initializes if list in null and adds if not. And make the findUnsupportedParameters also return void and just use a getUnknownParameters() after all the checks to return the complete list.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
```diff
@@ -0,0 +1,114 @@
...
+import org.cloudfoundry.multiapps.mta.model.Resource;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.context.annotation.Scope;
+
+import jakarta.inject.Inject;
+import jakarta.inject.Named;
+
+@Named("supportedParametersChecker")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class SupportedParametersChecker {
```
ikasarov, line 23: Could this whole recursive parameter checking be done via the Visitor pattern we already use for the Validators? (not saying it's a good idea to do it, but it might be better?)

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
```diff
@@ -0,0 +1,114 @@
...
+                  .forEach(this::checkResourceParameters);
+        checkParameters(getGlobalParameters(), descriptor.getParameters());
+        return unknownParameters;
+    }
+
+    private void checkModuleParameters(Module module) {
+        checkParameters(getModuleParameters(), module.getParameters());
+        getModuleHooks(module).forEach(this::checkHookParameters);
+        module.getRequiredDependencies()
+              .forEach(dependency -> checkParameters(getDependencyParameters(), dependency.getParameters()));
```
ikasarov, line 54: invoking getDependencyParameters() twice here seems bad, just save it in a variable

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ReferenceFinder.java
```diff
@@ -0,0 +1,157 @@
...
+
+import jakarta.inject.Named;
+
+@Named("referenceFinder")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class ReferenceFinder {
+
+    private String parameterToFind;
+
+    private static final List<Pattern> COMPILED_PATTERNS = Arrays.asList(Pattern.compile(ReferencePattern.PLACEHOLDER.getPattern()),
```
ikasarov, line 28: consider changing ReferencePattern to just have the compiled patterns saved and get those directly. Besides that - no need to persist another copy of the enum values here - just use the directly ReferencePattern.values()

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStep.java
```diff
@@ -42,11 +45,21 @@ protected StepPhase executeStep(ProcessContext context) {
...
 
+        warnForUnknownParameters(descriptor);
+
         backupDeploymentDescriptor(context, descriptor);
         getStepLogger().debug(Messages.DESCRIPTORS_MERGED);
         return StepPhase.DONE;
     }
 
+    private void warnForUnknownParameters(DeploymentDescriptor descriptor) {
+        List<String> unknownParameters = supportedParametersChecker.getUnknownParameters(descriptor);
```
ikasarov, line 56: to me, it seems more logical to initialize the supportedParametersChecker with the descriptor variable. Since it doesn't make sense to invoke .getUnknownParameters() with different descriptors for the same instance. Also - this would probably mean we don't need to inject it.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
vkalapov: The idea is to have the option to override them in an inheritor class, in case the parameters need to be extended.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
vkalapov: Copied some of the parameters to be also in supported module parameters and removed the additional global parameters check.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/SupportedParametersChecker.java
vkalapov: Done

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStep.java
vkalapov: Leaving as it is, as we agreed in our internal discussion.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public List<String> findUnsupportedParameters(DeploymentDescriptor descriptor) {
+        List<String> unknownParametersContainer = findUnsupportedParametersInDescriptor(descriptor);
+        Set<String> references = findReferencesInDescriptor(descriptor);
+        return filterUnsupportedButReferencedParameters(unknownParametersContainer, references);
+    }
+
+    private List<String> findUnsupportedParametersInDescriptor(DeploymentDescriptor descriptor) {
+        List<String> unknownParametersContainer = new ArrayList<>();
+        supportedParameterChecker.fillListWithUnmatched(descriptor, unknownParametersContainer);
```
ikasarov, line 31: maybe make the fillListWithUnmatched() method just init and return the list internally, same with the .fillWithReferences below. Since we contruct a new instance for the check anyways, no need to pass an empty list to be filled inside

ikasarov: maybe also rename the checks' method to be more descriptive - e.g. supportedParameterChecker.getCustomParameters
ReferencesFinder().getAllReferences

vkalapov: Done

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStep.java
```diff
@@ -42,11 +49,23 @@ protected StepPhase executeStep(ProcessContext context) {
...
         context.setVariable(Variables.DEPLOYMENT_DESCRIPTOR, descriptor);
 
+        warnForUnknownParameters(descriptor);
+
         backupDeploymentDescriptor(context, descriptor);
         getStepLogger().debug(Messages.DESCRIPTORS_MERGED);
         return StepPhase.DONE;
     }
 
+    private void warnForUnknownParameters(DeploymentDescriptor descriptor) {
```
ikasarov, line 59: rename method to 'warnForUnsupportedParameters' just to keep it consistent

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ModuleHooksAggregator.java
```diff
@@ -41,8 +38,8 @@ private List<Hook> collectHooksWithPhase(Module moduleToDeploy, List<HookPhase>
                                              .collect(Collectors.toList());
     }
 
-    private List<Hook> getModuleHooks(Module moduleToDeploy) {
-        return moduleToDeploy.getMajorSchemaVersion() < MAJOR_SCHEMA_VERSION_THREE ? Collections.emptyList() : moduleToDeploy.getHooks();
+    private List<Hook> getModuleHooks(Module module) {
+        return SchemaVersionUtils.getEntityData(module, Module::getMajorSchemaVersion, Module::getHooks);
```
ikasarov, line 42: I don't really like this change - makes it harder to read and doesn't seem to be overridden anywhere

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
```diff
@@ -0,0 +1,113 @@
...
+    private SupportedParameterChecker supportedParameterChecker;
+
+    private static final String GLOBAL_PARAMETER = "global-parameter";
+
+    public Map<String, List<String>> findUnsupportedParameters(DeploymentDescriptor descriptor, List<ReferenceContainer> references) {
+        List<CustomParameterContainer> unknownParametersContainer = findUnsupportedParametersInDescriptor(descriptor);
+        return collectUnsupportedParametersByOwner(unknownParametersContainer, references);
+    }
+
+    private List<CustomParameterContainer> findUnsupportedParametersInDescriptor(DeploymentDescriptor descriptor) {
```
ikasarov, line 32: unnecessary method

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
```diff
@@ -0,0 +1,113 @@
...
+        return supportedParameterChecker.getCustomParameters(descriptor);
+    }
+
+    private Map<String, List<String>> collectUnsupportedParametersByOwner(List<CustomParameterContainer> containers,
+                                                                          List<ReferenceContainer> references) {
+        return containers.stream()
+                         .map(container -> toFilteredEntry(container, references))
+                         .filter(this::hasUnsupportedParameters)
```
ikasarov, line 40: this method is also maybe unnecessary and could be replaced with a !isEmpty()

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ModuleHooksAggregator.java
vkalapov: Reverted.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
vkalapov: Done

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
vkalapov: Removed

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
```diff
@@ -0,0 +1,105 @@
...
+    private Map<String, List<String>> collectUnsupportedParametersByOwner(List<CustomParameterContainer> containers,
+                                                                          List<ReferenceContainer> references) {
+        return containers.stream()
+                         .map(container -> toFilteredEntry(container, references))
+                         .filter(entry -> !entry.getValue()
+                                                .isEmpty())
+                         .collect(unsupportedParametersCollector());
+    }
+
+    private Map.Entry<String, List<String>> toFilteredEntry(CustomParameterContainer container, List<ReferenceContainer> references) {
```
s-yonkov-yonkov, line 41: The name is not very clear maybe groupUnreferencedParamsByOwner() is better

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
```diff
@@ -0,0 +1,105 @@
...
+    private boolean hasMatchingReference(ReferenceContainer referenceContainer,
+                                         CustomParameterContainer container,
+                                         String customParameter) {
+        return referenceContainer.getParameters()
+                                 .stream()
+                                 .anyMatch(reference -> isReferenceNameMatched(reference, customParameter)
+                                     && isReferenceDependencyMatched(reference, container));
+    }
+
+    private boolean isReferenceOwnerCompatible(String referenceOwner, String parameterOwner) {
```
s-yonkov-yonkov, line 80: doesReferenceMatchOwner() could be better

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
```diff
@@ -0,0 +1,105 @@
...
+                                                 CustomParameterContainer container,
+                                                 String customParameter) {
+        if (!hasMatchingReference(referenceContainer, container, customParameter)) {
+            return false;
+        }
+
+        return isReferenceOwnerCompatible(referenceContainer.getReferenceOwner(), container.getParameterOwner());
+    }
+
+    private boolean hasMatchingReference(ReferenceContainer referenceContainer,
```
s-yonkov-yonkov, line 71: consider isReferenceMatchingParameter()

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
```diff
@@ -0,0 +1,105 @@
...
+@Named("UnsupportedParameterFinder")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class UnsupportedParameterFinder {
+
+    @Inject
+    private SupportedParameterChecker supportedParameterChecker;
+
+    private static final String GLOBAL_PARAMETER = "global-parameter";
+
+    public Map<String, List<String>> findUnsupportedParameters(DeploymentDescriptor descriptor, List<ReferenceContainer> references) {
```
s-yonkov-yonkov, line 27: The entire logic is heavily nested and hard to read, do you think that it could be simplified?

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStepTest.java
```diff
@@ -53,6 +58,7 @@ private void prepareContext() {
 
     @Test
     void testExecute1() {
+        when(unsupportedParameterFinder.findUnsupportedParameters(Mockito.any(), Mockito.any())).thenReturn(Collections.emptyMap());
```
s-yonkov-yonkov, line 61: Does this code test something?

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStepTest.java
```diff
@@ -65,6 +71,7 @@ void testExecute1() {
 
     @Test
     void testExecute2() {
+        when(unsupportedParameterFinder.findUnsupportedParameters(Mockito.any(), Mockito.any())).thenReturn(Collections.emptyMap());
```
s-yonkov-yonkov, line 74: Does this code test something?

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStepTest.java
vkalapov: Yes, I've added a check to test whether the method is called. The warning message should be checked by the integration tests, but I don't think its a bad idea to have a simple unit test check to catch it earlier.

s-yonkov-yonkov (dismissed):

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/UnsupportedParameterFinder.java
```diff
@@ -61,23 +62,23 @@ private boolean isParameterReferenced(List<ReferenceContainer> references,
     private boolean isReferenceContainerMatching(ReferenceContainer referenceContainer,
                                                  CustomParameterContainer container,
                                                  String customParameter) {
-        if (!hasMatchingReference(referenceContainer, container, customParameter)) {
+        if (!isReferenceMatchingParameter(referenceContainer, container, customParameter)) {
             return false;
         }
 
-        return isReferenceOwnerCompatible(referenceContainer.getReferenceOwner(), container.getParameterOwner());
+        return doesReferenceMatchOwner(referenceContainer.getReferenceOwner(), container.getParameterOwner());
```
ikasarov: rename to 'is...' to be consistent

s-yonkov-yonkov (approved):

ikasarov (approved):

PR #1682, opened 2025-08-15
Enable OS on CCEE

Yavor16:


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/ApplicationConfiguration.java
```diff
@@ -666,6 +677,12 @@ private Long getMaxUploadSizeFromEnvironment() {
         return value;
     }
 
+    private Set<String> getObjectStoreRegionsFromEnvironment() {
+        String value = environment.getString(OBJECTSTORE_REGIONS, Strings.EMPTY);
+        logEnvironmentVariable(OBJECTSTORE_REGIONS, Messages.OBJECTSTORE_REGIONS_ARE, value);
+        return Set.of(value.split(","));
```
IvanBorislavovDimitrov, line 683: does this function always return at least 1 element, even though it is empty?

IvanBorislavovDimitrov (approved):

PR #1687, opened 2025-08-21
Revert "fix UPS defined as existing service resources are not working"

stiv03:


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/util/CloudModelBuilderUtil.java
```diff
@@ -76,6 +78,11 @@ public static ResourceType getResourceType(Map<String, Object> properties) {
         return ResourceType.get(type);
     }
 
+    public static boolean isExistingUserProvidedService(Resource resource, CloudControllerClient client) {
+        CloudServiceInstance serviceInstance = client.getServiceInstanceWithoutAuxiliaryContent(NameUtil.getServiceName(resource), false);
+        return serviceInstance != null && serviceInstance.isUserProvided();
+    }
```
IvanBorislavovDimitrov, line 84: This function does too many things, gets the service instance, checks for null, for user provided, etc, got the idea.
You can remove it and make proper checks in the function where this function is invoked and split them in small functions

IvanBorislavovDimitrov (approved):

PR #1812, opened 2026-04-03
Add phase-configs support to Hook model for target-app resolution

stiv03:
[LMCROSSITXSADEPLOY-3038](https://jira.tools.sap/browse/LMCROSSITXSADEPLOY-3038)

### DIFF
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
@@ -218,8 +218,9 @@ public class SupportedParameters {
                                                                    SERVICE_BINDING_CONFIG, DELETE_SERVICE_KEY_AFTER_DEPLOYMENT);
 
     public static final String HOOK_TARGET_APP = "hook-target-app";
+    public static final String PHASES_CONFIG = "phases-config";
 
-    public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES, HOOK_TARGET_APP);
+    public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES, HOOK_TARGET_APP, PHASES_CONFIG);
 
     public static final Set<String> CONFIGURATION_REFERENCE_PARAMETERS = Set.of(PROVIDER_NID, PROVIDER_ID, TARGET, VERSION,
                                                                                 DEPRECATED_CONFIG_MTA_ID, DEPRECATED_CONFIG_MTA_VERSION,
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
@@ -217,7 +217,9 @@ public class SupportedParameters {
     public static final Set<String> DEPENDENCY_PARAMETERS = Set.of(BINDING_NAME, ENV_VAR_NAME, VISIBILITY, USE_LIVE_ROUTES,
                                                                    SERVICE_BINDING_CONFIG, DELETE_SERVICE_KEY_AFTER_DEPLOYMENT);
 
-    public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES);
+    public static final String HOOK_TARGET_APP = "hook-target-app";
+
+    public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES, HOOK_TARGET_APP);
 
     public static final Set<String> CONFIGURATION_REFERENCE_PARAMETERS = Set.of(PROVIDER_NID, PROVIDER_ID, TARGET, VERSION,
                                                                                 DEPRECATED_CONFIG_MTA_ID, DEPRECATED_CONFIG_MTA_VERSION,
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
@@ -127,6 +127,7 @@ public class SupportedParameters {
     public static final String SERVICE_BROKER_SPACE_SCOPED = "service-broker-space-scoped";
     public static final String MODULE_CONFIG = "config";
     public static final String HOOK_REQUIRES = "requires";
+    public static final String PHASES_CONFIG = "phases-config";
 
     // Required dependency parameters:
     public static final String SERVICE_BINDING_CONFIG = "config";
@@ -216,8 +217,6 @@ public class SupportedParameters {
 
     public static final Set<String> DEPENDENCY_PARAMETERS = Set.of(BINDING_NAME, ENV_VAR_NAME, VISIBILITY, USE_LIVE_ROUTES,
                                                                    SERVICE_BINDING_CONFIG, DELETE_SERVICE_KEY_AFTER_DEPLOYMENT);
-
-    public static final String PHASES_CONFIG = "phases-config";
 
     public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES, PHASES_CONFIG);
 
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
@@ -217,10 +217,9 @@ public class SupportedParameters {
     public static final Set<String> DEPENDENCY_PARAMETERS = Set.of(BINDING_NAME, ENV_VAR_NAME, VISIBILITY, USE_LIVE_ROUTES,
                                                                    SERVICE_BINDING_CONFIG, DELETE_SERVICE_KEY_AFTER_DEPLOYMENT);
 
-    public static final String HOOK_TARGET_APP = "hook-target-app";
     public static final String PHASES_CONFIG = "phases-config";
 
-    public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES, HOOK_TARGET_APP, PHASES_CONFIG);
+    public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES, PHASES_CONFIG);
 
     public static final Set<String> CONFIGURATION_REFERENCE_PARAMETERS = Set.of(PROVIDER_NID, PROVIDER_ID, TARGET, VERSION,
                                                                                 DEPRECATED_CONFIG_MTA_ID, DEPRECATED_CONFIG_MTA_VERSION,
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
@@ -150,6 +150,7 @@ public class Messages {
     public static final String ERROR_PREPARING_TO_RESTART_SERVICE_BROKER_SUBSCRIBERS = "Error preparing to restart service broker subscribers";
     public static final String ERROR_EXECUTING_TASK_0_ON_APP_1 = "Execution of task \"{0}\" on application \"{1}\" failed.";
     public static final String DUPLICATE_PHASE_IN_PHASES_CONFIG = "Duplicate phase \"{0}\" in \"phases-config\" of hook \"{1}\". Only one target-app per phase is supported.";
+    public static final String INVALID_PHASES_CONFIG_NOT_A_LIST = "Parameter \"phases-config\" of hook \"{0}\" must be a list.";
     public static final String ERROR_DETECTING_COMPONENTS_TO_UNDEPLOY = "Error detecting components to undeploy";
     public static final String ERROR_DELETING_IDLE_ROUTES = "Error deleting idle routes";
     public static final String ERROR_CREATING_SERVICE_BROKERS = "Error creating service brokers";
@@ -410,6 +411,7 @@ public class Messages {
     public static final String WILL_NOT_REBIND_APP_TO_SERVICE_SAME_PARAMETERS = "Service instance \"{0}\" will not be rebound to application \"{1}\" because the binding parameters are not modified";
     public static final String SERVICE_BROKER_DOES_NOT_EXIST = "Service broker with name \"{0}\" does not exist";
     public static final String EXECUTING_HOOK_0 = "Executing hook \"{0}\"";
+    public static final String SKIPPING_HOOK_TASK_NO_LIVE_APP = "Skipping hook task on application \"{0}\" with target-app \"live\": no live application exists yet (initial deployment)";
     public static final String WAITING_PREVIOUS_OPERATIONS_TO_FINISH = "Waiting for previous service operations to finish...";
     public static final String ASYNC_OPERATION_FOR_SERVICE_BROKER_FINISHED = "Async operation for service broker \"{0}\" has finished";
     public static final String STARTING_INCREMENTAL_APPLICATION_INSTANCE_UPDATE_FOR_0 = "Starting incremental application instance update for \"{0}\"...";
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
@@ -149,6 +149,7 @@ public class Messages {
     public static final String ERROR_PREPARING_TO_EXECUTE_TASKS_ON_APP = "Error preparing to execute tasks on application \"{0}\"";
     public static final String ERROR_PREPARING_TO_RESTART_SERVICE_BROKER_SUBSCRIBERS = "Error preparing to restart service broker subscribers";
     public static final String ERROR_EXECUTING_TASK_0_ON_APP_1 = "Execution of task \"{0}\" on application \"{1}\" failed.";
+    public static final String DUPLICATE_PHASE_IN_PHASES_CONFIG = "Duplicate phase \"{0}\" in \"phases-config\" of hook \"{1}\". Only one target-app per phase is supported.";
     public static final String ERROR_DETECTING_COMPONENTS_TO_UNDEPLOY = "Error detecting components to undeploy";
     public static final String ERROR_DELETING_IDLE_ROUTES = "Error deleting idle routes";
     public static final String ERROR_CREATING_SERVICE_BROKERS = "Error creating service brokers";
--- a/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
+++ b/multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
@@ -215,6 +215,7 @@ public class Messages {
     public static final String CANNOT_RETRIEVE_PARAMETERS_OF_BINDING_BETWEEN_APPLICATION_0_AND_SERVICE_INSTANCE_1 = "Cannot retrieve parameters of binding between application \"{0}\" and service instance \"{1}\"";
     public static final String CANNOT_RETRIEVE_PARAMETERS_OF_BINDING_BETWEEN_APPLICATION_0_AND_SERVICE_INSTANCE_1_FIX = "Cannot retrieve parameters of binding between application \"{0}\" and service instance \"{1}\". Got 502.";
     public static final String CANNOT_RETRIEVE_INSTANCE_OF_SERVICE = "Cannot retrieve service instance of service \"{0}\"";
+    public static final String HOOK_PHASE_BLUE_GREEN_BEFORE_UNMAP_ROUTES_IDLE_USED = "Hook \"{0}\" uses deprecated phase \"blue-green.application.before-unmap-routes.idle\" which is unreachable and will be removed in a future version";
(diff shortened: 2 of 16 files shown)


multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStep.java
```diff
@@ -59,13 +67,40 @@ protected StepPhase executeStep(ProcessContext context) {
                                                                                                                            .toList());
         context.setVariable(Variables.DEPLOYMENT_DESCRIPTOR, descriptor);
 
+        validatePhasesConfig(context, descriptor);
```
IvanBorislavovDimitrov, line 70: extract this validation in a new class

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStep.java
```diff
@@ -59,13 +67,40 @@ protected StepPhase executeStep(ProcessContext context) {
...
+                  .stream()
+                  .flatMap(module -> module.getHooks().stream())
+                  .forEach(this::validateHookHasNoDuplicatePhaseConfigs);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void validateHookHasNoDuplicatePhaseConfigs(Hook hook) {
+        Object phasesConfigValue = hook.getParameters().get(SupportedParameters.PHASES_CONFIG);
+        if (!(phasesConfigValue instanceof List)) {
+            return;
```
IvanBorislavovDimitrov, line 92: shouldn't the deployment fail if phases-config is not a List?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/MergeDescriptorsStep.java
```diff
@@ -59,13 +67,40 @@ protected StepPhase executeStep(ProcessContext context) {
...
+        }
+        descriptor.getModules()
+                  .stream()
+                  .flatMap(module -> module.getHooks().stream())
+                  .forEach(this::validateHookHasNoDuplicatePhaseConfigs);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void validateHookHasNoDuplicatePhaseConfigs(Hook hook) {
+        Object phasesConfigValue = hook.getParameters().get(SupportedParameters.PHASES_CONFIG);
```
IvanBorislavovDimitrov, line 90: Null handling?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,150 @@
...
+public class ResolveHookTaskTargetAppStep extends SyncFlowableStep {
+
+    private static final String PHASE_KEY = "phase";
+    private static final String TARGET_APP_KEY = "target-app";
+    private static final String APPLICATION_PHASE_SUBSTRING = ".application.";
+
+    @Override
+    protected StepPhase executeStep(ProcessContext context) {
+        CloudApplicationExtended app = context.getVariable(Variables.APP_TO_PROCESS);
+        Hook hook = context.getVariable(Variables.HOOK_FOR_EXECUTION);
```
IvanBorislavovDimitrov, line 30: Instead of passing it as argument, just get it in resolveTargetAppName

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,150 @@
...
+        String hookExecutionPhase = context.getVariable(Variables.HOOK_EXECUTION_PHASE);
+        return hook.getPhases()
+                   .stream()
+                   .filter(p -> p.equals(hookExecutionPhase))
+                   .findFirst()
+                   .orElseGet(() -> hook.getPhases()
+                                        .stream()
+                                        .filter(p -> p.contains(APPLICATION_PHASE_SUBSTRING))
+                                        .findFirst()
+                                        .orElse(""));
```
IvanBorislavovDimitrov, line 84: Why are we searching for the first application hook?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    private String resolveAppNameForTarget(ProcessContext context, CloudApplicationExtended app, String targetApp) {
```
IvanBorislavovDimitrov, line 87: What will happen if we using hook:
blue-green.application.before-start.idle
with target-app: live 
During initial deployment when live app still does not exist?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+        return app.getName();
+    }
+
+    private String resolveAppNameWithLiveIdleSuffix(String appName, String targetApp) {
+        String liveSuffix = BlueGreenApplicationNameSuffix.LIVE.asSuffix();
+        String idleSuffix = BlueGreenApplicationNameSuffix.IDLE.asSuffix();
+
+        if (HookPhaseProcessType.HookProcessPhase.IDLE.getType().equals(targetApp)) {
+            if (appName.endsWith(liveSuffix)) {
```
IvanBorislavovDimitrov, line 109: When blue-green.application.before-start.idle is used. The apps are still called (for example):
cf-app-with-hooks-idle
and:
cf-app-with-hooks-live
If we remove the suffix we will get a name of application that does not exist.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,150 @@
...
+                return appName.substring(0, appName.length() - liveSuffix.length());
+            }
+            if (!appName.endsWith(idleSuffix)) {
+                return appName + idleSuffix;
+            }
+        }
+        if (HookPhaseProcessType.HookProcessPhase.LIVE.getType().equals(targetApp)) {
```
IvanBorislavovDimitrov, line 116: Can we simplify these conditions as they are hard to read?
We have only one hook that is executed before Phase.AFTER_RESUME of the blue-green (blue-green.application.before-start.idle).
When this hook is used apps are called:
cf-app-with-hooks-idle
cf-app-with-hooks-live
So we can just remove the suffix from appName and append idle or live. (more readable) BlueGreenApplicationNameSuffix::removeSuffix

--
All other hooks are executed after Phase.AFTER_RESUME. Then apps are called:
cf-app-with-hooks
cf-app-with-hooks-live
So if idle needed -> remove suffix, if live needed remove suffix append Live. (no conditions at all)

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private String swapColorSuffix(String appName, ApplicationColor fromColor, ApplicationColor toColor) {
+        String fromSuffix = fromColor.asSuffix();
+        String toSuffix = toColor.asSuffix();
+        if (appName.endsWith(fromSuffix)) {
+            return appName.substring(0, appName.length() - fromSuffix.length()) + toSuffix;
+        }
+        if (!appName.endsWith(toSuffix)) {
+            return appName + toSuffix;
+        }
```
IvanBorislavovDimitrov, line 135: Can you just remove suffix BlueGreenApplicationNameSuffix::removeSuffix then append toColor?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/DeploymentTypeDeterminer.java
```diff
@@ -20,4 +20,8 @@ public ProcessType determineDeploymentType(ProcessContext context) {
         return processTypeParser.getProcessType(context.getExecution());
     }
 
+    public ProcessType determineDeploymentTypeSafely(ProcessContext context) {
+        return processTypeParser.getProcessType(context.getExecution(), false);
+    }
```
IvanBorislavovDimitrov, line 25: What is this?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/HooksExecutor.java
```diff
@@ -50,6 +50,11 @@ private List<Hook> executeHooks(StepPhase currentStepPhase) {
                                                                                                       moduleToDeploy.getName());
         updateExecutedHooksForModule(alreadyExecutedHooksForModule, hooksWithPhases.getHookPhases(), hooksWithPhases.getHooks());
         context.setVariable(Variables.HOOKS_FOR_EXECUTION, hooksWithPhases.getHooks());
+        context.setVariable(Variables.HOOK_EXECUTION_PHASE, hooksWithPhases.getHookPhases()
+                                                                            .stream()
+                                                                            .findFirst()
+                                                                            .map(HookPhase::getValue)
+                                                                            .orElse(null));
```
IvanBorislavovDimitrov, line 57: This variable breaks the idea of having multiple phases for one step. While technically:
getHookPhasesBeforeStep
getHookPhasesAfterStep
always return only one hook phase, in future they can be many, so this variable should comply with one to many relationship (step to hook phases)

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
stiv03: With the current implementation we hit 404. What’s the right approach here- should the process fail or just log a warning? Also, should the combination of phase: blue-green.application.before-start.idle and target-app: live even be allowed during an initial deployment?

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
```diff
@@ -217,7 +217,10 @@ public class SupportedParameters {
     public static final Set<String> DEPENDENCY_PARAMETERS = Set.of(BINDING_NAME, ENV_VAR_NAME, VISIBILITY, USE_LIVE_ROUTES,
                                                                    SERVICE_BINDING_CONFIG, DELETE_SERVICE_KEY_AFTER_DEPLOYMENT);
 
-    public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES);
+    public static final String HOOK_TARGET_APP = "hook-target-app";
```
IvanBorislavovDimitrov: Is this parameter used?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
IvanBorislavovDimitrov: The execution of the hook should be skipped because no live application exists

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,133 @@
...
+        return List.of();
+    }
+
+    private String buildCurrentPhaseString(ProcessContext context, Hook hook) {
+        List<String> hookExecutionPhases = context.getVariable(Variables.HOOK_EXECUTION_PHASES);
+        return hook.getPhases()
+                   .stream()
+                   .filter(hookExecutionPhases::contains)
+                   .findFirst()
+                   .orElse("");
```
IvanBorislavovDimitrov: can we reuse HookPhase.NONE?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,133 @@
...
+
+    private String resolveAppNameForTarget(ProcessContext context, CloudApplicationExtended app, String targetApp) {
+        if (HookPhaseProcessType.HookProcessPhase.LIVE.getType()
+                                                      .equals(targetApp)) {
+            if (isInitialDeploy(context)) {
+                getStepLogger().warn(Messages.SKIPPING_HOOK_TASK_NO_LIVE_APP, app.getName());
+                return null;
+            }
+            ApplicationColor liveColor = context.getVariable(Variables.LIVE_MTA_COLOR);
+            String suffix = liveColor != null ? liveColor.asSuffix() : BlueGreenApplicationNameSuffix.LIVE.asSuffix();
```
IvanBorislavovDimitrov: Can you extract in smaller method?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,133 @@
...
+                                                      .equals(targetApp)) {
+            String baseName = BlueGreenApplicationNameSuffix.removeSuffix(app.getName());
+            ApplicationColor idleColor = context.getVariable(Variables.IDLE_MTA_COLOR);
+            if (idleColor != null) {
+                return baseName + idleColor.asSuffix();
+            }
+            if (isAfterRenamePhase(context)) {
+                return baseName;
+            }
+            return baseName + BlueGreenApplicationNameSuffix.IDLE.asSuffix();
```
IvanBorislavovDimitrov: Can you extract in smaller method?

multiapps-controller-process/src/main/resources/org/cloudfoundry/multiapps/controller/process/backup-existing-app.bpmn
```diff
@@ -24,6 +24,12 @@
       <flowable:in source="isDisposableUserProvidedServiceEnabled" target="isDisposableUserProvidedServiceEnabled"></flowable:in>
       <flowable:in source="disposableUserProvidedServiceName" target="disposableUserProvidedServiceName"></flowable:in>
       <flowable:in source="mtaId" target="mtaId"></flowable:in>
+      <flowable:in source="subprocessPhase" target="subprocessPhase"></flowable:in>
+      <flowable:in source="phase" target="phase"></flowable:in>
+      <flowable:in source="hookExecutionPhases" target="hookExecutionPhases"></flowable:in>
+        <flowable:in source="deployedMta" target="deployedMta"></flowable:in>
+        <flowable:in source="idleMtaColor" target="idleMtaColor"></flowable:in>
+      <flowable:in source="liveMtaColor" target="liveMtaColor"></flowable:in>
```
IvanBorislavovDimitrov, line 32: could you double check the diagrams so they are exported from the modeler because there is some space indentation here

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStepTest.java
```diff
@@ -0,0 +1,276 @@
...
+
+        step.execute(execution);
+
+        assertStepFinishedSuccessfully();
+        assertEquals(APP_NAME_LIVE, context.getVariable(Variables.APP_TO_PROCESS)
+                                           .getName());
+    }
+
+    @Test
+    void bgDeploy_targetIdle_resolvesToIdleColorApp() {
```
vkalapov: Stick to one naming convention for the test method names

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/Messages.java
```diff
@@ -213,6 +215,7 @@ public class Messages {
     public static final String CANNOT_RETRIEVE_PARAMETERS_OF_BINDING_BETWEEN_APPLICATION_0_AND_SERVICE_INSTANCE_1 = "Cannot retrieve parameters of binding between application \"{0}\" and service instance \"{1}\"";
     public static final String CANNOT_RETRIEVE_PARAMETERS_OF_BINDING_BETWEEN_APPLICATION_0_AND_SERVICE_INSTANCE_1_FIX = "Cannot retrieve parameters of binding between application \"{0}\" and service instance \"{1}\". Got 502.";
     public static final String CANNOT_RETRIEVE_INSTANCE_OF_SERVICE = "Cannot retrieve service instance of service \"{0}\"";
+    public static final String HOOK_PHASE_BLUE_GREEN_BEFORE_UNMAP_ROUTES_IDLE_USED = "Hook \"{0}\" uses deprecated phase \"blue-green.application.before-unmap-routes.idle\" which is unreachable and will be removed in a future version";
```
vkalapov: This log message flows a bit better to me:
`Hook \"{0}\" uses the deprecated phase \"blue-green.application.before-unmap-routes.idle\". This phase is unreachable and will be removed in a future version.`

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/model/SupportedParameters.java
```diff
@@ -217,7 +217,9 @@ public class SupportedParameters {
     public static final Set<String> DEPENDENCY_PARAMETERS = Set.of(BINDING_NAME, ENV_VAR_NAME, VISIBILITY, USE_LIVE_ROUTES,
                                                                    SERVICE_BINDING_CONFIG, DELETE_SERVICE_KEY_AFTER_DEPLOYMENT);
 
-    public static final Set<String> MODULE_HOOK_PARAMETERS = Set.of(NAME, COMMAND, MEMORY, DISK_QUOTA, HOOK_REQUIRES);
+    public static final String PHASES_CONFIG = "phases-config";
```
vkalapov: Move PHASES_CONFIG up with the other public static final declarations.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/HookPhasesConfigValidator.java
```diff
@@ -0,0 +1,62 @@
...
+    @SuppressWarnings("unchecked")
+    private void validateHookHasNoDuplicatePhaseConfigs(Hook hook) {
+        Object phasesConfigValue = hook.getParameters().get(SupportedParameters.PHASES_CONFIG);
+        if (phasesConfigValue == null) {
+            return;
+        }
+        if (!(phasesConfigValue instanceof List)) {
+            throw new SLException(MessageFormat.format(Messages.INVALID_PHASES_CONFIG_NOT_A_LIST, hook.getName()));
+        }
+        List<Map<String, String>> phasesConfig = (List<Map<String, String>>) phasesConfigValue;
```
vkalapov: Can we scope @SuppressWarnings("unchecked") to just the cast instead of the whole method?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @SuppressWarnings("unchecked")
+    private List<Map<String, String>> getPhasesConfig(Hook hook) {
+        Object phasesConfigValue = hook.getParameters()
+                                       .get(SupportedParameters.PHASES_CONFIG);
+        if (phasesConfigValue instanceof List) {
+            return (List<Map<String, String>>) phasesConfigValue;
+        }
+        return List.of();
+    }
```
vkalapov, line 74: We treat missing or null PHASES_CONFIG as empty, but silently ignore cases where it’s present with the wrong type. Would it make sense to log or validate that here?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,145 @@
...
+                                       .map(config -> config.get(TARGET_APP_KEY))
+                                       .findFirst()
+                                       .orElse(null);
+
+        if (targetApp == null) {
+            return app.getName();
+        }
+
+        return resolveAppNameForTarget(context, app, targetApp);
+    }
```
vkalapov, line 60: Small refactoring suggestion: `app.getName()` is returned in multiple branches here. Consider assigning it as the default result and only changing it when a target app is found, to reduce repetition.

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
```diff
@@ -0,0 +1,145 @@
...
+        if (resolvedAppName == null) {
+            context.setVariable(Variables.TASKS_TO_EXECUTE, List.of());
+            return StepPhase.DONE;
+        }
+        if (!resolvedAppName.equals(app.getName())) {
+            context.setVariable(Variables.APP_TO_PROCESS, buildAppWithName(app, resolvedAppName));
+        }
+
+        return StepPhase.DONE;
+    }
```
vkalapov, line 39: Here all branches return StepPhase.DONE. Would it be clearer to return once at the end and update the context conditionally (e.g., using an else if)?

vkalapov (changes requested):

IvanBorislavovDimitrov (dismissed):

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/ResolveHookTaskTargetAppStep.java
stiv03: The suggested refactoring would work if resolveAppNameForTarget always returns a non-null value, but resolveLiveAppName returns null as a sentinel to signal that the hook execution should be skipped (initial deploy, no live app exists). Using Optional.map() with a lambda that can return null throws a NPE.

vkalapov (approved):

IvanBorislavovDimitrov (approved):

PR #1820, opened 2026-04-22
Handle multiple bindings between app and service

theghost5800:
JIRA:LMCROSSITXSADEPLOY-3409


multiapps-controller-client/src/main/java/org/cloudfoundry/multiapps/controller/client/facade/rest/CloudControllerRestClientImpl.java
```diff
@@ -2035,23 +2037,51 @@ private void doDeleteServiceInstance(UUID serviceInstanceGuid) {
...
+        throwOnErrors(errors);
+        return jobIds;
+    }
+
+    private void handleServiceBindingDeletion(UUID guid, List<String> jobIds, List<RuntimeException> errors) {
+        try {
+            Optional.ofNullable(deleteSingleServiceBinding(guid))
+                    .ifPresent(jobIds::add);
+        } catch (RuntimeException e) {
+            errors.add(e);
```
IvanBorislavovDimitrov: Why not fast-failing immediately, but collecting the errors? If there is some regression with a service broker, won't this produce too much noise?

multiapps-controller-client/src/main/java/org/cloudfoundry/multiapps/controller/client/facade/rest/CloudControllerRestClientImpl.java
```diff
@@ -2035,23 +2037,51 @@ private void doDeleteServiceInstance(UUID serviceInstanceGuid) {
...
+        }
+        throwOnErrors(errors);
+        return jobIds;
+    }
+
+    private void handleServiceBindingDeletion(UUID guid, List<String> jobIds, List<RuntimeException> errors) {
+        try {
+            Optional.ofNullable(deleteSingleServiceBinding(guid))
+                    .ifPresent(jobIds::add);
+        } catch (RuntimeException e) {
```
IvanBorislavovDimitrov: Why catching so generic exception here? Can't we use AbstractCloudFoundryException?

multiapps-controller-client/src/main/java/org/cloudfoundry/multiapps/controller/client/facade/rest/CloudControllerRestClientImpl.java
```diff
@@ -2035,23 +2037,51 @@ private void doDeleteServiceInstance(UUID serviceInstanceGuid) {
...
-    private Optional<String> doDeleteServiceBinding(UUID guid) {
-        String jobId = delegate.serviceBindingsV3()
-                               .delete(DeleteServiceBindingRequest.builder()
-                                                                  .serviceBindingId(guid.toString())
-                                                                  .build())
-                               .block();
-        return Optional.ofNullable(jobId);
+    private List<String> doDeleteServiceBindings(List<UUID> guids) {
+        List<String> jobIds = new ArrayList<>();
+        List<RuntimeException> errors = new ArrayList<>();
```
IvanBorislavovDimitrov: These two parameters are in-parameters which is not a best practice. Can you split the method to just return a List of some results instead?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/PollServiceBindingLastOperationExecution.java
```diff
@@ -18,57 +19,70 @@ public class PollServiceBindingLastOperationExecution implements AsyncExecution
...
         UUID applicationGuid = controllerClient.getApplicationGuid(app.getName());
         UUID serviceInstanceGuid = controllerClient.getRequiredServiceInstanceGuid(serviceInstanceName);
-        return controllerClient.getServiceBindingForApplication(applicationGuid, serviceInstanceGuid);
+        return controllerClient.getServiceBindingsForApplication(applicationGuid, serviceInstanceGuid);
     }
 
-    protected AsyncExecutionState checkServiceBindingOperationState(CloudServiceBinding serviceBinding, ProcessContext context) {
-        ServiceCredentialBindingOperation lastOperation = serviceBinding.getServiceBindingOperation();
+    protected AsyncExecutionState checkServiceBindingOperationState(List<CloudServiceBinding> serviceBindings, ProcessContext context) {
+        CloudServiceBinding failedServiceBinding = null;
```
IvanBorislavovDimitrov: Why only the last failed binding is tracked and not all? Is it always the latest?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/PollServiceBindingLastOperationExecution.java
IvanBorislavovDimitrov: I think we should rename these type of classes so they reflect the class is used to monitor multiple bindings between app and service

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/PollServiceUnbindingsOperationExecution.java
```diff
@@ -1,9 +1,32 @@
...
+
 import org.cloudfoundry.multiapps.controller.process.variables.Variables;
 
 public class PollServiceUnbindingOperationExecution extends PollServiceBindingUnbindingOperationBaseExecution {
 
+    @Override
+    public AsyncExecutionState execute(ProcessContext context) {
+        List<String> jobIds = context.getVariable(Variables.SERVICE_UNBINDING_JOB_IDS);
+        if (jobIds.isEmpty()) {
+            return super.execute(context);
```
IvanBorislavovDimitrov, line 14: Won't this produce a null pointer exception because
            context.setVariable(Variables.SERVICE_UNBINDING_JOB_ID, jobId);
is not set if it is a first execution of this step?

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/util/ServiceBindingParametersGetter.java
```diff
@@ -158,4 +163,13 @@ public Map<String, Object> getServiceBindingParametersFromExistingInstance(Cloud
         }
     }
 
+    private CloudServiceBinding selectNewestBinding(List<CloudServiceBinding> bindings) {
+        return bindings.stream()
+                       .filter(binding -> binding.getServiceBindingOperation()
+                                                 .getState() == ServiceCredentialBindingOperation.State.SUCCEEDED)
+                       .max(Comparator.comparing(binding -> binding.getMetadata()
+                                                                   .getCreatedAt()))
+                       .orElseGet(() -> bindings.isEmpty() ? null : bindings.get(0));
```
IvanBorislavovDimitrov: I guess we will never get here, but isn't it more correctly at least to return the latest created binding instead of bindings.get(0)? Or just throw an exception?

multiapps-controller-client/src/main/java/org/cloudfoundry/multiapps/controller/client/facade/rest/CloudControllerRestClientImpl.java
theghost5800: The reason for this change is due to common from claude review that if we use this codeblock
```
return Flux.fromIterable(guids)
             .map(UUID::toString)
             .flatMap(guid -> delegate.serviceBindingsV3().delete(...))
             .collectList()
             .block();
 ```
 
 it will try to run all deletion requests concurrently and it's possible to lose track of other delete request responses. It's interesting which approach is better.

IvanBorislavovDimitrov (approved):

Yavor16 (approved):

PR #1821, opened 2026-04-22
add claude config

s-yonkov-yonkov:


pom.xml
```diff
@@ -249,6 +249,26 @@
                     <artifactId>jacoco-maven-plugin</artifactId>
                     <version>0.8.14</version>
                 </plugin>
+                <plugin>
+                    <groupId>com.diffplug.spotless</groupId>
+                    <artifactId>spotless-maven-plugin</artifactId>
+                    <version>3.4.0</version>
+                    <configuration>
+                        <!-- THE MAGIC LINE: Spotless will only format files changed relative to this branch -->
```
IvanBorislavovDimitrov: same like in multiapps?

s-yonkov-yonkov: It won't break formatting, it just limits which files get formatted. With this setup when spotless:apply is executed it will be triggered only for the files that are changed compared to master and the legacy files that were never formatted or updated will remain untouched.

IvanBorislavovDimitrov (dismissed):

IvanBorislavovDimitrov (approved):

ikasarov (approved):

PR #1864, opened 2026-06-29
Add LogSanitizer for tainted values in log statements

stiv03:
LMCROSSITXSADEPLOY-3479


theghost5800 (approved):

theghost5800:
Why sanitize is not used for filter classes?

stiv03:
> theghost5800

The filter classes log - request.getRequestURI(), and Tomcat's HTTP parser rejects request lines containing \r, \n or \t with 400 Bad Request before the filter chain runs. So the URI in a filter can never contain those control characters, so they are false positive

Yavor16 (approved):

PR #1879, opened 2026-07-16
Add cf-metadata parameter for user-defined CF labels and annotations

Yavor16:
## What

Adds support for a new `cf-metadata` MTA parameter that lets descriptor authors
attach user-defined Cloud Foundry V3 labels and annotations to applications and
managed service instances directly from `mtad.yaml`.

```yaml
modules:
  - name: my-app
    type: org.cloudfoundry.application
    parameters:
      cf-metadata:
        labels:
          commit-id: "abc123"
        annotations:
          description: "Built from the current descriptor"
resources:
  - name: my-service
    type: org.cloudfoundry.managed-service
    parameters:
      cf-metadata:
        labels:
          env: "prod"
```

## Why

Previously, user-defined CF metadata (labels and annotations) could only be set
manually after deployment, which was error-prone and not reproducible from the
MTA descriptor. This change makes that metadata declarative and part of the
deployment lifecycle.

## How

- New `cf-metadata` module/resource parameter registered in `SupportedParameters`.
- `CfMetadataParser` parses the `labels` and `annotations` sub-maps into a new
  `CfUserMetadata` value object.
- `CfMetadataValidator` enforces CF metadata key/value constraints and rejects
  reserved `mta.cloudfoundry.org/` keys with a clear error.
- User metadata is threaded through `CloudApplicationExtended` and
  `CloudServiceInstanceExtended`, and merged into the built `Metadata` by
  `ApplicationMetadataBuilder` and `ServiceMetadataBuilder`.
- Change detection in `CreateOrUpdateAppStep` and
  `DetermineServiceCreateUpdateServiceActionsStep` compares only user-defined
  keys, so updates to MTA-internal metadata no longer trigger spurious CF
  metadata API calls, and user metadata changes correctly trigger an update.

## Behaviour notes

- MTA-internal `mta.cloudfoundry.org/` labels and annotations are always
  preserved; user descriptors cannot overwrite or delete them.
- User-declared label/annotation values support standard descriptor variable
  interpolation.
- MTA archives without `cf-metadata` deploy identically to prior behaviour.

## Tests

- `CfMetadataParserTest`, `CfMetadataValidatorTest`
- `ApplicationMetadataBuilderTest`, `ServiceMetadataBuilderTest`,
  `ApplicationCloudModelBuilderCfMetadataTest`
- `CreateOrUpdateStepWithExistingAppTest`,
  `DetermineServiceCreateUpdateServiceActionsStepTest`

JIRA:LMCROSSITXSADEPLOY-1550

### DIFF
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java
@@ -1,0 +1,76 @@
+package org.cloudfoundry.multiapps.controller.core.validators.parameters;
+
+import java.text.MessageFormat;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.cloudfoundry.multiapps.common.ContentException;
+import org.cloudfoundry.multiapps.controller.client.lib.domain.CfUserMetadata;
+import org.cloudfoundry.multiapps.controller.core.Messages;
+import org.cloudfoundry.multiapps.controller.core.cf.metadata.MtaMetadataAnnotations;
+import org.cloudfoundry.multiapps.controller.core.cf.metadata.MtaMetadataLabels;
+
+public class CfMetadataValidator {
+
+    private static final Pattern KEY_PATTERN = Pattern.compile(
+        "([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,251}[a-zA-Z0-9]/)?[a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,61}[a-zA-Z0-9]|[a-zA-Z0-9]");
+
+    private static final Set<String> RESERVED_KEY_NAMES = Set.of(
+        MtaMetadataLabels.MTA_ID,
+        MtaMetadataLabels.MTA_NAMESPACE,
+        MtaMetadataLabels.SPACE_GUID,
+        MtaMetadataAnnotations.MTA_ID,
+        MtaMetadataAnnotations.MTA_VERSION,
+        MtaMetadataAnnotations.MTA_NAMESPACE,
+        MtaMetadataAnnotations.MTA_MODULE,
+        MtaMetadataAnnotations.MTA_MODULE_PUBLIC_PROVIDED_DEPENDENCIES,
+        MtaMetadataAnnotations.MTA_MODULE_BOUND_SERVICES,
+        MtaMetadataAnnotations.MTA_RESOURCE);
+
+    private static final String MTA_KEY_PREFIX = "mta_";
+    private static final int MAX_LABEL_VALUE_LENGTH = 63;
+    private static final int MAX_ANNOTATION_VALUE_LENGTH = 5000;
+
+    public void validate(CfUserMetadata metadata) {
+        metadata.getLabels()
+                .forEach((key, value) -> validateLabelEntry(key, value));
+        metadata.getAnnotations()
+                .forEach((key, value) -> validateAnnotationEntry(key, value));
+    }
+
+    private void validateLabelEntry(String key, String value) {
+        validateKeyFormat(key);
+        validateKeyNotReserved(key);
+        if (value != null && value.length() > MAX_LABEL_VALUE_LENGTH) {
+            throw new ContentException(MessageFormat.format(Messages.CF_METADATA_LABEL_VALUE_TOO_LONG_0, key));
+        }
+    }
+
+    private void validateAnnotationEntry(String key, String value) {
+        validateKeyFormat(key);
+        validateKeyNotReserved(key);
+        if (value != null && value.length() > MAX_ANNOTATION_VALUE_LENGTH) {
+            throw new ContentException(MessageFormat.format(Messages.CF_METADATA_ANNOTATION_VALUE_TOO_LONG_0, key));
+        }
+    }
+
+    private void validateKeyFormat(String key) {
+        if (!KEY_PATTERN.matcher(key)
+                        .matches()) {
+            throw new ContentException(MessageFormat.format(Messages.CF_METADATA_INVALID_LABEL_KEY_0, key));
+        }
+    }
+
+    private void validateKeyNotReserved(String key) {
+        String nameSegment = extractNameSegment(key);
+        if (nameSegment.startsWith(MTA_KEY_PREFIX) || RESERVED_KEY_NAMES.contains(nameSegment) || RESERVED_KEY_NAMES.contains(key)) {
+            throw new ContentException(MessageFormat.format(Messages.CF_METADATA_RESERVED_LABEL_KEY_0, key));
+        }
+    }
+
+    private String extractNameSegment(String key) {
+        int slashIndex = key.indexOf('/');
+        return slashIndex >= 0 ? key.substring(slashIndex + 1) : key;
+    }
+
+}
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java
@@ -16,7 +16,7 @@ import org.cloudfoundry.multiapps.controller.core.cf.metadata.MtaMetadataLabels;
 public class CfMetadataValidator {
 
     private static final Pattern KEY_PATTERN = Pattern.compile(
-        "([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,251}[a-zA-Z0-9]/)?[a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,61}[a-zA-Z0-9]|[a-zA-Z0-9]");
+        "([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,251}[a-zA-Z0-9]/)?([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,61}[a-zA-Z0-9]|[a-zA-Z0-9])");
 
     private static final Set<String> RESERVED_KEY_NAMES = Collections.unmodifiableSet(
         new HashSet<>(Arrays.asList(MtaMetadataLabels.MTA_ID,
--- a/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java
+++ b/multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java
@@ -1,6 +1,9 @@
 package org.cloudfoundry.multiapps.controller.core.validators.parameters;
 
 import java.text.MessageFormat;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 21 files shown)


multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.cloudfoundry.multiapps.common.ContentException;
+import org.cloudfoundry.multiapps.controller.client.lib.domain.CfUserMetadata;
+import org.cloudfoundry.multiapps.controller.core.Messages;
+import org.cloudfoundry.multiapps.controller.core.cf.metadata.MtaMetadataAnnotations;
+import org.cloudfoundry.multiapps.controller.core.cf.metadata.MtaMetadataLabels;
+
+public class CfMetadataValidator {
+
+    private static final Pattern KEY_PATTERN = Pattern.compile(
```
Yavor16, line 18: **Severity:** IMPORTANT · **Confidence:** medium

`KEY_PATTERN` does not match single-character name-segments in prefixed keys. The prefix branch requires `[a-zA-Z0-9][a-zA-Z0-9\-_.]{0,61}[a-zA-Z0-9]` as the name segment after `/`, which requires ≥ 2 chars. A key like `foo.com/x` (valid per CF) fails validation because the single-char alternative `|[a-zA-Z0-9]` only applies to the whole unprefixed key, not to the name segment following a prefix.

**Evidence:** `private static final Pattern KEY_PATTERN = Pattern.compile(`

**Fix:**
```suggestion
    private static final Pattern KEY_PATTERN = Pattern.compile(
        "([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,251}[a-zA-Z0-9]/)?([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,61}[a-zA-Z0-9]|[a-zA-Z0-9])");
```

Yavor16 (commented):
Automated code review by MTA Quality Agent — 1 finding.

Yavor16:
## MTA Quality Report — cloudfoundry/multiapps-controller PR #1879

**Jira:** LMCROSSITXSADEPLOY-1550 — (backlog item could not be fetched in this environment)
**Backlog alignment:** WARN
- _Implements Jira scope?_ **unclear** — The Jira MCP tool was unavailable, so the backlog summary/description could not be retrieved to compare against the diff.
- _Changes outside Jira scope?_ **unclear** — Same reason; scope boundaries could not be verified. The PR description is internally consistent (adds `cf-metadata` parsing, validation, model threading, metadata building, change detection).

---

### Code Review

**6** finding(s): 1 posted inline · 5 general (see below).
_See the inline review on the PR diff for posted findings._

---

### Security

- **HIGH** — `CfMetadataValidator.java:~21-25` (`validateKeyNotReserved` / `RESERVED_KEY_NAMES`): Reserved-key allowlist bypass allowing tampering with the platform autoscaler control label (metadata spoofing / functionality tampering, CWE-284 Improper Access Control). The builders write internal MTA metadata first, then apply user-supplied labels/annotations, so a colliding user key overwrites the internal value. `RESERVED_KEY_NAMES` does **not** include `MtaMetadataLabels.AUTOSCALER_LABEL` (`app-autoscaler.cloudfoundry.org/disable-autoscaling`); `validateKeyNotReserved` extracts name segment `disable-autoscaling`, which neither starts with `mta_` nor is in the reserved set, so it passes validation. Change detection also treats `AUTOSCALER_LABEL` as internal and strips it before comparing, so a user-controlled value would not even be reconciled.
  **Fix:** Add `MtaMetadataLabels.AUTOSCALER_LABEL` to `RESERVED_KEY_NAMES`. Consider applying user metadata **before** internal MTA metadata so internal keys always win (defense in depth), and consolidate the three independent key lists into shared constants to prevent future drift.
- **LOW** — Three independently maintained lists of internal-key names (validator `RESERVED_KEY_NAMES` + two step classes) — maintainability risk that will produce more bypasses as new internal labels are added.

---

### SonarCloud

| Check | Status | Result | Annotations | Details |
|---|---|---|---|---|
| Build and analyze | in_progress | ⏳ RUNNING | — | [link](https://github.com/cloudfoundry/multiapps-controller/actions/runs/29492172731/job/87600662312) |

**Quality gate:** UNKNOWN

The `Build and analyze` job (which runs the SonarCloud analysis) is still `in_progress`, so no quality gate conclusion or annotations are available yet. Re-run this check once the `Build and analyze` job completes.

---

### Dependency CVEs

✅ No new CVEs — this PR does not modify any dependency files (`pom.xml`, `build.gradle`, lockfiles).

---

### General findings

- **Severity:** CRITICAL · **Confidence:** low · **File:** `multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java`
  The code-reviewer reported a compile-breaking typo — `Messages.CF_METADATA_INVALID_label_KEY_0` (mixed-case) instead of the declared `CF_METADATA_INVALID_LABEL_KEY_0`. **This could not be confirmed against the diff:** RIGHT-side line 63 of the file reads `throw new ContentException(MessageFormat.format(Messages.CF_METADATA_INVALID_LABEL_KEY_0, key));` (correct all-caps), and `Messages.java` declares `CF_METADATA_INVALID_LABEL_KEY_0`. Demoted to a general note and not posted inline because the claimed evidence is contradicted by the actual source. Please double-check locally; if the constant reference is correct as shown, no action is needed.

- **Severity:** IMPORTANT · **File:** `ServicesCloudModelBuilder.java`
  `cf-metadata` on a user-provided or existing service is silently ignored. `createUserProvidedService` and `createExistingService` hard-code `null` for `userCfMetadata` without parsing the parameter, so `cf-metadata` on those resource types is dropped rather than rejected.
  **Fix:** Either parse and pass through `cf-metadata` for those service types, or add explicit validation that throws a `ContentException` when `cf-metadata` is present on unsupported resource types.

- **Severity:** IMPORTANT · **File:** `CfMetadataParser.java` + `CloudApplicationExtended.java`
  Inconsistent null-vs-empty contract on `getUserCfMetadata()` / `CfMetadataParser`. The parser returns an empty `ImmutableCfUserMetadata` when `cf-metadata` is absent, but `getUserCfMetadata()` is `@Nullable`. Callers that null-check (e.g. `userCfMetadata != null` in `ApplicationMetadataBuilder`) will never skip the branch even for modules that never declared `cf-metadata`, making the `@Nullable` annotation misleading.
  **Fix:** Pick one contract — either return `null` from the parser when absent (keeping `@Nullable`), or drop `@Nullable` and always pass an empty object.

- **Severity:** IMPORTANT · **File:** `CreateOrUpdateAppStep.java` + `DetermineServiceCreateUpdateServiceActionsStep.java`
  Duplicated internal-key sets across two step classes plus the validator. Both steps inline identical `Set<String>` literals for `internalLabelKeys` / `internalAnnotationKeys`, re-created on every call, and the same concept lives in `CfMetadataValidator.RESERVED_KEY_NAMES`. Future additions must be updated in three places (this is also the root cause of the HIGH security finding above).
  **Fix:** Extract as shared `private static final` constants (e.g. in `MtaMetadataLabels`/`MtaMetadataAnnotations` or a new `MtaInternalMetadataKeys` class).

- **Severity:** STYLE · **File:** `ServicesCloudModelBuilder.java`
  Import ordering violation (Spotless). The `CfMetadataParser` import (`org.cloudfoundry.multiapps.controller.*`) is placed after `org.cloudfoundry.multiapps.mta.*`, violating alphabetical ordering enforced by Spotless; CI will catch this.
  **Fix:** Move the `CfMetadataParser` import above the `org.cloudfoundry.multiapps.mta.*` imports.

multiapps-controller-core/src/main/java/org/cloudfoundry/multiapps/controller/core/validators/parameters/CfMetadataValidator.java
```diff
@@ -0,0 +1,79 @@
...
+        "([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,251}[a-zA-Z0-9]/)?([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,61}[a-zA-Z0-9]|[a-zA-Z0-9])");
```
Yavor16, line 19: **Severity:** IMPORTANT · **Confidence:** 85

The prefix sub-pattern `[a-zA-Z0-9][a-zA-Z0-9\-_.]{0,251}[a-zA-Z0-9]` requires a minimum of two characters, so a single-character DNS prefix such as `a/mykey` is incorrectly rejected as invalid. RFC 1123 allows single-character DNS labels, and the error message does not document a two-character minimum for the prefix, so users will receive a misleading "invalid key" error for valid keys.

**Evidence:** `"([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,251}[a-zA-Z0-9]/)?([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,61}[a-zA-Z0-9]|[a-zA-Z0-9])");`

**Fix:**
```suggestion
        "([a-zA-Z0-9]([a-zA-Z0-9\\-_.]{0,251}[a-zA-Z0-9])?/)?([a-zA-Z0-9][a-zA-Z0-9\\-_.]{0,61}[a-zA-Z0-9]|[a-zA-Z0-9])");
```

multiapps-controller-process/src/main/java/org/cloudfoundry/multiapps/controller/process/steps/CreateOrUpdateAppStep.java
```diff
@@ -250,11 +254,41 @@ private void updateMetadata() {
             if (app.getV3Metadata() == null) {
                 return;
             }
-            if (!Objects.equals(existingApp.getV3Metadata(), app.getV3Metadata())) {
+            if (hasUserMetadataChanged(existingApp.getV3Metadata(), app.getV3Metadata())) {
```
Yavor16, line 257: **Severity:** IMPORTANT · **Confidence:** 82

The sets `internalLabelKeys` and `internalAnnotationKeys` are defined as inline literals inside `hasUserMetadataChanged`, duplicating the identical definitions in `DetermineServiceCreateUpdateServiceActionsStep.shouldUpdateMetadata`. Any future addition of a new internal metadata key (or removal of an obsolete one) must be applied in both places; missing one silently changes the change-detection semantics for either apps or services. Both sets should be extracted to shared constants.

**Evidence:** `Set<String> internalLabelKeys = Set.of(MtaMetadataLabels.MTA_ID, MtaMetadataLabels.MTA_NAMESPACE, MtaMetadataLabels.SPACE_GUID,`

**Fix:** Extract both `internalLabelKeys` and `internalAnnotationKeys` as `private static final Set<String>` constants in a shared location (e.g. `MtaMetadataLabels`/`MtaMetadataAnnotations` or a new `MtaInternalMetadataKeys` class), and reuse them here, in `DetermineServiceCreateUpdateServiceActionsStep`, and in `CfMetadataValidator.RESERVED_KEY_NAMES`.

Yavor16 (commented):
Automated code review by MTA Quality Agent — 2 finding(s).

Yavor16:
## MTA Quality Report — cloudfoundry/multiapps-controller PR #1879

**Jira:** LMCROSSITXSADEPLOY-1550 — (backlog item could not be fetched in this environment; the `sap-jira` MCP tool was unavailable)
**Backlog alignment:** WARN
- _Implements Jira scope?_ **unclear** — The Jira MCP tool was unavailable, so the backlog summary/description could not be retrieved to compare against the diff. The PR description is internally consistent (adds `cf-metadata` parsing, validation, model threading, metadata building, and change detection).
- _Changes outside Jira scope?_ **unclear** — Same reason; scope boundaries could not be verified against the backlog item.

---

### Code Review

**2** finding(s): 2 posted inline.
_See the inline review on the PR diff for posted findings._

---

### Security

- **HIGH** — `CfMetadataValidator.java` (`RESERVED_KEY_NAMES`): Authorization / security-control bypass (CWE-284: Improper Access Control). User-controlled cf-metadata labels are validated only against `RESERVED_KEY_NAMES`, which omits `MtaMetadataLabels.AUTOSCALER_LABEL` (`app-autoscaler.cloudfoundry.org/disable-autoscaling`). A user can therefore set this label via the `cf-metadata` parameter; the value flows unfiltered through `CfMetadataParser.parse()` into `ApplicationMetadataBuilder.build()` (`userCfMetadata.getLabels().forEach(builder::label)`) and is applied to the CF app. The `app-autoscaler` component reads exactly this label to suspend autoscaling, so an untrusted MTA descriptor can disable autoscaling on a deployed app — a controller-managed operational control the user is not meant to touch. The team's own code confirms the label is controller-internal: `CreateOrUpdateAppStep.hasUserMetadataChanged` lists `AUTOSCALER_LABEL` in `internalLabelKeys` to exclude it from user-change detection. The two sets are inconsistent (internal for change-detection, but not reserved for validation), which makes the gap concrete.
  **Fix:** Add `MtaMetadataLabels.AUTOSCALER_LABEL` to `RESERVED_KEY_NAMES`. Best practice: derive both the validator's reserved set and `CreateOrUpdateAppStep`'s `internalLabelKeys`/`internalAnnotationKeys` from a single shared constant collection so the lists cannot drift out of sync again.
- **LOW** — `CfMetadataParser.java` (`extractStringMap`, unchecked cast): After per-entry type checks, the whole map is cast `(Map<String, String>) rawMap`. Functionally guarded by the preceding entry validation, so not exploitable as introduced by this PR. No action required for security.
- **LOW** — `CfMetadataValidator.java` (reserved-key check on name segment / exact key): `validateKeyNotReserved` checks the name segment, the full key, and the `mta_` prefix. This correctly blocks prefixed variants; once `AUTOSCALER_LABEL` is added to the set, `RESERVED_KEY_NAMES.contains(key)` will match its full domain-prefixed form. No separate fix needed beyond the HIGH finding.

---

### SonarCloud

| Check | Status | Result | Annotations | Details |
|---|---|---|---|---|
| Build and analyze | completed | ❌ FAILED | — | [link](https://github.com/cloudfoundry/multiapps-controller/actions/runs/29494484839/job/87608099214) |
| build | completed | ❌ FAILED | — | [link](https://github.com/cloudfoundry/multiapps-controller/actions/runs/29494484770/job/87608098963) |

**Quality gate:** UNKNOWN

The SonarCloud-associated job ("Build and analyze") failed at the build/workflow level before producing a quality-gate result. No SonarCloud quality-gate conclusion or per-line issue annotations were emitted.

#### Findings

No SonarCloud per-line code smells, bugs, or vulnerabilities were surfaced as check-run annotations.

---

### Dependency CVEs

⏭️ CVE scan skipped on this review-fix re-run — dependencies were scanned on the initial quality run.

PR #1884, opened 2026-07-28
Migrate DigestHelperTest from MD5 to SHA-256 (FIPS 140-3 )

vkalapov:
## What

Change `Constants.DIGEST_ALGORITHM` from `"MD5"` to `"SHA-384"` and update all test fixtures accordingly.

## Why

SHA-384 satisfies both FIPS 140-3 (FIPS 180-4) and NCS MODERATE requirements. MD5 and SHA-256 are not approved at NCS MODERATE level.

## Compatibility

**Backward compatible — no breaking change.**

The server always computes and stores its own digest on upload (`DatabaseFileService` → `SqlFileQueryProvider`). No client-sent digest is ever read or verified on the server side.

- **Old CLI users**: `isFileAlreadyUploaded` will hit `"Unsupported digest algorithm"` for the SHA-384 entries returned by `GET /files`, re-upload on every retry instead of skipping — deploys complete successfully.
- **First deploy after DS restart**: `UploadAppStep` detects mismatch between freshly computed SHA-384 and the MD5 value stored in the CF app environment → one-time forced re-upload of every app binary. Self-heals on the second deploy.

LMCROSSITXSADEPLOY-3224


multiapps-controller-persistence/src/test/java/org/cloudfoundry/multiapps/controller/persistence/services/DatabaseFileServiceTest.java
```diff
@@ -54,8 +54,8 @@ class DatabaseFileServiceTest {
     private static final String SELECT_FILE_WITH_CONTENT = "SELECT FILE_ID FROM {0} WHERE FILE_ID=? AND CONTENT IS NOT NULL";
     private static final String UPDATE_MODIFICATION_TIME = "UPDATE {0} SET MODIFIED=? WHERE FILE_ID=?";
     private static final String LIQUIBASE_CHANGELOG_LOCATION = "org/cloudfoundry/multiapps/controller/persistence/db/changelog/db-changelog.xml";
-    private static final String DIGEST_METHOD = "MD5";
-    private static final String PIC_MD5_DIGEST = "b39a167875c3771c384c9aa5601fc2d6";
+    private static final String DIGEST_METHOD = "SHA-384";
+    private static final String PIC_SHA256_DIGEST = "b43c60daa2a0f2e4eeadd057da75976d26d21e79ca784970c62da10edb6928692f0ab9bf5e82ca55a2905c0307e9b2b3";
```
theghost5800: Why sha256 digest was used instead of sha-384?

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/Constants.java
```diff
@@ -8,7 +8,7 @@ public class Constants {
     public static final String INDEXED_STEP_NAME = "__INDEXED_STEP_NAME";
     public static final String VARIABLE_NAME_SERVICE_ID = "__SERVICE_ID";
 
-    public static final String DIGEST_ALGORITHM = "MD5";
+    public static final String DIGEST_ALGORITHM = "SHA-384";
```
theghost5800: Shouldn't we use SHA3 alrogirthm instead of SHA2? Also recommended length isn't it 512 instead of 384?

Another important thing is what will happen when this change is applied during blue-green update and there are running mta operations or files upload? 

What will happen if files were uploaded and persisted with md5 digest and then read/retrieved from updated instance of deploy service and try to parse stored md5 checksums as SHA-384? I think we have to implement some backwards compatible logic in order to avoid breaking running mta operations/running file uploads.

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/DeleteRemainingFilePartsTest.java
```diff
@@ -18,9 +18,9 @@
 
 class DeleteRemainingFilePartsTest extends SyncFlowableStepTest<DeleteRemainingFileParts> {
 
-    private static final String DIGEST_ALGORITHM = "MD5";
+    private static final String DIGEST_ALGORITHM = "SHA-384";
     private static final String NAMESPACE = "namespace";
-    private static final String DIGEST = "2252290BC44BEAD16AA1BF89948472E8";
+    private static final String DIGEST = "A1B2C3D4E5F6A1B2C3D4E5F6A1B2C3D4E5F6A1B2C3D4E5F6A1B2C3D4E5F6A1B2";
```
IvanBorislavovDimitrov: This digest is 64bit sha 256 not sha 384.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/Constants.java
IvanBorislavovDimitrov: Is there a performance degradation when the new algorithm is used?

IvanBorislavovDimitrov:
We also use MD5 in MtaMetadataUtil to hash the mta id. Do we need to change the algorithm there?

multiapps-controller-persistence/src/test/java/org/cloudfoundry/multiapps/controller/persistence/services/DatabaseFileServiceTest.java
vkalapov: Fixed with [93a4681](https://github.com/cloudfoundry/multiapps-controller/pull/1884/commits/93a4681b5cba9f01b3d9bb0f6e948e1c2e72a914), initial changes were done with sha256 but that was invalidated when the KSA requirements came up, and that resulted in some leftovers.

multiapps-controller-process/src/test/java/org/cloudfoundry/multiapps/controller/process/steps/DeleteRemainingFilePartsTest.java
vkalapov: Fixed with [71f2338](https://github.com/cloudfoundry/multiapps-controller/pull/1884/commits/71f2338e881a827aac39f65c04f41c3975f0d018), leftover from sha256 changes before KSA.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/Constants.java
vkalapov: SHA3 is needed for ADVANCED requirements, SHA-384 satisfies the moderate requirements in our case. 

Mid operation swap was tasted without failures. After the initial upload, the digest is only fetched and checked during write operations (i.e. uploading a new file). 
Also tested directly through the API endpoints to upload and start deploy operations separately with the initial being a md5 digest and the latter being after swap with sha-384 changes.

vkalapov:
> We also use MD5 in MtaMetadataUtil to hash the mta id. Do we need to change the algorithm there?

The MD5 in MtaMetadataUtil is out of scope for this PR - this one addresses file upload digest only. Since it  requires a migration strategy to avoid breaking detection of existing deployed MTAs.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/Constants.java
vkalapov: For file uploads, the operation is typically I/O-bound rather than hash-bound, so we don't expect a meaningful performance difference in practice. Either way, the mandatory performance tests should reveal any regressions.

IvanBorislavovDimitrov (dismissed):

IvanBorislavovDimitrov (approved):

theghost5800 (approved):

PR #1885, opened 2026-07-31
Adding upload statistics and alerts for both app binary uploads towards the CF API and OS uploads

karrgov:
Adding upload statistics and alerts for both app binary uploads towards the CF API and OS uploads

LMCROSSITXSADEPLOY-3054

### DIFF
--- a/multiapps-controller-persistence/src/main/java/module-info.java
+++ b/multiapps-controller-persistence/src/main/java/module-info.java
@@ -9,6 +9,7 @@ open module org.cloudfoundry.multiapps.controller.persistence {
     exports org.cloudfoundry.multiapps.controller.persistence.model;
     exports org.cloudfoundry.multiapps.controller.persistence.model.adapter;
     exports org.cloudfoundry.multiapps.controller.persistence.model.filters;
+    exports org.cloudfoundry.multiapps.controller.persistence.monitoring;
     exports org.cloudfoundry.multiapps.controller.persistence.query;
     exports org.cloudfoundry.multiapps.controller.persistence.query.criteria;
     exports org.cloudfoundry.multiapps.controller.persistence.query.impl;
--- a/multiapps-controller-persistence/src/main/java/module-info.java
+++ b/multiapps-controller-persistence/src/main/java/module-info.java
@@ -81,4 +81,5 @@ open module org.cloudfoundry.multiapps.controller.persistence {
     requires software.amazon.awssdk.retries.api;
     requires static java.compiler;
     requires static org.immutables.value;
+    requires io.netty.handler;
 }
--- a/multiapps-controller-persistence/src/main/java/module-info.java
+++ b/multiapps-controller-persistence/src/main/java/module-info.java
@@ -81,5 +81,4 @@ open module org.cloudfoundry.multiapps.controller.persistence {
     requires software.amazon.awssdk.retries.api;
     requires static java.compiler;
     requires static org.immutables.value;
-    requires io.netty.handler;
 }
--- a/multiapps-controller-persistence/src/main/java/module-info.java
+++ b/multiapps-controller-persistence/src/main/java/module-info.java
@@ -74,4 +74,5 @@ open module org.cloudfoundry.multiapps.controller.persistence {
     requires software.amazon.awssdk.retries.api;
     requires static java.compiler;
     requires static org.immutables.value;
+    requires io.netty.handler;
 }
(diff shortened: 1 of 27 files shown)


multiapps-controller-persistence/src/main/java/module-info.java
```diff
@@ -73,4 +74,5 @@
     requires software.amazon.awssdk.retries.api;
     requires static java.compiler;
     requires static org.immutables.value;
+    requires io.netty.handler;
```
IvanBorislavovDimitrov: If this is needed, sort it alphabetically

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/monitoring/UploadDurationTracker.java
```diff
@@ -0,0 +1,99 @@
...
+        public long totalCount() {
+            return total.sum();
+        }
+
+        public long timeoutCount() {
+            return timeouts.sum();
+        }
+
+        public long maxDurationMs() {
+            return maxDuration.getAndSet(0);
```
IvanBorislavovDimitrov, line 78: why is this set to 0 after the value is read?

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/monitoring/UploadDurationTracker.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        public long sumDurationMs() {
+            return sumDuration.sum();
+        }
+
+        public long timeoutsInWindow() {
+            return timeoutsInWindow.getAndSet(0);
```
IvanBorislavovDimitrov, line 86: same?

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/monitoring/UploadDurationTracker.java
```diff
@@ -0,0 +1,99 @@
+package org.cloudfoundry.multiapps.controller.persistence.monitoring;
+
+import jakarta.inject.Named;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.LongAdder;
+
+@Named
+public class UploadDurationTracker {
```
IvanBorislavovDimitrov, line 9: This class which dynatrace uses for monitoring is only in memory - per instance and the values will be zero after restart of instance of DS, LSS. This is expected right?

karrgov: Yes, I made it like that on purpose so we are better able to see when and if an instance has any problems and also we do not have to think and be responsible about the persistence part of the whole thing, since Dynatrace automatically saves the information on the charts, even after if lets say there have been lots of restarts. We would only have to change the timeframe and we will be able to see if and which metric exactly has abnormal values.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/monitoring/UploadDurationTracker.java
karrgov: This is on purpose as well, since Dynatrave is configured to poll each one minute which means that the poll happens at a random moment during these 60 seconds, which I would say that it is not fully correct, since a better way would be the way I have done it now - I am saving the maximum value per every 60 second poll and that is the value that dynatrace then displays on the charts. The idea is that in this one minute there might be a problematic deployment with lets say an abnormally high upload time and then after a couple of seconds there might be another totally normal one and the value would be overwritten and then if the Dynatrace poll is right after that - the chart would show that everything is totally alright.

multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/monitoring/UploadDurationTracker.java
karrgov: same as the one above :D

IvanBorislavovDimitrov (dismissed):

multiapps-controller-persistence/src/main/java/module-info.java
karrgov: discussed :D

IvanBorislavovDimitrov (dismissed):

Yavor16 (dismissed):

IvanBorislavovDimitrov (dismissed):

Yavor16 (approved):

IvanBorislavovDimitrov (approved):

PR #1889, opened 2026-08-03
Log async upload job details correlated with operation id

IvanBorislavovDimitrov:
## What

When an operation starts, log the async upload job (deploy-from-URL scenario) associated with that operation, correlated by the operation id — in a single, greppable line.

Previously there was no log line tying a deploy operation to its async upload job. The only shared key between the two is the uploaded file id (`appArchiveId`), and it was never logged next to the operation id, so correlating an operation with its upload required manual cross-referencing.

## Changes

- **`StartProcessListener`** — after files are stamped with the operation id, resolve the operation's file ids (`appArchiveId` + ext-descriptor ids) via `OperationFileIdsUtil` and query `AsyncUploadJobService.withFileIds(...)`. For each associated job, log one INFO line:

  ```
  Async upload job for operation "<operationId>" - id: …, state: FINISHED, fileId: …, mtaId: …, bytesRead: …, addedAt: …, startedAt: …, finishedAt: …, queueWaitTime: 30000 ms, uploadDuration: 90000 ms, totalTime: 120000 ms, error: null
  ```

  Best-effort: operations without file ids (e.g. undeploy) short-circuit, and any lookup failure logs a WARN rather than failing the operation. Injects `AsyncUploadJobService` (already available transitively, same as `OrphanedFilesCleaner`).

- **`AsyncUploadJobEntry`** — adds a credential-free `buildLogSummary()` plus timing derivations:
  - `getQueueWaitTime()` = `addedAt → startedAt` (time parked in the queue)
  - `getUploadDuration()` = `startedAt → finishedAt` (actual upload time)
  - `getTotalTime()` = `addedAt → finishedAt`
  - Timings return `null` (rendered as `N/A`) when the relevant timestamps are not set yet.

- **Sensitive data** — the summary intentionally omits `url` and `user` (may carry basic-auth credentials / PII) and `spaceGuid`.

## Tests

- New `AsyncUploadJobEntryTest` — timing derivations for finished/running/initial jobs, sensitive-data redaction, and `N/A` rendering.
- Updated `StartProcessListenerTest` for the new constructor dependency.

Both suites green; `multiapps-controller-persistence` and `-process` build clean.

### DIFF
--- a/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/model/AsyncUploadJobEntry.java
+++ b/multiapps-controller-persistence/src/main/java/org/cloudfoundry/multiapps/controller/persistence/model/AsyncUploadJobEntry.java
@@ -1,6 +1,7 @@
 package org.cloudfoundry.multiapps.controller.persistence.model;
 
 import java.text.MessageFormat;
+import java.time.Duration;
 import java.time.LocalDateTime;
 
 import org.cloudfoundry.multiapps.common.Nullable;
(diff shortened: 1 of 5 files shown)


vkalapov (dismissed):
Looks good but check the sonar scan warnings before merge.

IvanBorislavovDimitrov:
Addressed the SonarCloud warnings in bdb2e3457:

- **S8700 (reliability bug, ×3, was failing the quality gate)** — `getQueueWaitTime`/`getUploadDuration`/`getTotalTime` now compute the `Duration` over time-zone-aware `Instant`s (UTC) instead of `LocalDateTime`. Extracted a shared `durationBetween(...)` helper, which also removes the duplicated null-checks.
- **S2629** — the async-upload-job INFO line is now guarded with `LOGGER.isInfoEnabled()` so the `MessageFormat.format(...)` / `buildLogSummary()` aren't invoked unless INFO is on.
- **S8694** — test uses the `java.time.Month.AUGUST` enum instead of the `8` int literal.
- **S8924** — statically import `Mockito.when` in the test.

Both `AsyncUploadJobEntryTest` and `StartProcessListenerTest` are green and both modules build clean.

vkalapov (approved):

PR #1893, opened 2026-08-07
Remove inheritVariables for processBatches call activities

stiv03:
LMCROSSITXSADEPLOY-3029

Replace flowable:inheritVariables="true" on the processBatches call activities in rollback-mta, xs2-deploy and xs2-bg-deploy with an explicit flowable:in variable list, so only the variables the subprocess actually needs are passed instead of the whole parent variable map.

The required set was derived from:
- the processBatches step (ExtractBatchedServicesWithResolvedDynamicParametersStep) and its SyncFlowableStep base (correlationId, isSecurityEnabled, stepPhase),
- the variables it forwards to createOrUpdateServicesSubProcess (already an explicit flowable:in list), minus variables produced inside processBatches itself: servicesToCreate (written, never read) and serviceToProcess (the child multi-instance element variable). dynamicResolvableParameters is kept because the step reads it before overwriting it. batchToProcess remains the caller's multi-instance element variable and is injected per iteration.

### DIFF
--- a/multiapps-controller-process/src/main/resources/org/cloudfoundry/multiapps/controller/process/xs2-deploy.bpmn
+++ b/multiapps-controller-process/src/main/resources/org/cloudfoundry/multiapps/controller/process/xs2-deploy.bpmn
@@ -119,7 +119,34 @@
         <flowable:executionListener event="take" delegateExpression="${doNotDeleteServicesListener}"></flowable:executionListener>
       </extensionElements>
     </sequenceFlow>
-    <callActivity id="processBatchesSequentially" name="Process Resource Batches Sequentially" flowable:async="true" calledElement="processBatches" flowable:calledElementType="key" flowable:inheritVariables="true" flowable:completeAsync="true" flowable:fallbackToDefaultTenant="false">
+    <callActivity id="processBatchesSequentially" name="Process Resource Batches Sequentially" flowable:async="true" calledElement="processBatches" flowable:calledElementType="key" flowable:completeAsync="true" flowable:fallbackToDefaultTenant="false">
+      <extensionElements>
+        <flowable:in source="user" target="user"></flowable:in>
+        <flowable:in source="userGuid" target="userGuid"></flowable:in>
+        <flowable:in source="correlationId" target="correlationId"></flowable:in>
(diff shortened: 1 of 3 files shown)


multiapps-controller-process/src/main/resources/org/cloudfoundry/multiapps/controller/process/xs2-deploy.bpmn
```diff
@@ -119,7 +119,34 @@
         <flowable:executionListener event="take" delegateExpression="${doNotDeleteServicesListener}"></flowable:executionListener>
       </extensionElements>
     </sequenceFlow>
-    <callActivity id="processBatchesSequentially" name="Process Resource Batches Sequentially" flowable:async="true" calledElement="processBatches" flowable:calledElementType="key" flowable:inheritVariables="true" flowable:completeAsync="true" flowable:fallbackToDefaultTenant="false">
+    <callActivity id="processBatchesSequentially" name="Process Resource Batches Sequentially" flowable:async="true" calledElement="processBatches" flowable:calledElementType="key" flowable:completeAsync="true" flowable:fallbackToDefaultTenant="false">
```
IvanBorislavovDimitrov, line 122: Variable externalLoggingServiceConfigurations seems to be missing for the new cloud logging feature. Maybe needed in all diagrams

IvanBorislavovDimitrov:
Add LMCROSSITXSADEPLOY-3029 in commit as well

multiapps-controller-process/src/main/resources/org/cloudfoundry/multiapps/controller/process/xs2-deploy.bpmn
stiv03: I added it

Yavor16 (changes requested):
Can you check if the following variables must be added:
isServiceBindingKeyOperationInProgress
waitBindServiceTimeout
wasServiceBindingKeyOperationAlreadyDone

Yavor16 (approved):

IvanBorislavovDimitrov (approved):

PR #1897, opened 2026-08-14
Replace inheritVariables with explicit in-mappings for undeployAppSubProcess

Yavor16:
#### Description:

Removes the implicit full-scope variable inheritance from every `undeployAppSubProcess` call activity in the process definitions. All 5 call sites previously used `flowable:inheritVariables="true"`, which passes the entire parent process variable scope into the subprocess. This creates hidden coupling and can cause variables to leak unintentionally between process instances in multi-instance scenarios.

**What changed:** each call site now sets `flowable:inheritVariables="false"` and declares an explicit list of 27 `flowable:in` variable mappings, so the subprocess receives exactly the variables it needs and nothing more. This mirrors the pattern already applied to the `processBatches` subprocess in #1893.

**Affected BPMN files (4 files, 5 call sites):**
- `rollback-mta.bpmn` (1 call site)
- `xs2-bg-deploy.bpmn` (2 call sites)
- `xs2-deploy.bpmn` (1 call site)
- `xs2-undeploy.bpmn` (1 call site)

**Areas to focus on:** please verify the 27-variable `flowable:in` list is complete for each call site — a missing mapping would surface as an unset variable inside `undeployAppSubProcess` at runtime rather than at build time.

Reference: JIRA:LMCROSSITXSADEPLOY-3028

### DIFF
--- a/multiapps-controller-process/src/main/resources/org/cloudfoundry/multiapps/controller/process/xs2-undeploy.bpmn
+++ b/multiapps-controller-process/src/main/resources/org/cloudfoundry/multiapps/controller/process/xs2-undeploy.bpmn
@@ -89,9 +89,6 @@
         <flowable:in source="__SERVICE_ID" target="__SERVICE_ID"></flowable:in>
         <flowable:in source="mtaMajorSchemaVersion" target="mtaMajorSchemaVersion"></flowable:in>
         <flowable:in source="namespace" target="namespace"></flowable:in>
-        <flowable:in source="keepFiles" target="keepFiles"></flowable:in>
-        <flowable:in source="appArchiveId" target="appArchiveId"></flowable:in>
-        <flowable:in source="mtaExtDescriptorId" target="mtaExtDescriptorId"></flowable:in>
         <flowable:in source="appToProcess" target="appToProcess"></flowable:in>
       </extensionElements>
       <multiInstanceLoopCharacteristics isSequential="false" flowable:collection="appsToUndeploy" flowable:elementVariable="appToProcess">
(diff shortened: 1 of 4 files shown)


multiapps-controller-process/src/main/resources/org/cloudfoundry/multiapps/controller/process/xs2-undeploy.bpmn
```diff
@@ -63,7 +63,37 @@
...
+        <flowable:in source="isDisposableUserProvidedServiceEnabled" target="isDisposableUserProvidedServiceEnabled"></flowable:in>
+        <flowable:in source="disposableUserProvidedServiceName" target="disposableUserProvidedServiceName"></flowable:in>
+        <flowable:in source="processId" target="processId"></flowable:in>
+        <flowable:in source="useLastOperationForServiceBindingDeletion" target="useLastOperationForServiceBindingDeletion"></flowable:in>
+        <flowable:in source="__SERVICE_ID" target="__SERVICE_ID"></flowable:in>
+        <flowable:in source="mtaMajorSchemaVersion" target="mtaMajorSchemaVersion"></flowable:in>
+        <flowable:in source="namespace" target="namespace"></flowable:in>
+        <flowable:in source="keepFiles" target="keepFiles"></flowable:in>
+        <flowable:in source="appArchiveId" target="appArchiveId"></flowable:in>
+        <flowable:in source="mtaExtDescriptorId" target="mtaExtDescriptorId"></flowable:in>
```
stiv03: Are keepFiles/ appArchiveId/ mtaExtDescriptorId needed here in the undeploy diagram?

stiv03 (changes requested):

stiv03 (approved):

vkalapov (approved):

PR #1901, opened 2026-08-24
Introduce operations rate limits

IvanBorislavovDimitrov:
[LMCROSSITXSADEPLOY-3360](https://jira.tools.sap/browse/LMCROSSITXSADEPLOY-3360)


multiapps-controller-core/src/test/java/org/cloudfoundry/multiapps/controller/core/util/ApplicationConfigurationTest.java
```diff
@@ -464,6 +464,131 @@ void testGetSpringSchedulerTaskExecutorThreads() {
         Assertions.assertEquals(executorThreads, configuration.getSpringSchedulerTaskExecutorThreads());
     }
 
+    @Test
+    void testIsOperationRateLimitingEnabled() {
+        Mockito.when(environment.getBoolean(ApplicationConfiguration.CFG_OPERATION_RATE_LIMITING_ENABLED,
```
karrgov: maybe remove the "Mockito." and make it a static import in order to be consistent with how it is in the OperationRateLimitBucketCleanerTest

multiapps-controller-web/src/test/java/org/cloudfoundry/multiapps/controller/web/util/OperationRateLimitKeysTest.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        int expectedDistinctKeys = spaceGuids.size() + spaceGuids.size() * users.size();
+        assertEquals(expectedDistinctKeys, keys.size());
+    }
+
+    @Test
+    void testSpaceKeyIsStableAcrossRuns() {
+        long firstValue = OperationRateLimitKeys.spaceKey(SPACE_GUID);
+        long secondValue = OperationRateLimitKeys.spaceKey(SPACE_GUID);
+        assertTrue(firstValue == secondValue);
```
vkalapov: Why isn't assertEquals used here?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/util/OperationRateLimiter.java
```diff
@@ -0,0 +1,113 @@
...
+                                                      .size();
+        if (activeOperationsPerUser >= applicationConfiguration.getMaxActiveOperationsPerUser()) {
+            rejectAndLog(user, spaceGuid, Messages.TOO_MANY_ACTIVE_OPERATIONS_FOR_USER, NO_RETRY_AFTER_SECONDS);
+        }
+    }
+
+    private void checkTokenBuckets(String user, String spaceGuid) {
+        consumeSpaceToken(spaceGuid, user);
+        consumeUserToken(spaceGuid, user);
+    }
```
vkalapov, line 82: What happens if the space token is consumed but after that the user bucket is exhausted?  Wouldn't that lead to losing space tokens for other users in the space? Would checking the user first make more sense in that case?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/util/OperationRateLimitKeys.java
```diff
@@ -0,0 +1,39 @@
...
+ * Keys are computed from the SHA-256 digest of a namespaced input string and are therefore deterministic across restarts and JVMs. The
+ * space and user namespaces are disjoint by construction, so a space key can never collide with a user key. Truncating the 256-bit digest
+ * to its first 64 bits keeps the collision probability negligible for the number of distinct spaces and users a single landscape handles.
+ */
+public final class OperationRateLimitKeys {
+
+    private static final String SPACE_NAMESPACE_PREFIX = "space:";
+    private static final String USER_NAMESPACE_PREFIX = "user:";
+    private static final String SEGMENT_SEPARATOR = ":";
+    private static final HashFunction HASH_FUNCTION = Hashing.sha256();
```
vkalapov: We shouldn't use sha256 to be compliant according to internal cryptography requirements. com.google.common.hash.Hashing provides sha384 - https://guava.dev/releases/33.5.0-jre/api/docs/com/google/common/hash/Hashing.html#sha384() which is compliant.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/monitoring/ActiveOperationsJmxReporter.java
```diff
@@ -0,0 +1,114 @@
...
+                                    Map<ObjectName, ActiveOperationsCount> registry) throws Exception {
+        for (ObjectName name : stale) {
+            mBeanServer.unregisterMBean(name);
+            registry.remove(name);
+        }
+    }
+
+    static String hashUser(String user) {
+        try {
+            byte[] digest = MessageDigest.getInstance("SHA-256")
```
vkalapov: We should avoid using SHA-256 for internal cryptography requirements. Please use a compliant algorithm such as SHA-384.

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/util/OperationRateLimiter.java
```diff
@@ -0,0 +1,113 @@
...
+                                                       .size();
+        if (activeOperationsPerSpace >= applicationConfiguration.getMaxActiveOperationsPerSpace()) {
+            rejectAndLog(user, spaceGuid, Messages.TOO_MANY_ACTIVE_OPERATIONS_IN_SPACE, NO_RETRY_AFTER_SECONDS);
+        }
+        int activeOperationsPerUser = operationService.createQuery()
+                                                      .user(user)
+                                                      .spaceId(spaceGuid)
+                                                      .inNonFinalState()
+                                                      .list()
+                                                      .size();
```
vkalapov: The second query is a subset of the first - both filter by spaceId + inNonFinalState. Could the user count be derived from the already-fetched list to avoid the second DB query?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/util/OperationRateLimiter.java
```diff
@@ -0,0 +1,113 @@
...
+        Bucket bucket = bucketStore.getBucket(OperationRateLimitKeys.spaceKey(spaceGuid), configuration);
+        consumeToken(bucket, user, spaceGuid);
+    }
+
+    private void consumeUserToken(String spaceGuid, String user) {
+        BucketConfiguration configuration = buildBucketConfiguration(applicationConfiguration.getOperationRateLimitPerUserCapacity(),
+                                                                     applicationConfiguration.getOperationRateLimitPerUserRefillPerHour());
+        Bucket bucket = bucketStore.getBucket(OperationRateLimitKeys.userKey(spaceGuid, user), configuration);
+        consumeToken(bucket, user, spaceGuid);
+    }
```
vkalapov, line 92: BucketConfiguration is built on every call to checkStartAllowed - once for space, once for user. Since ApplicationConfiguration values don't change at runtime, could these be built once and reused?

multiapps-controller-process/pom.xml
```diff
@@ -25,6 +25,23 @@
...
+                            <goal>verify</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <includes>
+                        <include>**/*IntegrationTest</include>
+                    </includes>
+                </configuration>
+            </plugin>
```
karrgov: is that needed?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/monitoring/ActiveOperationsJmxReporter.java
```diff
@@ -0,0 +1,114 @@
...
+    private final Map<ObjectName, ActiveOperationsCount> userBeans = new ConcurrentHashMap<>();
+    private final Map<ObjectName, ActiveOperationsCount> spaceBeans = new ConcurrentHashMap<>();
+
+    @Inject
+    public ActiveOperationsJmxReporter(OperationService operationService, MBeanServer mBeanServer) {
+        this.operationService = operationService;
+        this.mBeanServer = mBeanServer;
+    }
+
+    @Scheduled(fixedRate = 1, timeUnit = TimeUnit.MINUTES)
```
vkalapov, line 51: Should we add the same instance guard - SELECTED_INSTANCE_FOR_CLEAN_UP from OperationRateLimitBucketCleaner in order to not run queries on every instance?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/monitoring/ActiveOperationsJmxReporter.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private Map<String, Long> countActiveOperationsBySpace() {
+        Map<String, Long> counts = new HashMap<>();
+        operationService.createQuery()
+                        .inNonFinalState()
+                        .list()
+                        .forEach(op -> counts.merge(op.getSpaceId(), 1L, Long::sum));
+        return counts;
+    }
```
vkalapov: Consider implementing a count query instead of list() to avoid loading all operations into memory just to count them.

vkalapov (changes requested):

multiapps-controller-web/src/test/java/org/cloudfoundry/multiapps/controller/web/api/impl/OperationsApiServiceImplTest.java
```diff
@@ -213,6 +219,44 @@ void testStartOperation() {
                .startProcess(Mockito.any(), Mockito.anyMap());
     }
 
+    @Test
+    void testStartOperationWhenRateLimitAllowsStartsProcess() {
```
karrgov, line 234: just to remove "Mockito." for when and verify and make it static

pom.xml
```diff
@@ -176,6 +177,12 @@
                         <runOrder>random</runOrder>
                     </configuration>
                 </plugin>
+                <!-- https://mvnrepository.com/artifact/org.apache.maven.plugins/maven-failsafe-plugin -->
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-failsafe-plugin</artifactId>
+                    <version>3.5.4</version>
+                </plugin>
```
karrgov: probably related to the other thing with the integration tests, is it needed?

multiapps-controller-web/src/main/java/org/cloudfoundry/multiapps/controller/web/monitoring/ActiveOperationsJmxReporter.java
Yavor16: I tried to do it but the code looked very ugly so I changed it a little bit but it is still in memory

karrgov (approved):

